PR #5, opened 2021-09-16
Generate and deploy bubble test coverage reports. 

cqc-alec:


lmondada (approved):
That's a nice simplification! How come the lengthy cmake files are no longer necessary?

cqc-alec:
> That's a nice simplification! How come the lengthy cmake files are no longer necessary?

They were based on some third-party code that was more general than we need -- this basically does the same thing, but implemented via conan options on the `tket` and `tket-tests` recipes.

PR #11, opened 2021-09-19
Fix special-case handling in `tk1_angles_from_unitary`.

### LINKED ISSUE #9: Phase error in compilation of circuit with X and BRIDGE.
A pi phase error is introduced when applying `FullPeepholeOptimise` to a circuit consisting of an X and a BRIDGE gate:

```python
from pytket.circuit import Circuit, OpType
from pytket.passes import FullPeepholeOptimise
import numpy as np

c = Circuit(3)
c.add_gate(OpType.X, [1])
c.add_gate(OpType.BRIDGE, [0, 1, 2])

u = c.get_unitary()

FullPeepholeOptimise().apply(c)

u1 = c.get_unitary()

assert np.allclose(u, -u1)
```

cqc-alec:
And add a test.

The handling of the first case was incorrect: I'm not sure what I was thinking there. I've added commentary.

Fixes #9 .


yao-cqc (approved):
Looks good!

PR #39, opened 2021-09-30
Add `allow_swaps` parameter to `FullPeepholeOptimise`

### LINKED ISSUE #28: Add `allow_swaps` option to `FullPeepholeOptimise`.
There isn't an option to disallow wireswaps when applying `FullPeepholeOptimise`. One can construct an equivalent pass that doesn't allow wireswaps:
```python
FullPeepholeOptimiseNoSwaps = SequencePass(
    [
        SynthesiseTket(),
        KAKDecomposition(),
        CliffordSimp(allow_swaps=False),
        SynthesiseTket(),
        ThreeQubitSquash(allow_swaps=False),
        CliffordSimp(allow_swaps=False),
        SynthesiseTket(),
    ]
)
```
But it would be better to add an `allow_swaps` option to the high-level pass.

cqc-alec:
Closes #28 .


pytket/binders/passes.cpp
```diff
@@ -343,7 +343,9 @@ PYBIND11_MODULE(passes, m) {
       "FullPeepholeOptimise", &FullPeepholeOptimise,
       "Performs peephole optimisation including resynthesis of 2- and 3-qubit "
       "gate sequences, and converts to a circuit containing only CX and TK1 "
-      "gates.");
+      "gates."
+      "\n\n:param: allow_swaps whether to allow implicit wire swaps",
```
lmondada, line 347:
```suggestion
      "\n\n:param: allow_swaps whether to allow implicit wire swaps.",
```

lmondada (changes requested):
Looking good, there is just one tiny dot missing in the docstring.

pytket/binders/passes.cpp
cqc-alec: Done (and fixed a few other instances too).

lmondada (approved):

PR #44, opened 2021-10-04
Use header-only version of spdlog/fmt.

### LINKED ISSUE #43: Avoid linker warnings from `fmt` on MacOS
MacOS builds of pytket include a lot of warnings like:
```
ld: warning: direct access in function 'fmt::v8::appender fmt::v8::detail::write<char, fmt::v8::appender, int, 0>(fmt::v8::appender, int)' from file 'CMakeFiles/circuit.dir/binders/circuit/Circuit/main.cpp.o' to global weak symbol 'fmt::v8::detail::basic_data<void>::digits' from file '/Users/runner/.conan/data/fmt/8.0.1/_/_/package/e49065ecb5f0d44c4e32f570e965f6a1bd08dd0e/lib/libfmt.a(format.cc.o)' means the weak symbol cannot be overridden at runtime. This was likely caused by different translation units being compiled with different visibility settings.
```
Not sure if these matter but it would be good to remove them.

The `fmt` library is a depenency of `spdlog`. It has a `header_only` option which could be worth investigating.

cqc-alec:
Closes #43 .

I am in two minds about merging this. It gets rid of the linker warnings, but may make builds slightly slower.

### DIFF
--- a/bubble/src/Routing/Swap_Analysis.cpp
+++ b/bubble/src/Routing/Swap_Analysis.cpp
@@ -202,7 +202,7 @@ std::vector<Swap> Routing::path_to_swaps(const std::vector<Node> &path) {
 // it.
 bool Routing::solve_furthest() {
   bool success = false;
-  std::optional<Node> max_node = std::nullopt;
+  std::optional<Node> max_node;
   unsigned max_dist = 0;
   for (auto [q1, q2] : interaction) {
     unsigned dist = current_arc_.get_distance(q1, q2);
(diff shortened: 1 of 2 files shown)


drzenharper (approved):

cqc-alec:
@lmondada There may be some tricks we can do to speed it up, but I don't know enough about spdlog internals to guess what those are, and ultimately the right approach would probably be to wrap tket logging in its own library -- as part of the modularization project.

PR #49, opened 2021-10-06
fix aas working with less qubits in circ then nodes in arch

cqc-melf:


cqc-melf:
If there are no other requests for changes I will try to apply the rename bubble changes if possible or make a new PR to get this merged.

PR #52, opened 2021-10-07
Fix catch test order in code and make namespaces consistent

willsimmons1465:
Ported over from the ZX feature PR on the old private development repository.

The test order of catch2 is linker-dependent. Whilst tests from different .cpp files will be run in the order the files are listed in the CMakeLists.txt on most systems, this is not guaranteed. For example, when previously trying to link in the python interpreter through pybind11 it would reverse this order on our Linux CI. This PR includes all test files into a single .cpp file to ensure that the order is coded in. One test that relied on the order of test files (i.e. was dependent on the symbol table being empty) has been edited to remove this dependency.

During this, I also noticed we were inconsistently using namespaces `tket::graph` and `tket::graphs`, so I have made this consistent across the codebase.


cqc-alec:
I'm not convinced we need or want this.
* It will make recompilation slower when only a single test file is changed, because everything will have to be recompiled.
* It is already possible to restrict a test run to a single source file, so when fixing specific failures we don't need to rerun everything.
* Including cpp files in other cpp files is pretty confusing and may confuse some tools as well.

PR #86, opened 2021-10-20
Exclude lines containing TKET_ASSERT from coverage analysis.

### LINKED ISSUE #83: Coverage analysis should ignore the branch in `TKET_ASSERT`.
By design, the code that logs a message and aborts in a `TKET_ASSERT` should never be executed. Therefore this should be excluded from the analysis of line and branch coverage.

cqc-alec:
Closes #83 .


cqc-melf:
I think it is fine now, Maybe we should add as task for the future to check of some exceptions should be replaced by a tket assert in the current code.

PR #92, opened 2021-10-22
Optimize handling of fully-connected architectures.

cqc-alec:
Makes it possible to define fully-connected architectures on thousands of nodes without crashing the system. The price we pay is that the graph is unweighted, so not suitable for noise-aware placement. (After merging I plan to make an RC and then a separate extensions PR to fix up pytket-braket which breaks because of this.)

### DIFF
--- a/pytket/binders/routing.cpp
+++ b/pytket/binders/routing.cpp
@@ -15,6 +15,7 @@
 #include "Routing/Routing.hpp"
 
 #include <pybind11/eigen.h>
+#include <pybind11/operators.h>
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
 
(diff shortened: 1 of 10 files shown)


lmondada:
I feel a bit queezy with all the `if/else` branching. This seems like a case for polymorphism.

Ideally, I would refactor the code as follows:
* `UIDConnectivityBase`: base (abstract) class that defines the interface
* `UIDConnectivityAdjList`: class with BGL graph implementation of interface (inherits from `UIDConnectivityBase`). The caching currently in `UIDConnectivity` could be in this class.
* `UIDConnectivityFull`: class with fully connected implementation of interface (inherits from `UIDConnectivityBase`)
* `Architecture<GraphType>`: templated class inheriting from `GraphType`
* Define shortcuts such as `FullyConnected = Architecture<UIDConnectivityFull>`. We could also change the templated class to eg `ArchitectureBase<T>` and define `Architecture = ArchitectureBase<UIDConnectivityAdjList>`

This would be cleaner and extensible, in case we want to define graphs with other structure, eg a square lattice that scales well.

I understand this is quite some work, so we could merge this as is for the moment, and refactor later, so that the FullyConnected architecture is available sooner.

What do you think?

cqc-alec:
> What do you think?

I think that's a good idea. I had originally intended something like this, but was put off by some of the complications (around caching, for instance). But yes, on reflection we should do it this way rather than add more complexity.

lmondada:
One more thing maybe to keep in mind during the refactor: serialization of `Architecture`s.

As far as I remember, it is currently impossible to recover the type of the architecture if it is from a subtype of Architecture. It would be good if we could change this.

cqc-alec:
Replaced by #116 .

PR #99, opened 2021-10-30
Added comment to `Gate::get_tk1_angles` and ran clang-format

alexcowtan:


tket/src/Gate/Gate.hpp
```diff
@@ -35,6 +35,8 @@ class Gate : public Op {
 
   std::string get_command_str(const unit_vector_t &args) const override;
 
+  // return the gate decomposition in terms of Rz(a)Rx(b)Rz(c), and a global
+  // phase.
```
cqc-alec, line 39:
```suggestion
  /**
   * Return the gate decomposition in terms of Rz(a)Rx(b)Rz(c).
   * 
   * This decomposition is in matrix-multiplication order, i.e. the reverse of
   * circuit order.
   * 
   * @return a, b, c and a global phase
   */
```
Thanks! Better to use doxygen-style comment markers as these then show up in the API docs (https://cqcl.github.io/tket/tket/api/Gate_8hpp.html). Also clarify that matrix-multiplication order is intended.

cqc-alec (changes requested):

cqc-alec (dismissed):

PR #107, opened 2021-11-05
[infra] reordered test files

lmondada:
I've just reordered the tests because I am in the habit of commenting out whatever I do not need, and the random ordering was making it harder to find what I was looking for.


sjdilkes (changes requested):
Hiya, sorry this PR completely missed me. It may be worth closing and reopening this given the changes that Alec has since made to the build infrastructure, or at least double checking it is up to date if you still think this reordering is important.

lmondada:
Hi, let's close this for now. I still think that separating test utils files from actual test cases is useful so that we can easily remove the uninteresting test cases. However, since this was all refactored this PR is useless. If this pops up again in my workflow I might address this in another PR in the future.

PR #108, opened 2021-11-08
Assert candidate swaps size

yao-cqc:


sjdilkes (approved):
Looks good, thanks Yao.

PR #113, opened 2021-11-10
Support Python 3.10; drop 3.7.

cqc-alec:
No openfermion test in 3.10. (Build issues due to an indirect dependency on an
older version of numpy. These should be fixed when h5py releases a 3.10 wheel.)

Python versions are strings, not numbers.

3.10 workflows tested [here](https://github.com/CQCL/tket/actions/runs/1443109875) and [here](https://github.com/CQCL/tket/actions/runs/1443109876).


.github/workflows/build_and_test.yml
```diff
@@ -268,37 +266,53 @@ jobs:
     - name: Build and run tket proptests
       run: conan create --profile=tket recipes/tket-proptests --build=missing -o boost:without_fiber=True -o boost:without_json=True -o boost:without_nowide=True
     - name: Build pytket (3.8)
-      if: github.event_name == 'pull_request' || github.event_name == 'push'
```
ss2165, line 271: are we also using this PR to reduce the frequency of some runs?

.github/workflows/build_and_test.yml
```diff
@@ -268,37 +266,53 @@ jobs:
...
-      if: github.event_name == 'schedule'
+      if: github.event_name == 'pull_request'
       run: |
         eval "$(pyenv init -)"
         pyenv shell tket-3.9
         cd pytket/tests
         pip install -r requirements.txt
         pytest --ignore=simulator/ --doctest-modules
+    - name: Build pytket (3.10)
+      if: github.event_name == 'schedule'
```
ss2165, line 301: the others are on "pull_request" but this is still "schedule"?

.github/workflows/build_and_test.yml
cqc-alec: No: we could not build 3.7 on the M1 previously, so used 3.8 for both PRs and pushes. Now M1 supports all 3 versions so we are following the same pattern as for the other platforms,

.github/workflows/build_and_test.yml
cqc-alec: The pattern is: 3.8 for pushes, 3.9 for PRs, and 3.10 for the weekly scheduled run.

ss2165 (approved):

PR #116, opened 2021-11-11
Refactor graph classes

cqc-alec:
Replacement for #92 .

Because of #117 and the amount of template code, I've resorted to improving the test coverage in some unrelated code, mainly by refactoring and removing some unused functions.

Will require updates to pytket-braket and pytket-qiskit.

Sorry about the size of this PR, but I think it's quite easy to follow if you step through the commits (then at least you can see where files have been renamed, which isn't evident on the PR).

The general idea is:

* Define an `AbstractGraph` base class, templated on node type. (When apple-clang-13 has proper support for concepts we will add suitable constraints on the node type.)
* Define `DirectedGraph` as a derivative of `AbstractGraph`, implemented using the BGL. This corresponds to the old `UIDConnectivity` class, and like that it is implemented in two tiers, with all the functionality in `DirectedGraphBase` and caching added in `DirectedGraph` itself.
* Define `CompleteGraph` as another derivative of `AbstractGraph`. This doesn't use the BGL; it basically just stores the node set.
* Define `ArchitectureBase`, templated on the graph type.
* Derive `Architecture` from `ArchitectureBase<DirectedGraph<Node>>`, and `FullyConnected` from `ArchitectureBase<CompleteGraph<Node>>`.


tket/src/Architecture/Architecture.hpp
```diff
@@ -67,48 +67,68 @@ class ArchitectureBase : public T {
...
+   * Nodes that cannot be removed without breaking connectivity.
+   */
   node_set_t get_articulation_points() const;
-  node_set_t get_articulation_points(const ArchitectureBase &subarc) const;
 
-  /* returns new Architecture that is generated by a subset of nodes of `this`
+  /**
+   * Nodes that cannot be removed without breaking connectivity of the subarc.
+   */
+  node_set_t get_articulation_points(const Architecture &subarc) const;
```
lmondada, line 98: Could `subarc` be of type `ArchitectureBase`?

pytket/pytket/backends/backendinfo.py
```diff
@@ -143,7 +143,11 @@ def from_dict(cls, d: Dict[str, Any]) -> "BackendInfo":
         :rtype: BackendInfo
         """
         args = dict(**d)
-        args["architecture"] = Architecture.from_dict(args["architecture"])
+        arch = args["architecture"]
+        if "links" in arch:
+            args["architecture"] = Architecture.from_dict(args["architecture"])
+        else:
+            args["architecture"] = FullyConnected.from_dict(args["architecture"])
```
lmondada, line 150: Could we store a name or ID of the Architecture class used, so that it is easily extensible?

tket/src/Architecture/Architecture.hpp
```diff
@@ -48,44 +49,77 @@ static std::vector<std::pair<Node, Node>> as_nodepairs(
...
+   * Vectors of nodes corresponding to lines of given lengths
+   */
   std::vector<node_vector_t> get_lines(
       std::vector<unsigned> required_lengths) const;
 
-  // get architecture diameter
+  /**
+   * Diameter of graph
+   */
   unsigned get_diameter() const;
```
lmondada, line 114: It might make sense to move some of these methods (at least the basic ones such as `get_diameter` and `create_subarch`) to be part of a common API in `ArchitectureBase`.

tket/src/Architecture/Architecture.hpp
```diff
@@ -97,27 +131,33 @@ class Architecture : public graphs::UIDConnectivity<Node> {
...
+      nodes_.insert(Node("fcNode", i));
+    }
+  }
 };
 
+JSON_DECL(FullyConnected)
+
 // Subclass, constructor generates adjacency matrix corresponding to a ring
 // architecture
 class RingArch : public Architecture {
```
lmondada, line 156: This could be inheriting from `ArchitectureBase` directly, without relying on an explicit BGL graph.

lmondada: Same for SquareGrid below.

pytket/pytket/backends/backendinfo.py
lmondada: This would mean it would also work for RingArch and SquareArch.

lmondada: (this should probably be in serialization code, outside of `BackendInfo`)

lmondada (changes requested):
Thank you so much Alec! You've made this code much cleaner, it now all makes a lot more sense I think.

I have a few small comments, see below. In particular: do you know why you were not able to keep some of the templated code in `cpp` files? Given that we only ever instantiate `T` with `Node` or `Qubit`, it would make sense for compilation times to try to keep these templated classes in `cpp` files with extern templates. But I'm assuming you must have run into an issue.

tket/src/Architecture/Architecture.hpp
```diff
@@ -49,29 +49,35 @@ static std::vector<std::pair<Node, Node>> as_nodepairs(
...
 }
 
-class Architecture : public graphs::UIDConnectivity<Node> {
- private:
-  using Base = graphs::UIDConnectivity<Node>;
-
+/**
+ * Generic architecture class
+ *
+ * Constraint: the underlying graph type must have node type @ref Node.
```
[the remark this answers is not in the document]
cqc-alec: Yep, added that in a comment.

tket/src/Graphs/UIDConnectivityBase.hpp
```diff
@@ -24,13 +24,6 @@
 
 namespace tket::graphs {
 
-/** edges of connectivity graph */
-struct UIDInteraction {
-  explicit UIDInteraction(unsigned d = 1) : weight(d) {}
```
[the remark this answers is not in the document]
cqc-alec: This is never required.

tket/src/Architecture/Architecture.hpp
cqc-alec: I have moved `get_diameter` into `AbstractGraph`. I don't think `create_subarch` should be moved down because a subset of nodes of a given architecture type won't necessarily be of that type (e.g. a `RingArch`).

tket/src/Architecture/Architecture.hpp
cqc-alec: Would rather leave these here to avoid having to reimplement a bunch of methods for them, at least until there is some advantage to doing so.

pytket/pytket/backends/backendinfo.py
cqc-alec: Yup, I think this would make sense if we wanted to introduce special handling for the other types, but maybe a bit too much complexity just to handle `FullyConnected`?

pytket/binders/routing.cpp
```diff
@@ -196,7 +196,7 @@ PYBIND11_MODULE(routing, m) {
                ", columns=" + std::to_string(arc.get_columns()) +
                ", layers=" + std::to_string(arc.get_layers()) + ">";
       });
-  py::class_<FullyConnected, Architecture>(
+  py::class_<FullyConnected>(
```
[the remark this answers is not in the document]
cqc-alec: Don't want to expose too much implementation detail but I see your point. Perhaps something to think about if we have further proliferation of types?

tket/src/Architecture/Architecture.hpp
cqc-alec: The implementation actually uses the BGL methods so I think not easily.

tket/src/Architecture/Architecture.hpp
lmondada: I was thinking that these architectures, just like `FullyConnected` don't need to keep the graph in memory because of their regular structure. But I guess that unlike `FullyConnected`, these are sparse, so it shouldn't be too much of a problem.

pytket/binders/routing.cpp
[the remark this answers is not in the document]
lmondada: I will let you override me on this one if you feel strongly about it, but I personally do feel that having a single `Architecture` base class in the Python level users can rely on every architecture to be an instance of is useful.

tket/src/Graphs/UIDConnectivityBase.cpp
```diff
@@ -12,276 +12,4 @@
...
-
-namespace tket::graphs {
-
-template <typename UID_t>
-UIDConnectivityBase<UID_t>::UIDConnectivityBase(const std::vector<UID_t>& uids)
-    : AbstractGraph<UID_t>(uids), graph() {
-  for (const UID_t& uid : uids) {
-    add_uid(uid);
-  }
-}
```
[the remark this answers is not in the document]
lmondada: Hmmm, I see, that makes sense.

### Considerations
If we added the extern definitions to `Architecture.hpp` (which would be the ideal solution), then we'd have to include the `DirectedGraph.cpp` file into `Architecture.cpp`, which is terrible programming practice. 

However, removing the external definitions altogether is a bad idea in my opinion: the `DirectedGraph` code, and especially the BGL-related code is pretty heavy to compile and will not be changed very often, so it makes a lot of sense to pre-compile it.

### Proposal
We could create two new files `DirectedGraph_impl.{hpp,cpp}` (or `DirectedGraph_extern` or similar -- or `Architecture_graph` if we want it closer to the Architecture code) that have the `UnitID` dependency and the extern definitions. That should solve our issues, right?

lmondada (changes requested):
Hi Alec,

I agree with you that most of the requests I had made about unifying the interface is probably not worth it at the moment -- as you said, too much work for little benefit, and nothing will stop us from making these changes when it is actually useful.

However, the one place where I would argue this is worth doing is with at the Python interface level -- see my comment. Plus adding the extern definitions back. Let me know what you think.

Thanks for everything!

cqc-alec:
Thanks Luca. After merging from `develop` locally I now find I can't build `pytket` on this branch because of some compiler warnings. I will fix those, and then address your concerns.

On the "extern templates" question, could we put the `Architecture_extern.{cpp,hpp}` files in the `Architecture` directory? If so I think the code might as well go in `Architecture.{cpp,hpp}` -- or am I misunderstanding?

lmondada:
My thinking was that if you put it in Architecture you'd be recompiling the BGL code everytime you need to recompile Architecture -- this may or may not be a reasonable compromise, depending on how often we think we wlll be changing that. If it's bundled with Architecture it will also make it harder to reuse the same instantiation elsewhere in the code. I think right now that only applies to Routing/Placement.hpp, which depends on Architecture anyway.

Either is fine with me.

lmondada (approved):
Thank you so much Alec. Thanks to you I finally feel we have found a structure that works nicely for this part of the code base!

pytket/binders/routing.cpp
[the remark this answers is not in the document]
sjdilkes: If we want to expose this, do we also want to update methods that use Architectures (Placement, Routing, Compilation Passes) to accept the base class NodeGraph?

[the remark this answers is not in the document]
sjdilkes: Also note that currently a Placement method can not accept a `FullyConnected` object, when there is good reason to want to relabel logical qubits to `FullyConnected` Nodes.

PR #118, opened 2021-11-16
Revert "[infra] Support Python 3.10; drop 3.7. (#113)"

cqc-alec:
This reverts commit eaa1e3a4d7f5ff8f2f9a77fd4bf649be9d9648a4.

Unfortunately, pytket is subject to a crash-on-program-exit bug with Python 3.10:

```
$ python -c "from pytket import Circuit; c = Circuit(1).Rx(0.5, 0)"
Segmentation fault (core dumped)
```

My guess is that this is related to some issues in the symengine library which have been [fixed](https://github.com/symengine/symengine/pull/1849) on the master branch but not yet in a release.


ss2165 (approved):
:(

PR #121, opened 2021-11-16
Replace uses of `new` with `make_unique`.

### LINKED ISSUE #111: Abolish `new`.
There are a few places in the code where we still use `new` to construct a `unique_ptr`; we should replace these with `make_unique`.

cqc-alec:
Closes #111 .


yao-cqc (approved):
Looks good!

PR #128, opened 2021-11-22
Release/0.17.0

cqc-alec:


cqc-alec:
Just remembered about https://github.com/CQCL/tket/pull/127. Should try to get that into this release.

PR #141, opened 2021-12-06
[TKET-1652] Circuit::add_qubit modifies unit_bimaps_

sjdilkes:
unit_bimaps_ is a Circuit attribute used when applying some Pass through CompilationUnit. 
Typically when a pass is applied something like the following is done:
`
  c_unit.circ_.unit_bimaps_ = {&c_unit.initial_map_, &c_unit.final_map_};
  bool changed = trans_.apply(c_unit.circ_);
  c_unit.circ_.unit_bimaps_ = {nullptr, nullptr};`

i.e. Circuit.unit_bimaps_ is set by reference to that held in CompilationUnit, and is then cleansed to a nullptr afterwards.
It is possible that some pass may want to add a new qubit to a circuit, via `Circuit::add_qubit` -> currently this method does not update unit_bimaps_, meaning such a change would not be tracked by the associated CompilationUnit.

This PR updates add_qubit to modify unit_bimaps_ if it exists.

### DIFF
--- a/tket/src/Circuit/basic_circ_manip.cpp
+++ b/tket/src/Circuit/basic_circ_manip.cpp
@@ -392,6 +392,21 @@ void Circuit::add_qubit(const Qubit& id, bool reject_dups) {
   Vertex out = add_vertex(OpType::Output);
   add_edge({in, 0}, {out, 0}, EdgeType::Quantum);
   boundary.insert({id, in, out});
+
+  unit_bimap_t* ubmap_initial = unit_bimaps_.initial;
+  unit_bimap_t* ubmap_final = unit_bimaps_.final;
+
+  if (ubmap_initial && ubmap_final) {
+    if (ubmap_initial->right.find(id) != ubmap_initial->right.end() ||
+        ubmap_initial->left.find(id) != ubmap_initial->left.end() ||
+        ubmap_final->right.find(id) != ubmap_final->right.end() ||
+        ubmap_final->left.find(id) != ubmap_final->left.end()) {
+      throw CircuitInvalidity(
+          "A unit with ID \"" + id.repr() + "\" already exists");
+    }
+    ubmap_initial->left.insert({id, id});
+    ubmap_final->left.insert({id, id});
+  }
 }
 
 void Circuit::add_bit(const Bit& id, bool reject_dups) {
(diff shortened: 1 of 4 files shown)


tket/src/Circuit/basic_circ_manip.cpp
```diff
@@ -392,6 +392,21 @@ void Circuit::add_qubit(const Qubit& id, bool reject_dups) {
   Vertex out = add_vertex(OpType::Output);
   add_edge({in, 0}, {out, 0}, EdgeType::Quantum);
   boundary.insert({id, in, out});
+
+  unit_bimap_t* ubmap_initial = unit_bimaps_.initial;
+  unit_bimap_t* ubmap_final = unit_bimaps_.final;
+
+  if (ubmap_initial && ubmap_final) {
```
cqc-alec, line 399: Hmm, I don't know what should happen if `reject_dups` is `false`, i.e. we allow duplicates, but not in unit maps? Maybe circuits having duplicate IDs breaks some compilation passes? (Should we make a separate issue for this?)

tket/tests/test_CompilerPass.cpp
```diff
@@ -89,6 +89,58 @@ SCENARIO("Run some basic Compiler Passes") {
...
+    PostConditions postcon;
+    auto pass =
+        std::make_shared<StandardPass>(s_ps, t, postcon, nlohmann::json{});
+
+    // Comparison qubit
+    Qubit weird_qb4("weird_qb", 9);
+    pass->apply(cu);
+    cu_initial = cu.get_initial_map_ref();
+    // check all maps to show weird_qb4 is mapped to self in both initial and
+    // final
```
cqc-alec, line 136: Are we checking it's mapped to itself, or just that it's mapped to something?

tket/src/Circuit/basic_circ_manip.cpp
sjdilkes: My understanding was that `reject_dups` essentially determined if on a duplicate Qubit being added, whether an error is thrown telling the user, or its just ignored and everything continues.

Any Qubit in unit_bimaps_ should also be in the boundary, and so I believe the current logic covers this case and nothing needs to be changed.

The one case not covered with reject_dups is when a Qubit added is in unit_bimaps_ but not the boundary - something which within usual pass applying logic should never happen. If this does happen then I think it is better that an error is thrown even if reject_dups is False - the CircuitInvalidity at 404 covers this.

tket/tests/test_CompilerPass.cpp
sjdilkes: I've updated the test to check its also mapped to itself - i think its more important its just mapped but it seems harmless to additionally check.

cqc-alec (approved):

PR #143, opened 2021-12-07
Require openfermion ~= 1.3 for tests.

cqc-alec:
(With this version openfermion is not tied to a specific version of cirq.)


cqc-melf (approved):

cqc-alec:
Looks like we can't install openfermion on the Mac M1. Will close this PR now and wait till this issue is resolved before making it a requirement.

PR #144, opened 2021-12-07
[TKET-1603] Version report

Roland-djee:
Fixes [TKET-1603](https://cqc.atlassian.net/browse/TKET-1603?atlOrigin=eyJpIjoiNWRiMGU4ZDA0ZGI0NDI1YmJlN2ZiNWM5MjU5ZjM1NzEiLCJwIjoiaiJ9).

This PR addresses the report of the `pytket` version at runtime when query `pytket.__version__`. It uses `dunamai` as a runtime dependency to query `git` source tree.

### DIFF
--- a/pytket/pyproject.toml
+++ b/pytket/pyproject.toml
@@ -4,3 +4,5 @@ build-backend = "setuptools.build_meta"
 
 [tool.setuptools_scm]
 root = ".."
+write_to = "pytket/pytket/_version.py"
+write_to_template = '__version__ = "{version}"'
--- a/pytket/pytket/__init__.py
+++ b/pytket/pytket/__init__.py
@@ -14,7 +14,7 @@
 
 """Python Interface to CQC tket
 """
-
+import dunamai as _dunamai
 from pytket.circuit import (  # type: ignore
     Circuit,
     OpType,
@@ -26,3 +26,4 @@ import pytket.transform
 import pytket.telemetry
 
 __path__ = __import__("pkgutil").extend_path(__path__, __name__)  # type: ignore
+__version__ = _dunamai.Version.from_git().base
--- a/pytket/pytket/__init__.py
+++ b/pytket/pytket/__init__.py
@@ -14,7 +14,6 @@
 
 """Python Interface to CQC tket
 """
-import dunamai as _dunamai
 from pytket.circuit import (  # type: ignore
     Circuit,
     OpType,
@@ -25,5 +24,7 @@ import pytket.routing
 import pytket.transform
 import pytket.telemetry
 
+from pytket._version import __version__
+
 __path__ = __import__("pkgutil").extend_path(__path__, __name__)  # type: ignore
-__version__ = _dunamai.Version.from_git().base
+__version__ = '.'.join(__version__.split('.')[:3])
--- a/pytket/pytket/__init__.py
+++ b/pytket/pytket/__init__.py
@@ -27,4 +27,4 @@ import pytket.telemetry
 from pytket._version import __version__
 
 __path__ = __import__("pkgutil").extend_path(__path__, __name__)  # type: ignore
-__version__ = '.'.join(__version__.split('.')[:3])
+__version__ = ".".join(__version__.split(".")[:3])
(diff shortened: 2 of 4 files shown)


pytket/pytket/__init__.py
```diff
@@ -26,3 +26,4 @@
 import pytket.telemetry
 
 __path__ = __import__("pkgutil").extend_path(__path__, __name__)  # type: ignore
+__version__ = _dunamai.Version.from_git().base
```
cqc-alec, line 29: Surely this will depend on the directory the user is in when they import pytket?

cqc-alec (changes requested):

pytket/pytket/__init__.py
Roland-djee: Well, the user has to be where the VCS can be queried. Maybe best to have it static ?

cqc-alec: But a typical pytket user has not cloned the git repo, they just pip install the package.

Roland-djee: I guess you can integrate it when you build the wheel, from the [docs](https://pypi.org/project/dunamai/):

```
from setuptools import setup
from dunamai import Version

setup(
    name="your-library",
    version=Version.from_any_vcs().serialize(),
)
```

Roland-djee: Dunamai exists as a [GitHub Actions](https://github.com/marketplace/actions/run-dunamai). Might want that for every release ?

Roland-djee: We can set up a workflow that sets the version statically at every release. Might do the trick.

cqc-alec: I think I prefer the approach of adding it in `setup.py` (or `pyproject.toml`?), replacing the current use of `setuptools_scm`, but we should thoroughly test it.

Roland-djee: Yeah, can do in `pyproject.toml` too I think. Well, either one or the other unless there is a preference for the later.

cqc-alec: As we will soon be moving away from `setup.py` I would do it in `pyproject.toml` if possible (that is where `setuptools_scm` is at the moment).

Roland-djee: OK got that. I wasn't sure about the status of that transition yet. Will do.

pytket/pytket/__init__.py
```diff
@@ -25,4 +24,7 @@
 import pytket.transform
 import pytket.telemetry
 
+from pytket._version import __version__
+
 __path__ = __import__("pkgutil").extend_path(__path__, __name__)  # type: ignore
+__version__ = ".".join(__version__.split(".")[:3])
```
cqc-alec, line 30: Is this truncation needed? For released versions it will make no difference, and for dev versions is there any reason not to have the full version string?

pytket/pyproject.toml
```diff
@@ -4,3 +4,5 @@ build-backend = "setuptools.build_meta"
 
 [tool.setuptools_scm]
 root = ".."
+write_to = "pytket/pytket/_version.py"
```
cqc-alec, line 7: Please add `pytket/pytket/_version.py` to the `.gitignore` file since it's not under source control.

cqc-alec (changes requested):
Much nicer.

pytket/pyproject.toml
Roland-djee: Yes sure will do.

pytket/pytket/__init__.py
Roland-djee: No particular reasons really. I can make a case for different versions if I can discriminate between the two ?

Roland-djee: Thing is I haven't found an easy way to format the version string with `setuptools_scm` in the `pyproject.toml`_i.e._ to extract the base, commit id etc. which is made easy with `dunamai` for instance. However, there is no easy `dunamai` integration with `pyproject.toml` unfortunately.

cqc-alec: I'm just suggesting we leave it as it is, e.g. "0.17.1.dev12+g6129ff39" for a dev version. Is there any problem with that? (For a release version it would just be "0.17.1" or whatever.)

cqc-alec: (i.e. don't add line 30)

Roland-djee: Oh OK well sure. I thought release tags where under the same kind of long format. But it they are under the "X.Y.Z" form only then it makes sense.

cqc-alec (approved):

PR #145, opened 2021-12-07
Update check_correctness for compiled circuits with additional qubits

sjdilkes:
The proptests compare many circuits pre and post compilation. At comparison there are a few implicit assumptions about the circuits being compiled, one being that post-compiled circuits will have the same number of qubits as they did pre.

This is no longer a reasonable assumption - some compilation passes may add additional qubits to a circuit, one example being the use of auxilary qubits in circuit mapping to some Architecture (i.e. "RoutingV3").

We may also expect to shrink circuits in the future through compilation - [arxiv.org/pdf/2105.09324.pdf](url).

This PR asserts that at comparison compiled circuits have the same or more qubits than pre-compiled circuits (currently we expect no compiler passes to shrink circuits), and also updates the comparison circuits such that the unitary matrix comparison can be completed.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -185,7 +185,9 @@ jobs:
     - name: Build and run tket tests
       run: conan create --profile=tket recipes/tket-tests
     - name: Build and run tket proptests
-      run: conan create --profile=tket recipes/tket-proptests
+      run: |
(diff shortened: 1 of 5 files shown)


cqc-melf (dismissed):
Looks good!

cqc-melf (approved):

Issue #147, opened 2021-12-08
`SquashHQS` does not squash 2x `PhasedX`


lmondada:
Consider the following circuit

```
import pytket as tk

c = tk.Circuit(1)
c.add_gate(tk.OpType.PhasedX, [0.5, 0.5], [0])
c.add_gate(tk.OpType.PhasedX, [0.5, 0.5], [0])
```

The `SquashHQS` pass should merge these two gates into one, but it does not:

```
from pytket.passes import SquashHQS

SquashHQS().apply(c)
```
gives `[PhasedX(0.5, 0.5) q[0]; PhasedX(0.5, 0.5) q[0]; ]`. Expected: `[PhasedX(1, 0.5) q[0]; ]`.

PR #148, opened 2021-12-08
[bugfix] Fixing #147

lmondada:


tket/src/Transformations/Rebase.cpp
```diff
@@ -118,10 +118,18 @@ Transform Transform::rebase_tket() {
...
-    c.add_op<unsigned>(OpType::PhasedX, px_params, {0});
+  if (equiv_expr(beta, 1)) {
+    // Angles β ∈ {π, 3π}
+    c.add_op<unsigned>(OpType::PhasedX, {beta, (alpha - gamma) / 2.}, {0});
+  } else if (equiv_expr(beta, 2, 4)) {
+    // Angle β = 2π
+    c.add_op<unsigned>(OpType::Rz, alpha + gamma + 2, {0});
+  } else if (equiv_0(beta, 4)) {
+    // Angle β = 0
+    c.add_op<unsigned>(OpType::Rz, alpha + gamma, {0});
```
cqc-alec, line 129:
```suggestion
  } else if (equiv_expr(beta, 0)) {
    // Angle β ∈ {0, 2π}
    c.add_op<unsigned>(OpType::Rz, alpha + beta + gamma, {0});
```
I think we can combine these cases into one like this?

cqc-alec (changes requested):

cqc-alec (approved):

PR #151, opened 2021-12-13
Clifford Tableau refactor and boxes

willsimmons1465:
Rewritten `CliffTableau` to `UnitaryTableau` to more closely resemble the unitary stabilizer tableau of Aaronson&Gottesman, where rows indicate the action at the output caused by a Pauli at the input rather than vice versa. This is not intended to be fast and effective as stim, but will be a useful data structure for compilation. `UnitaryTableauBox` encapsulates this into an `Op` for embedding into a `Circuit`.

`StabilizerTableau` has been added to more easily represent states which will have more benefit in future for (measurement/rotation) diagonalisation and state preparation, and is used within `UnitaryTableau` to handle the underlying tableau.

`CliffTableau` remains for now, as there are several places in code where it is currently useful to use the reverse lookup and will require further refactoring to eliminate in the neatest way, e.g. change `PauliGraph` from having the tableau at the end to the tableau at the start.


tket/src/Clifford/UnitaryTableau.hpp
```diff
@@ -0,0 +1,172 @@
...
+#include "SymplecticTableau.hpp"
+
+namespace tket {
+
+// Forward declare Circuit for friend converter
+class Circuit;
+
+class UnitaryTableau {
+  /**
+   * An implementation of the stabilizer-destabilizer tablea for unitary
```
sjdilkes, line 26: tableau*

tket/src/Converters/UnitaryTableauBox.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+namespace tket {
+
+class UnitaryTableauBox : public Box {
+ public:
+  explicit UnitaryTableauBox(const UnitaryTableau& tab);
+  explicit UnitaryTableauBox(
+      const MatrixXb& xx, const MatrixXb& xz, const VectorXb& xph,
+      const MatrixXb& zx, const MatrixXb& zz, const VectorXb& zph);
+
```
sjdilkes, line 28: docstrings for methods in this file

sjdilkes (changes requested):
Only doc changes, code all looks good. 
Also - please can you update the copyright to `2022` (sorry I didn't get this done before the end of the year!).

sjdilkes (approved):

PR #158, opened 2021-12-17
[TKET-1670] Expose free symbols

Roland-djee:
Fixes [TKET-1670](https://cqc.atlassian.net/browse/TKET-1670?atlOrigin=eyJpIjoiM2ViZTZjNWQ4MmY3NGQwYThjMTQ4MThkNTAwMzM3NWYiLCJwIjoiaiJ9).

This PR addresses exposing the `free_symbols` method from the underlying gate at the op level in Python. 
It also adds an `pybind11` custom exception to be thrown in Python if the method is not defined for that gate. 

However, local checks return an empty set in that case.

### DIFF
--- a/pytket/binders/circuit/main.cpp
+++ b/pytket/binders/circuit/main.cpp
@@ -31,6 +31,15 @@
 
 namespace py = pybind11;
 
+// A type that should be raised as an exception in Python
+class OpException : public std::exception {
+public:
+    explicit OpException(const char * m) : message{m} {}
+    const char * what() const noexcept override {return message.c_str();}
+private:
+    std::string message = "";
+};
+
 namespace tket {
 
 void init_unitid(py::module &m);
--- a/pytket/binders/circuit/main.cpp
+++ b/pytket/binders/circuit/main.cpp
@@ -74,7 +74,7 @@ PYBIND11_MODULE(circuit, m) {
           py::arg("latex") = false)
       .def("__eq__", &Op::operator==)
       .def("__repr__", [](const Op &op) { return op.get_name(); })
-      .def("free_symbols", [](const Op &op) { return op->free_symbols(); })
+      .def("free_symbols", [](const Op &op) { return op.free_symbols(); })
       .def(
           "get_unitary",
           [](const Op *op) {
--- a/pytket/binders/circuit/main.cpp
+++ b/pytket/binders/circuit/main.cpp
@@ -74,9 +74,7 @@ PYBIND11_MODULE(circuit, m) {
           py::arg("latex") = false)
       .def("__eq__", &Op::operator==)
       .def("__repr__", [](const Op &op) { return op.get_name(); })
-      .def(
-          "free_symbols",
-          [](const Op *op) { return op->free_symbols(); })
+      .def("free_symbols", [](const Op *op) { return op->free_symbols(); })
       .def(
           "get_unitary",
           [](const Op *op) {
--- a/pytket/binders/circuit/main.cpp
+++ b/pytket/binders/circuit/main.cpp
@@ -31,7 +31,6 @@
 
 namespace py = pybind11;
 
-
 namespace tket {
 
 void init_unitid(py::module &m);
@@ -82,7 +81,8 @@ PYBIND11_MODULE(circuit, m) {
             try {
               symbols = op->free_symbols();
             } catch (const std::runtime_error &e) {
-              throw std::runtime_error("Free symbols are not defined for that Op.");
+              throw std::runtime_error(
+                  "Free symbols are not defined for that Op.");
             }
             return symbols;
           })
--- a/pytket/binders/circuit/main.cpp
+++ b/pytket/binders/circuit/main.cpp
@@ -33,11 +33,12 @@ namespace py = pybind11;
 
 // A type that should be raised as an exception in Python
 class OpException : public std::exception {
-public:
-    explicit OpException(const char * m) : message{m} {}
-    const char * what() const noexcept override {return message.c_str();}
-private:
-    std::string message = "";
+ public:
+  explicit OpException(const char *m) : message{m} {}
+  const char *what() const noexcept override { return message.c_str(); }
+
+ private:
+  std::string message = "";
 };
 
 namespace tket {
@@ -51,12 +52,12 @@ PYBIND11_MODULE(circuit, m) {
   init_unitid(m);
   static py::exception<OpException> ex(m, "OpException");
   py::register_exception_translator([](std::exception_ptr p) {
-      try {
-          if (p) std::rethrow_exception(p);
-      } catch (const OpException &e) {
-          // Set OpException as the active python error
-          ex(e.what());
-      }
+    try {
+      if (p) std::rethrow_exception(p);
+    } catch (const OpException &e) {
+      // Set OpException as the active python error
+      ex(e.what());
+    }
   });
   py::class_<Op, std::shared_ptr<Op>>(
       m, "Op", "Encapsulates operation information")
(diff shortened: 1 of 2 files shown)


Roland-djee:
Hi @cqc-alec I think this is roughly in line with what we discussed. However, and as described in the preamble, if I run `free_symbols` for a gate that presumably doesn't support it, it returns an empty set:

```
>>> from pytket import Circuit
>>> c = Circuit(2)
>>> c.add_barrier([0, 1])
[Barrier q[0], q[1]; ]
>>> op = c.get_commands()[0].op
>>> op.free_symbols()
set()
```

I am happy for suggestions to test that case.

pytket/binders/circuit/main.cpp
```diff
@@ -31,6 +31,16 @@
 
 namespace py = pybind11;
 
+// A type that should be raised as an exception in Python
+class OpException : public std::exception {
```
cqc-alec, line 35: This feels quite ad-hoc. We do need to improve exception handling in the Python layer but for now we just let exceptions propagate from C++ and I think we should stick with that, rather than do anything special for this function.

pytket/binders/circuit/main.cpp
```diff
@@ -74,6 +93,18 @@ PYBIND11_MODULE(circuit, m) {
           py::arg("latex") = false)
       .def("__eq__", &Op::operator==)
       .def("__repr__", [](const Op &op) { return op.get_name(); })
+      .def(
+          "free_symbols",
+          [](const Op *op) {
+            SymSet symbols;
+            const auto &gate = static_cast<const Gate &>(*op);
```
cqc-alec, line 100: I don't see why we need to cast to a `Gate` (and the `static_cast` is dangerous anyway). The `free_symbols()` method is defined for any `Op`; it will just throw an exception in some cases.

cqc-alec (changes requested):
This is over-complicated: all we want is a simple pass-through to `Op::free_symbols()`.
Please add some unit tests too, covering cases where: a gate has some symbolic parameters; a gate does not have any symbolic parameters; a box has symbolic parameters; the method throws.

Roland-djee:
> This is over-complicated: all we want is a simple pass-through to `Op::free_symbols()`. Please add some unit tests too, covering cases where: a gate has some symbolic parameters; a gate does not have any symbolic parameters; a box has symbolic parameters; the method throws.

OK. As for the tests, I actually checked manually from Python (see above) that gates with symbolic parameters actually do return the symbols and that gates without do return an empty set. I was on the lookout for a case that would have made the exception thrown.

pytket/binders/circuit/main.cpp
Roland-djee: Does it ? I felt it would serve later as a model for custom exceptions in the bindings that would be specific. But OK I can remove it.

pytket/binders/circuit/main.cpp
Roland-djee: Actually that was my initial solution. But I felt it made more sense with the custom exception. On hindsight, this is less generic and would throw another error than the one we are interested in. Will revert.

pytket/binders/circuit/main.cpp
cqc-alec: I think what we really need is (a) improved exceptions in the C++ and (b) systematically translating these into Python exceptions. I'll make sure we have captured these tasks somewhere.

Roland-djee: Well, isn't it what this does ?

cqc-alec: Not exactly. This handles a particular case, but I don't want to have to add special exception-handling code in every pytket method, that would quickly become unmaintainable. Need some generic way to capture all relevant information in the C++ exception and pass it through to Python.

Roland-djee: OK if you're after genericity. But why not having a central error file that contains all variants and use the `pybind11` error registration machinery to translate them adequately to Python errors ? Just an idea really.

Roland-djee:
@cqc-alec Actually, since `free_symbols()` is defined for any `Op`, won't it just return either a set of symbols or an empty set ? I can't really figure out a case for which the exception is thrown. Any ideas maybe ?

cqc-alec:
> @cqc-alec Actually, since `free_symbols()` is defined for any `Op`, won't it just return either a set of symbols or an empty set ? I can't really figure out a case for which the exception is thrown. Any ideas maybe ?

For example an `Input`, `Output` or `Barrier` op should throw an exception.

Roland-djee:
> > @cqc-alec Actually, since `free_symbols()` is defined for any `Op`, won't it just return either a set of symbols or an empty set ? I can't really figure out a case for which the exception is thrown. Any ideas maybe ?
> 
> For example an `Input`, `Output` or `Barrier` op should throw an exception.

`Barrier` was my primary test for that case but it currently returns an empty set:
```
>>> from pytket import Circuit
>>> c = Circuit(2)
>>> from pytket import OpType
>>> c.add_gate(OpType.Barrier, [], [0, 1])
  File "<stdin>", line 1, in <module>
RuntimeError: Cannot add metaop. Please use `add_barrier` to add a barrier.
>>> c.add_barrier([0, 1])
[Barrier q[0], q[1]; ]
>>> op = c.get_commands()[0].op
>>> op
Barrier
>>> op.free_symbols()
set()
```

cqc-alec:
> > > @cqc-alec Actually, since `free_symbols()` is defined for any `Op`, won't it just return either a set of symbols or an empty set ? I can't really figure out a case for which the exception is thrown. Any ideas maybe ?
> > 
> > 
> > For example an `Input`, `Output` or `Barrier` op should throw an exception.

Ah you're right, it does return the empty set there. Maybe there are no exception cases after all.

Roland-djee:
> > > > @cqc-alec Actually, since `free_symbols()` is defined for any `Op`, won't it just return either a set of symbols or an empty set ? I can't really figure out a case for which the exception is thrown. Any ideas maybe ?
> > > 
> > > 
> > > For example an `Input`, `Output` or `Barrier` op should throw an exception.
> 
> Ah you're right, it does return the empty set there. Maybe there are no exception cases after all.

Yeah, it seems like the behaviour is to return an empty set when there are no symbols and a filled set otherwise. BTW correct me if I'm wrong but it seems that `Input`and `Output` aren't bound in Python. I will use `Barrier` in the test for the non-symbols op case.

pytket/binders/circuit/main.cpp
```diff
@@ -74,6 +74,7 @@ PYBIND11_MODULE(circuit, m) {
           py::arg("latex") = false)
       .def("__eq__", &Op::operator==)
       .def("__repr__", [](const Op &op) { return op.get_name(); })
+      .def("free_symbols", [](const Op *op) { return op->free_symbols(); })
```
cqc-alec, line 77: Can we pass by const reference rather than address, unless there's some reason why this wouldn't work?

Roland-djee: No reason. Done. Out of curiosity, or maybe just a naive question. Why is a reference preferred ? Because it offers more security guarantees than a pointer ?

cqc-alec: In this case it probably makes very little difference, but in general, passing pointers around is less safe than passing references (in the sense that bugs leading to segfaults or UB are more likely), so for preference we should pass by reference.

Roland-djee: Thanks. Yeah that's what I've been reading around. Mostly the unsafeness around the nullability of pointers.

cqc-alec (approved):

PR #161, opened 2022-01-05
Release/0.18.0

cqc-alec:


pytket/docs/conf.py
```diff
@@ -38,9 +38,9 @@
 author = "Cambridge Quantum Computing Ltd"
```
cqc-melf, line 38: Should this be updated?

cqc-melf (commented):
The year in the copyright comments are not updated yet, is that important?

Should we change the author and the licence from CQC to CQ?

cqc-alec:
> The year in the copyright comments are not updated yet, is that important?
> 
> Should we change the author and the licence from CQC to CQ?

It doesn't matter about the year -- we can change that later in a single commit (though we could also update it as and when files are modified). I'm pretty sure it has no legal force.

I think the change from CQC to CQ was just a branding thing, not a legal thing.

For the author, I'll wait for legal advice on that (but won't hold my breath!).

cqc-melf (approved):

PR #163, opened 2022-01-06
Add u and p gate aliases to match updated qelib1.inc

DomNomNom:
Context: https://github.com/Qiskit/qiskit-terra/pull/4765

Before this change the following code would fail
```py
from qiskit import QuantumCircuit

qc = QuantumCircuit(1)
qc.p(0, 0)
qc.u(0, 0, 0, 0)
print(circuit_from_qasm_str(qc.qasm()))
```

With a message like: `pytket.qasm.qasm.QASMParseError: Cannot parse gate of type: p`

### DIFF
--- a/pytket/pytket/qasm/qasm.py
+++ b/pytket/pytket/qasm/qasm.py
@@ -17,14 +17,28 @@
 import io
 import os
 from itertools import groupby
-from typing import (Any, Callable, Dict, List, Optional, TextIO, Tuple, Type,
-                    TypeVar, Union)
-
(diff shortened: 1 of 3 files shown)


cqc-alec (changes requested):
Thanks, looks good. Could you please add a test in `pytket/tests/qasm_test.py` (with a new qasm test file containing these gates) to confirm we can parse them correctly?

DomNomNom:
Done - added a test.

cqc-alec (approved):

cqc-alec:
Thank you!

PR #169, opened 2022-01-13
Support Python 3.10; drop 3.7.

cqc-alec:
* Fix how we import Python modules in the binders to avoid segfault on exit.
* Re-apply changes from #113.
* A few mypy fixes.

Python 3.10 tested on CI [here](https://github.com/CQCL/tket/actions/runs/1692829796).


pytket/binders/circuit/Circuit/add_op.cpp
```diff
@@ -76,7 +70,8 @@ static Circuit *add_gate_method(
         "`condition_value` specified without `condition_bits`");
   }
   if (condition_given) {
-    py::object add_condition = condition_module().attr("_add_condition");
+    py::module condition = py::module::import("pytket.circuit.add_condition");
```
ss2165, line 73: do we know if this is slower than importing once only?

cqc-alec: I did some very informal timing of tests and didn't notice any significant difference. Under the hood this calls `PyModule_Import` which I think must be more or less a no-op if the module is already imported.

ss2165 (approved):

PR #172, opened 2022-01-14
Feature/decompose cnx with gray code

yao-cqc:
Added a Gray code method for cnx decomposition.

### DIFF
--- a/pytket/docs/changelog.rst
+++ b/pytket/docs/changelog.rst
@@ -1,6 +1,13 @@
 Changelog
 ==================================
 
+0.19.0 (unreleased)
+-------------------
+
+Minor new features:
+
+* Add ``cnx_gray_decomp`` method for C5X, C6X and C7X decomposition.
+
 0.18.0 (January 2022)
 ---------------------
 
--- a/pytket/docs/changelog.rst
+++ b/pytket/docs/changelog.rst
@@ -6,7 +6,7 @@ Changelog
 
 Minor new features:
 
-* Add ``cnx_gray_decomp`` method for C5X, C6X and C7X decomposition.
+* Improve ``CnX`` gate decomposition for n=5,6,7.
 
 General:
 
--- a/tket/tests/test_ControlDecomp.cpp
+++ b/tket/tests/test_ControlDecomp.cpp
@@ -408,21 +408,27 @@ SCENARIO("Test a CnX is decomposed correctly using the Gray code method") {
       switch (n) {
         case 2: {
           REQUIRE(circ.count_gates(OpType::CX) <= 6);
+          break;
         }
         case 3: {
           REQUIRE(circ.count_gates(OpType::CX) <= 14);
+          break;
         }
         case 4: {
           REQUIRE(circ.count_gates(OpType::CX) <= 36);
+          break;
         }
         case 5: {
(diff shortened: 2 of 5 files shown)


pytket/docs/changelog.rst
```diff
@@ -4,6 +4,10 @@ Changelog
 0.19.0 (unreleased)
 -------------------
 
+Minor new features:
+
+* Add ``cnx_gray_decomp`` method for C5X, C6X and C7X decomposition.
```
cqc-alec, line 9: This isn't a pytket method so the wording isn't quite right here, but we could say something about improvements to CnX decomposition for n >= 5.

tket/tests/test_ControlDecomp.cpp
```diff
@@ -385,5 +385,22 @@ SCENARIO("Test a CnX is decomposed correctly when bootstrapped") {
   }
 }
 
+SCENARIO("Test a CnX is decomposed correctly using the Gray code method") {
+  GIVEN("Test CnX unitary for 3 to 8 controls") {
+    for (unsigned n = 3; n < 8; ++n) {
```
cqc-alec, line 390: Should we test n=0,1,2 as well, as the function accepts those?

tket/tests/test_ControlDecomp.cpp
```diff
@@ -385,5 +385,22 @@ SCENARIO("Test a CnX is decomposed correctly when bootstrapped") {
...
+          Eigen::MatrixXcd::Identity(m_size, m_size);
+      correct_matrix(m_size - 2, m_size - 1) = 1;
+      correct_matrix(m_size - 1, m_size - 2) = 1;
+      correct_matrix(m_size - 2, m_size - 2) = 0;
+      correct_matrix(m_size - 1, m_size - 1) = 0;
+      REQUIRE(m.isApprox(correct_matrix, ERR_EPS));
+    }
+  }
+}
+
```
cqc-alec, line 404: We could also add checks that the number of CX gates in the decomposed circuits are <= (whatever). That would prevent regressions if we rework the methods again.

cqc-alec (changes requested):

tket/tests/test_ControlDecomp.cpp
```diff
@@ -398,6 +405,26 @@ SCENARIO("Test a CnX is decomposed correctly using the Gray code method") {
       correct_matrix(m_size - 2, m_size - 2) = 0;
       correct_matrix(m_size - 1, m_size - 1) = 0;
       REQUIRE(m.isApprox(correct_matrix, ERR_EPS));
+      switch (n) {
+        case 2: {
+          REQUIRE(circ.count_gates(OpType::CX) <= 6);
```
cqc-alec, line 410: Please add `break` statements at the end of each `case`. (As it happens fall-through means we still pass because they're in order, but is unnecessary!)

cqc-alec (changes requested):

tket/tests/test_ControlDecomp.cpp
yao-cqc: Ahh, sorry, I'll make the change now!

cqc-alec (approved):

PR #174, opened 2022-01-17
Modularization

cqc-alec:
The main purpose of this PR is to provide proper separation between the various tket components, in terms of interdependencies and public versus private headers. Each has its own `CMakeLists` file which details its dependencies among the other components.

The most obvious change is the creation of a public `include` directory within each component, where header files intended for consumption by other components, or libraries and programs depending on tket, reside. Header files needed only within the component stayed where they were, alongside the cpp files.

It turned out that most of the header files needed to be public, but as a follow-up task we should do some refactoring here to only expose what we need to.

Each component is built as a shared library.

There is some clean-up of the cmake files.

There is also some clean-up of the conanfiles, preparatory to publishing tket on the conan-center. This work is not quite completed because I encountered an annoying issue on Windows when I tried to make the test_package conanfile conformant with the conan-center requitements, but we can leave that for later.

Build-time improvements are pretty small (only really noticeable on MacOS), but hopefully this is a stepping-stone to further improvements (which will involve more invasive code changes).


tket/src/Transformations/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+    ${TKET_Circuit_INCLUDE_DIR}
+    ${TKET_Graphs_INCLUDE_DIR}
+    ${TKET_PauliGraph_INCLUDE_DIR}
+    ${TKET_Converters_INCLUDE_DIR}
+    ${TKET_Characterisation_INCLUDE_DIR}
+    ${TKET_Architecture_INCLUDE_DIR})
+
+target_link_libraries(tket-Transformations
+    PRIVATE
+    tket-Utils
```
ss2165, line 52: is it reasonable to do `target_include_directories` and `target_include_directories` within a for loop like the way the environment variables for each module are originally set, looping over the module names

ss2165: It would make it to verify the include and linked are appropriately correlated

cqc-alec: Good idea.

PR #180, opened 2022-01-24
Explicit linking of library dependencies

cqc-alec:
A few changes are rolled together here:
* Update to boost 1.78.0.
* Install boost with all-default options on M1. (Previous build issues seem to have been fixed.)
* Specify target link libraries explicitly, rather than using `CONAN_LIBS` which includes many unnecessary libraries.
* Remove restriction on build parallelism on Windows.
* Eliminate the dependency on `boost::chrono` by modifying our boost patch to use `std::chrono`. (The original version used `boost::chrono` because it was intended as a patch to boost, which needs to build without `std::chrono`, but that PR has since been dropped and replaced by another one without this dependency.)

### DIFF
--- a/pytket/CMakeLists.txt
+++ b/pytket/CMakeLists.txt
@@ -38,13 +38,7 @@ else()
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Werror")
 endif()
 
-function(build_module module)
-    pybind11_add_module(${module} ${ARGN})
-    target_include_directories(${module} PRIVATE binders/include)
-    target_link_libraries(${module} PRIVATE ${CONAN_LIBS})
-endfunction(build_module)
-
-build_module(circuit
+pybind11_add_module(circuit
     binders/circuit/main.cpp
     binders/circuit/unitid.cpp
     binders/circuit/boxes.cpp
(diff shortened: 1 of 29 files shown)


pytket/CMakeLists.txt
```diff
@@ -38,31 +38,82 @@ else()
...
-build_module(passes binders/passes.cpp)
-build_module(program binders/program.cpp)
-build_module(partition binders/partition.cpp)
-build_module(pauli binders/pauli.cpp)
-build_module(logging binders/logging.cpp)
-build_module(utils_serialization binders/utils_serialization.cpp)
-build_module(tailoring binders/tailoring.cpp)
-build_module(tableau binders/tableau.cpp)
-build_module(zx
+target_include_directories(circuit PRIVATE binders/include)
```
cqc-melf, line 49: Have you thought about naming all the python packages something like `pytket-circuit` to make clear that they are self defined?

pytket/CMakeLists.txt
```diff
@@ -38,31 +38,82 @@ else()
...
+target_link_libraries(circuit PRIVATE
+    symengine
+    tket-Converters
+    tket-Routing
+    tket-Simulation)
+
+pybind11_add_module(routing binders/routing.cpp)
```
cqc-melf, line 65: Could be `pytket-routing`, see above

cqc-melf (commented):
Hi Alec,
Thank you for splitting the PR in the different commits, that was really helpful to understand what you have done.

I only have one small remark, but I am not sure if this rename is possible at all.

pytket/CMakeLists.txt
cqc-alec: This would mean changing the names of the Python modules we expose, e.g. `pytket.circuit` would become `pytket.pytket-circuit`.

At least the names are lower-case so not identical to the tket component names.

cqc-melf (approved):

cqc-melf:
For my side this is fine now.

PR #181, opened 2022-01-25
Feature/expose is_clifford to pytket Op

yao-cqc:
The logic is copied from https://github.com/CQCL/tket/blob/e2a2bb86d8728b5ed3dd4cc06d9bb2b0b8d62a38/tket/src/Predicates/Predicates.cpp#L433

### DIFF
--- a/pytket/binders/circuit/main.cpp
+++ b/pytket/binders/circuit/main.cpp
@@ -82,14 +82,8 @@ PYBIND11_MODULE(circuit, m) {
             return gate.get_unitary();
           })
       .def(
-          "is_clifford",
-          [](const Op &op) {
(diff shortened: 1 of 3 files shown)


pytket/binders/circuit/main.cpp
```diff
@@ -81,6 +81,15 @@ PYBIND11_MODULE(circuit, m) {
             const auto &gate = static_cast<const Gate &>(*op);
             return gate.get_unitary();
           })
+      .def(
+          "is_clifford",
+          [](const Op &op) {
+            if (op.get_desc().is_gate() && !op.get_desc().is_clifford_gate() &&
```
cqc-alec, line 87: This method will return `true` if the operation is not a gate. Is that what we want?

PR #182, opened 2022-01-26
Move transformations outside `Transform` class and into `Transforms` namespace

cqc-alec:


tket/src/Transformations/include/Transformations/ContextualReduction.hpp
```diff
@@ -71,7 +71,7 @@ Transform remove_discarded_ops();
 Transform simplify_initial(
     AllowClassical allow_classical = AllowClassical::Yes,
     CreateAllQubits create_all_qubits = CreateAllQubits::No,
-    std::shared_ptr<const Circuit> = 0);
+    std::shared_ptr<const Circuit> xcirc = 0);
```
drzenharper, line 74: should really be nullptr rather than 0

drzenharper (approved):

PR #185, opened 2022-01-28
Infra/use bimap for quantum boundary

yao-cqc:


sjdilkes (approved):
Excellent stuff, all looks good to me.

PR #186, opened 2022-01-31
Move "unit bimaps" out of the `Circuit` class and let `Transform` handle them.

### LINKED ISSUE #137: Move `unit_bimaps_` out of the `Circuit` class
These are housekeeping data that should be tracked by compilation passes, not in the `Circuit` class itself.

cqc-alec:
Closes #137 .


sjdilkes (approved):
Looks good. I am minorly concerned about how this plays with the ancilla qubits added by RV3, but suspect the best way to investigate this is to merge this, then merge into feature/routing-v3 and find out.

PR #190, opened 2022-02-01
Support `rxx` gate for `pytket_qasm`

vtomole:
I'm having some hiccups building tket locally so I'll open this as a draft to see if the CI passes with this commit.

PR #193, opened 2022-02-02
[feature] add RZZ qasm output support for hqslib

ss2165:


cqc-alec:
I wonder if we should add this to the main qasm parser instead. See also #190 . It seems that qiskit accepts both `rxx` and `rzz` as valid QASM:
```python
circ = qiskit.QuantumCircuit.from_qasm_str("""OPENQASM 2.0;
include "qelib1.inc";
qreg q[2];
creg c[2];
rzz(pi / 2) q[0],q[1];
measure q -> c;
""")
```

cqc-alec (approved):

Issue #97, opened 2021-10-26
ConnectivityPredicate fails after FullPeepholeOptimise and AASRouting


cqc-melf:
The problem can be generated with:

```from pytket.qasm import circuit_from_qasm
from pytket.predicates import ConnectivityPredicate
from pytket.passes import (
    FullPeepholeOptimise,
    AASRouting,
    SequencePass,
)
from pytket.routing import Architecture
from pytket import Circuit
import json

coupling_map = [[0, 1], [0, 14], [1, 0], [1, 2], [1, 13], [2, 1], [2, 3], [2, 12], [3, 2], [3, 4], [3, 11], [4, 3], [4, 5], [4, 10], [5, 4], [5, 6], [5, 9], [6, 5], [6, 8], [7, 8], [8, 6], [8, 7], [8, 9], [9, 5], [9, 8], [9, 10], [10, 4], [10, 9], [10, 11], [11, 3], [11, 10], [11, 12], [12, 2], [12, 11], [12, 13], [13, 1], [13, 12], [13, 14], [14, 0], [14, 13]]
device = Architecture(coupling_map)
connectivity_pred = ConnectivityPredicate(device)

with open('aas_connectivity_pred.json', 'r') as fp:
    orig_circ = Circuit().from_dict(json.load(fp))
    
new_circ_pytket_aas = orig_circ.copy()

FullPeepholeOptimise(allow_swaps=True).apply(new_circ_pytket_aas)
AASRouting(device).apply(new_circ_pytket_aas) 

assert connectivity_pred.verify(new_circ_pytket_aas)```

cqc-melf:
maybe related: https://github.com/CQCL/tket/issues/126

cqc-melf:
This is solved now.

PR #194, opened 2022-02-02
[TKET-1598] fix ConnectivityPredicate fails after FullPeepholeOptimise and AASRouting

cqc-melf:
This is fixing https://github.com/CQCL/tket/issues/97

The problem has been a wrong handling of the implicit wire swaps in the aas pass

### DIFF
--- a/tket/src/Predicates/include/Predicates/PassGenerators.hpp
+++ b/tket/src/Predicates/include/Predicates/PassGenerators.hpp
@@ -61,7 +61,7 @@ PassPtr gen_directed_cx_routing_pass(
 /**
  * execute architecture aware synthesis on a given architecture for an allready
  * place circuit, only for circuit which contains Cx+Rz+H gates
- * this pass it not able to handle implicit wire swaps
+ * this pass is not able to handle implicit wire swaps
  * @param arc architecture to route on
  * @param lookahead parameter for the recursion depth in the algorithm, the
  * value should be > 0
--- a/tket/src/Predicates/include/Predicates/PassGenerators.hpp
+++ b/tket/src/Predicates/include/Predicates/PassGenerators.hpp
@@ -61,6 +61,7 @@ PassPtr gen_directed_cx_routing_pass(
 /**
  * execute architecture aware synthesis on a given architecture for an allready
  * place circuit, only for circuit which contains Cx+Rz+H gates
+ * this pass it not able to handle implicit wire swaps
  * @param arc architecture to route on
  * @param lookahead parameter for the recursion depth in the algorithm, the
  * value should be > 0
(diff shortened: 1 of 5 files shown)


tket/src/Predicates/include/Predicates/PassGenerators.hpp
```diff
@@ -61,6 +61,7 @@ PassPtr gen_directed_cx_routing_pass(
 /**
  * execute architecture aware synthesis on a given architecture for an allready
  * place circuit, only for circuit which contains Cx+Rz+H gates
+ * this pass it not able to handle implicit wire swaps
```
alexcowtan, line 64:
```suggestion
 * this pass is not able to handle implicit wire swaps
```

tket/src/Transformations/Decomposition.cpp
```diff
@@ -746,6 +746,21 @@ Transform decomp_boxes() {
...
+      qubit_map_t perm = circ.implicit_qubit_permutation();
+      for (const std::pair<const Qubit, Qubit> &pair : perm) {
+        if (pair.first != pair.second) {
+          if (!foundswap) {
+            circ.replace_implicit_wire_swap(pair.first, pair.second);
+            foundswap = true;
+          }
+        }
+      }
+    }
```
alexcowtan, line 762: Is the `foundswap` variable really necessary? Can you not do something like:
```suggestion
    while (circ.has_implicit_wireswaps()) {

      qubit_map_t perm = circ.implicit_qubit_permutation();
      for (const std::pair<const Qubit, Qubit> &pair : perm) {
        if (pair.first != pair.second) {
            circ.replace_implicit_wire_swap(pair.first, pair.second);
          }
        }
      }
    }
```

alexcowtan (changes requested):

tket/src/Transformations/Decomposition.cpp
cqc-melf: Without the variable only one swap will create an infinity loop. The replace_implicit_wire_swap will remove one implicit wire swap and then add it again in the next iteration of the loop where pair.first != pair.second.

tket/src/Predicates/include/Predicates/PassGenerators.hpp
cqc-melf: Thank you for spotting this!

tket/src/Transformations/Decomposition.cpp
alexcowtan: Ah, okay. Would it be slightly faster then to have a `break` rather than keep iterating through the loop but not replacing the swaps?

cqc-melf: I have added that.

alexcowtan (approved):
Looks good to me.

PR #199, opened 2022-02-03
[RV3] [refactor] Clean up of the dependencies of the modules

cqc-melf:
Clean up of the dependencies of the modules, this should help to avoid loops while adding new modules.


sjdilkes (approved):

sjdilkes (approved):
Go for it

Issue #178, opened 2022-01-19
Occasional build failures on Windows CI due to heap exhaustion


cqc-alec:
I thought setting `CMAKE_BUILD_PARALLEL_LEVEL=2` in the environment when building on the CI had fixed this issue, but it seems to still occur sometimes: https://github.com/CQCL/tket/runs/4867491927

```
C:\Users\runneradmin\.conan\data\nlohmann_json\3.10.4\_\_\package\5ab84d6acfe1f23c4fae0ab88f26e3a396351ac9\include\nlohmann/json.hpp(930): fatal error C1060: compiler is out of heap space (compiling source file C:\Users\runneradmin\.conan\data\tket\1.0.1\_\_\build\5c7a2862abf41a4b4e1c2c6362efbc737e8bbe8a\Characterisation\FrameRandomisation.cpp) [C:\Users\runneradmin\.conan\data\tket\1.0.1\_\_\build\5c7a2862abf41a4b4e1c2c6362efbc737e8bbe8a\Characterisation\tket-Characterisation.vcxproj]
C:\.conan\8f95b3\1\include\boost/multiprecision/cpp_int.hpp(1293,1): fatal error C1060: compiler is out of heap space (compiling source file C:\Users\runneradmin\.conan\data\tket\1.0.1\_\_\build\5c7a2862abf41a4b4e1c2c6362efbc737e8bbe8a\Diagonalisation\Diagonalisation.cpp) [C:\Users\runneradmin\.conan\data\tket\1.0.1\_\_\build\5c7a2862abf41a4b4e1c2c6362efbc737e8bbe8a\Diagonalisation\tket-Diagonalisation.vcxproj]
C:\Users\runneradmin\.conan\data\nlohmann_json\3.10.4\_\_\package\5ab84d6acfe1f23c4fae0ab88f26e3a396351ac9\include\nlohmann/detail/input/json_sax.hpp(591,1): fatal error C1060: compiler is out of heap space (compiling source file C:\Users\runneradmin\.conan\data\tket\1.0.1\_\_\build\5c7a2862abf41a4b4e1c2c6362efbc737e8bbe8a\Characterisation\Cycles.cpp) [C:\Users\runneradmin\.conan\data\tket\1.0.1\_\_\build\5c7a2862abf41a4b4e1c2c6362efbc737e8bbe8a\Characterisation\tket-Characterisation.vcxproj]
[...]

PR #200, opened 2022-02-03
Use spdlog library

cqc-alec:
Go back to using the library version of spdlog/fmt, except when building Linux wheels.
This mostly reverts #44 . It should make builds a bit faster, and _may_ solve #178 (we'll see).

### DIFF
--- a/pytket/CMakeLists.txt
+++ b/pytket/CMakeLists.txt
@@ -38,6 +38,9 @@ else()
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Werror")
 endif()
 
+list(APPEND TKET_EXTRA_LIBS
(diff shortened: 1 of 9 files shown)


pytket/CMakeLists.txt
```diff
@@ -56,7 +56,8 @@ target_link_libraries(circuit PRIVATE
     tket-Routing
     tket-Simulation
     tket-Utils)
-target_link_libraries(circuit PRIVATE ${CONAN_LIBS_SYMENGINE})
+target_link_libraries(circuit PRIVATE
```
cqc-melf, line 59: Have you thought about defining a variable for the libraries that this only has be be changed in one line for the hole file in future changes?

cqc-alec: I will do that.

cqc-alec: Done.

Issue #192, opened 2022-02-02
Dictionary returned by `BackendInfo.to_dict()` is not JSON-serializable


cqc-alec:
The [documentation](https://cqcl.github.io/tket/pytket/api/backends.html#pytket.backends.backendinfo.BackendInfo.to_dict) says that it is. But:

```python
from pytket.extensions.qiskit import IBMQBackend, IBMQEmulatorBackend
import json

device_backend = IBMQEmulatorBackend("ibmq_santiago", hub="ibm-q", group="open", project="main")
backend_info_json = device_backend.backend_info.to_dict()
json.dumps(backend_info_json)
```

produces

```
TypeError: keys must be str, int, float, bool or None, not pytket._tket.circuit.Node
```

PR #201, opened 2022-02-04
Ensure that the standard fields of `BackendInfo` are JSON-serializable

cqc-alec:
This partially addresses #192 . The remaining issue is for the `misc` field, which is https://github.com/CQCL/pytket-extensions/issues/288 .

### DIFF
--- a/pytket/pytket/backends/backendinfo.py
+++ b/pytket/pytket/backends/backendinfo.py
@@ -13,11 +13,101 @@
 # limitations under the License.
 
 """ BackendInfo class: additional information on Backends """
+
+from ast import literal_eval
 from dataclasses import dataclass, field, asdict
 from typing import Any, Dict, List, Optional, Set, cast, Tuple, Union
 
 from pytket.routing import Architecture, FullyConnected  # type: ignore
 from pytket.circuit import Node, OpType  # type: ignore
+
+
+def serialize_all_node_gate_errors(d):
+    if d is None:
+        return None
+    return {
+        str(n.to_list()): {ot.value: err for ot, err in errs.items()}
+        for n, errs in d.items()
+    }
+
+
+def deserialize_all_node_gate_errors(d):
+    if d is None:
+        return None
+    return {
+        Node.from_list(literal_eval(n)): {
+            OpType(int(ot)): err for ot, err in errs.items()
+        }
+        for n, errs in d.items()
+    }
+
+
+def serialize_all_edge_gate_errors(d):
+    if d is None:
+        return None
+    return {
+        str((n0.to_list(), n1.to_list())): {ot.value: err for ot, err in errs.items()}
+        for (n0, n1), errs in d.items()
+    }
+
+
+def deserialize_all_edge_gate_errors(d):
+    if d is None:
+        return None
+    return {
+        tuple(map(Node.from_list, literal_eval(n))): {
+            OpType(int(ot)): err for ot, err in errs.items()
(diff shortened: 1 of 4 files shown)


pytket/pytket/backends/backendinfo.py
```diff
@@ -13,13 +13,127 @@
...
 from pytket.circuit import Node, OpType  # type: ignore
 
 
+def _serialize_all_node_gate_errors(
+    d: Optional[Dict[Node, Dict[OpType, float]]]
+) -> Optional[Dict[str, Dict[int, float]]]:
+    if d is None:
+        return None
+    return {
+        str(n.to_list()): {ot.value: err for ot, err in errs.items()}
```
ss2165, line 31: I would suggest either using `tuple(n.to_list())` or using a list of pairs instead of the dictionary (to avoid stringifying a list)

pytket/pytket/backends/backendinfo.py
```diff
@@ -13,13 +13,127 @@
...
+
+
+def _deserialize_all_node_gate_errors(
+    d: Optional[Dict[str, Dict[int, float]]]
+) -> Optional[Dict[Node, Dict[OpType, float]]]:
+    if d is None:
+        return None
+    return {
+        Node.from_list(literal_eval(n)): {
+            OpType(int(ot)): err for ot, err in errs.items()
```
ss2165, line 43: This doesn't match how Ciruit serializes optypes, where it uses the string of the optype (partially to try and keep the serialized form stable against the enum numbering changing).

pytket/tests/backendinfo_test.py
```diff
@@ -124,14 +124,21 @@ def test_serialization() -> None:
 
 def test_to_json() -> None:
     bi = BackendInfo(
-        "name",
```
ss2165, line 127: Could we also update the hypothesis test to generate samples for all the fields

ss2165 (changes requested):

pytket/pytket/backends/backendinfo.py
```diff
@@ -13,13 +13,127 @@
...
+        return None
+    return {str(n.to_list()): err for n, err in d.items()}
+
+
+def _deserialize_averaged_node_gate_errors(
+    d: Optional[Dict[str, float]]
+) -> Optional[Dict[Node, float]]:
+    if d is None:
+        return None
+    return {Node.from_list(literal_eval(n)): err for n, err in d.items()}
```
ss2165, line 102: Looks like there's some code duplication that can be reduced with more generic type signatures (e.g. by using `Sequence` or type variables), e.g. these two functions and the two before contain essentially the same body

pytket/pytket/backends/backendinfo.py
cqc-alec: That doesn't quite work because `tuple(n.to_list())` is something like `("node", [0])` which is not hashable, but we could do something similar by going one level down and converting the inner list to a tuple.

pytket/pytket/backends/backendinfo.py
cqc-alec: Right. This'll probably have to be exposed via the binders as I don't think it is currently.

pytket/pytket/backends/backendinfo.py
ss2165: I think the list of pairs is better then, its what we do with other 1-1 maps

PR #207, opened 2022-02-10
[WIP] [feature] Add stubs to tket

lmondada:
Adding type stubs to tket requires a few changes to the pytket API. This should probably be done before pytket 1.0. I have already solved a significant chunk of the issues in the commits below, but there are more issues to be discussed and resolved.

The stubs are automatically generated using mypy's `stubgen` in `setup.py`.

Most problems that arose in these generated stubs are addressed by changes to the `binders`. However some problems could not be solved that way (bugs in mypy, or maybe in some cases I haven't figured how), so I also need to patch the stubs after generation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 *.egg-info
 *.o
 *.so
+*.pyi
 .cache
 .eggs
 .hypothesis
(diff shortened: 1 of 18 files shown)


lmondada:
This is being broken up into several more manageable tasks.

PR #215, opened 2022-02-11

cqc-melf:


tket/src/Architecture/Architecture.cpp
```diff
@@ -24,37 +25,62 @@
...
+    if (uids.size() == 2) {
+      if (this->node_exists(uids[0]) && this->node_exists(uids[1]) &&
+          (this->edge_exists(uids[0], uids[1]) ||
+           this->edge_exists(uids[1], uids[0]))) {
+        return true;
+      }
+    }
+  } else if (is_single_qubit_type(op->get_type())) {
+    if (uids.size() == 1) {
+      if (this->node_exists(uids[0])) {
```
cqc-melf, line 49: @sjdilkes Should this be included in this PR or should I just return true in all single qubits cases independent of the placement of the qubit? If this is included a few c++ tests ( and python test, but I have not checked them in details yet)  need to be changed, see below.

tket/tests/test_MultiGateReorder.cpp
```diff
@@ -173,12 +173,28 @@ SCENARIO("Reorder circuits") {
...
+    }
+
     for (unsigned i = 0; i < 6; i++) {
       std::vector<Node> nodes;
       for (auto arg : commands[i].get_args()) {
         nodes.push_back(Node(arg));
       }
-      REQUIRE(shared_arc->valid_operation(commands[i].get_op_ptr(), nodes));
+
+      // will fail for
```
cqc-melf, line 188: @sjdilkes I am not sure I have understood this testcase here correctly, it is failing because some of the gates are violating the connectivity or they are three qubits gates. I have added a special handling for these cases.

cqc-melf:
see https://github.com/CQCL/tket/pull/243

PR #220, opened 2022-02-15
Don't assume expressions being substituted are mod 4

### LINKED ISSUE #219: Symbol substitution converts symbols that are a multiple of 4 to zero
This was a hasty workaround for a symengine issue: https://github.com/CQCL/tket/blob/87a4effbf4f81d444c37540638c7b7b9f7e2e13e/tket/src/Circuit/Circuit.cpp#L142 .
However it leads to incorrect results if the symbol is part of a larger expression.

cqc-alec:
Fixes #219 .

### DIFF
--- a/tket/src/Circuit/Circuit.cpp
+++ b/tket/src/Circuit/Circuit.cpp
@@ -134,11 +134,6 @@ Expr Circuit::get_phase() const {
 
 void Circuit::add_phase(Expr a) { phase += a; }
 
-static bool approx_0(const Expr &e) {
-  std::optional<double> v = eval_expr(e);
-  return v && (std::abs(v.value()) < EPS);
-}
-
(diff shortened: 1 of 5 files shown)


tket/src/Circuit/Circuit.cpp
```diff
@@ -142,7 +142,7 @@ void Circuit::symbol_substitution(const symbol_map_t &symbol_map) {
     // This is a workaround for a symengine issue: symengine currently has poor
     // handling of symbolic evaluations for atan2. However, this may not catch
     // every such issue, so we should revisit it.
```
ss2165, line 144: is this comment still true?

ss2165 (approved):

tket/src/Circuit/Circuit.cpp
cqc-alec: Yes I think it is still true.

PR #214, opened 2022-02-10
Rebase refactor with simpler RebaseCustom, exposed CircPool and auto_rebase_pass

ss2165:
* RebaseCustom didn't need to have separate single qubit and multi qubit gatesets -> merged
* expose CircPool so they can be reused in custom rebases
* naive auto rebase function that checks for known decompositions for given gateset. This can be expanded in future with more decomposition functions as we defined them, a CZ lambda is included as an example.

PR #226, opened 2022-02-16
Remove `Backend.characterisation`

cqc-alec:
After merging this I'll make a pytket pre-release and then update the extensions.

### DIFF
--- a/pytket/pytket/tailoring/__init__.py
+++ b/pytket/pytket/tailoring/__init__.py
@@ -13,9 +13,8 @@
 # limitations under the License.
 
 """
-The circuit module provides an API to interact with the tket
-:py:class:`Characterisation` suite, with methods for noise shaping, characterisation and
-mitigation. This module is provided in binary form during the PyPI installation.
+The `tailoring` module provides an API to interact with tket's charcterisation suite,
+with methods for noise shaping, characterisation and mitigation.
 """
 
 from pytket._tket.tailoring import *  # type: ignore
--- a/pytket/pytket/tailoring/__init__.py
+++ b/pytket/pytket/tailoring/__init__.py
@@ -13,8 +13,9 @@
 # limitations under the License.
 
 """
-The `tailoring` module provides an API to interact with tket's charcterisation suite,
-with methods for noise shaping, characterisation and mitigation.
+The circuit module provides an API to interact with the tket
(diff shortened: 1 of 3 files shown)


pytket/pytket/tailoring/__init__.py
```diff
@@ -13,9 +13,8 @@
 # limitations under the License.
 
 """
-The circuit module provides an API to interact with the tket
-:py:class:`Characterisation` suite, with methods for noise shaping, characterisation and
-mitigation. This module is provided in binary form during the PyPI installation.
+The `tailoring` module provides an API to interact with tket's charcterisation suite,
```
sjdilkes, line 16: Maybe this isn't the PR for it, but faithfully the `tailoring` module only provides an API for using frame randomisation methods, and maybe we should rename the whole module to frame randomisation in another PR?

If we save that for another review, it should be `tket's characterisation suite`.

sjdilkes (changes requested):
One query.

pytket/pytket/tailoring/__init__.py
cqc-alec: Done. (Just me being lazy, I saw it was referring to the wrong module name and thought I'd correct it at the same time. But yes it should be a new PR.)

ss2165:
> After merging this I'll make a pytket pre-release and then update the extensions.

The changes in #214 will also break things in extensions, should the fixes be sister PRs?

sjdilkes (approved):

cqc-alec:
> > After merging this I'll make a pytket pre-release and then update the extensions.
> 
> The changes in #214 will also break things in extensions, should the fixes be sister PRs?

Good point. Maybe we should do a pre-release first without this change, fix those, and then do another pre-release with this change?

cqc-alec:
> > > After merging this I'll make a pytket pre-release and then update the extensions.
> > 
> > 
> > The changes in #214 will also break things in extensions, should the fixes be sister PRs?
> 
> Good point. Maybe we should do a pre-release first without this change, fix those, and then do another pre-release with this change?

Actually, removing this doesn't break the extensions, so I'm going to merge and attempt another pre-release now.

(For some reason I'm having difficulty doing a pre-release. Seems the wrong version number is being applied by ertuptools_scm on some platforms...)

PR #228, opened 2022-02-16
Build symengine on manylinux.

cqc-alec:
Manylinux uses a different `compiler.libcxx` so there isn't a pre-built version available (currently).


cqc-melf (approved):
We could add that compiler version in the future?

cqc-alec:
> We could add that compiler version in the future?

Yes, that is the plan. This is a stop-gap.

PR #232, opened 2022-02-17
[feature] [rv3] Generate random nearby placements

lmondada:
Waiting to rebase on rv3 (https://github.com/CQCL/tket/pull/167) once that is merged into develop

### DIFF
--- a/tket/src/Placement/NeighbourPlacements.cpp
+++ b/tket/src/Placement/NeighbourPlacements.cpp
@@ -66,7 +66,7 @@ NeighbourPlacements::Result NeighbourPlacements::gen_result(
   return convert_to_res(swaps.to_vector());
 }
 
-NeighbourPlacements::SwapList NeighbourPlacements::gen_swap_list(
+SwapList NeighbourPlacements::gen_swap_list(
     unsigned dist) {
   SwapList swaps;
   auto edges = arc_.get_all_edges_vec();
--- a/tket/src/Placement/NeighbourPlacements.cpp
+++ b/tket/src/Placement/NeighbourPlacements.cpp
@@ -1,3 +1,17 @@
+// Copyright 2019-2022 Cambridge Quantum Computing
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 #include "NeighbourPlacements.hpp"
 
 #include <cstdlib>
--- a/tket/src/Placement/NeighbourPlacements.cpp
+++ b/tket/src/Placement/NeighbourPlacements.cpp
@@ -1,7 +1,6 @@
 #include "NeighbourPlacements.hpp"
 
 #include <cstdlib>
-// #include <string>
 
 #include "TokenSwapping/SwapListOptimiser.hpp"
 
@@ -13,7 +12,6 @@ NeighbourPlacements::NeighbourPlacements(
   auto nodes = arc_.get_all_nodes_vec();
   for (unsigned i = 0; i < nodes.size(); ++i) {
     u_to_node_.left.insert({i, nodes[i]});
-    // std::cout << nodes[i].repr() << std::endl;
   }
 }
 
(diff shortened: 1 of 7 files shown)


tket/src/Placement/NeighbourPlacements.cpp
```diff
@@ -0,0 +1,105 @@
...
+        ++n_unsuccessful;
+      }
+    } else {
+      for (auto swap : new_swaps.to_vector()) {
+        swaps.push_back(swap);
+      }
+    }
+  }
+
+  if (n_unsuccessful == N_TRIES) {
```
cqc-melf, line 60: What would be a good backup result for this case? Would it be possible to just use the initial mapping given as input?

tket/src/Placement/NeighbourPlacements.cpp
```diff
@@ -0,0 +1,105 @@
...
+  return convert_to_res(swaps.to_vector());
+}
+
+SwapList NeighbourPlacements::gen_swap_list(unsigned dist) {
+  SwapList swaps;
+  auto edges = arc_.get_all_edges_vec();
+  unsigned m = edges.size();
+  for (unsigned i = 0; i < dist; ++i) {
+    auto [n1, n2] = edges[rng_.get_size_t(m - 1)];
+    Swap new_swap{u_to_node_.right.at(n1), u_to_node_.right.at(n2)};
```
cqc-melf, line 75: It looks like to me that the second swap could be identical to the first one? have you thought about avoiding this case?

tket/src/Placement/include/Placement/NeighbourPlacements.hpp
```diff
@@ -0,0 +1,68 @@
...
+ * Optionally uses token swapping optimisations to try to ensure
+ * that the generated placements cannot be obtained in less than `m`
+ * swaps, but this cannot be guaranteed.
+ */
+class NeighbourPlacements {
+ public:
+  using SwapVec = std::vector<Swap>;
+  using NodeSwap = std::pair<Node, Node>;
+  using NodeSwapVec = std::vector<NodeSwap>;
+  struct Result {
```
cqc-melf, line 27: Would it be possible to add here the actual dist of the result from the initial mapping?

cqc-melf (commented):
Hi @lmondada,

I have not read all yet in details, but I already left a few comments.

if you have questions, please let me know.

tket/src/Placement/NeighbourPlacements.cpp
```diff
@@ -0,0 +1,105 @@
+#include "NeighbourPlacements.hpp"
```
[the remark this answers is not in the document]
lmondada: done

tket/src/Placement/NeighbourPlacements.cpp
```diff
@@ -0,0 +1,105 @@
...
+}
+
+NeighbourPlacements::Result NeighbourPlacements::gen_result(
+    unsigned dist, bool optimise) {
+  SwapList swaps;
+  tsa_internal::SwapListOptimiser optimiser;
+
+  // it might be impossible to find `dist` non-trivial swaps
+  unsigned n_unsuccessful = 0;
+  const unsigned N_TRIES = 10;
```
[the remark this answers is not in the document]
lmondada: True. It can now be passed to `NeighbourPlacements::get`.

tket/src/Placement/NeighbourPlacements.cpp
lmondada: You could return a placement with as many swaps (`< dist`) as possible. Would you prefer that?

tket/src/Placement/NeighbourPlacements.cpp
lmondada: Yes. There are lots of redundancies that are possible. I take care of them by using Zen's optimiser that simplifies a chain of swaps (and will definitely cancel two swaps that are identical). Whatever I would do here, it won't be better than what Zen has already implemented so I decided to leave all simplification with his code.

tket/src/Placement/include/Placement/NeighbourPlacements.hpp
```diff
@@ -0,0 +1,68 @@
+#pragma once
```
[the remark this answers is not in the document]
lmondada: done

tket/src/Placement/include/Placement/NeighbourPlacements.hpp
lmondada: The optimal dist is very hard to compute. I can add a `dist()` function that returns `swaps.size()`. Would that be useful?

lmondada:
I've addressed all of @cqc-melf's comments. I've also added a feature that makes sure the returned placement maps are all distinct and improved docs.

tket/src/Placement/include/Placement/NeighbourPlacements.hpp
cqc-melf: I think a swaps.size() function looks like the best solution for this.

tket/src/Placement/NeighbourPlacements.cpp
cqc-melf: Sound valid, lets keep it as it is.

tket/src/Placement/NeighbourPlacements.cpp
cqc-melf: Do you know how much additional runtime would be needed for an "as many swaps (< dist) as possible" solution?

lmondada: Sorry, I meant: I can return whatever maximum number of swaps I was able to generate, ie until I cannot greedily append swaps any further (not actually try to find a maximum).

tket/src/Placement/include/Placement/NeighbourPlacements.hpp
lmondada: Just to avoid confusion: `Result.swaps` is a vector, so it already has a `.size()`; are you fine with me not adding anythign?

cqc-melf: Yes that is fine

tket/src/Placement/NeighbourPlacements.cpp
cqc-melf: So there would be no additional runtime? For me that sounds like a good solution to just return that?

lmondada: OK. In that case, I will also return less than `n` solutions when I am unable to generate `n`. (see lines 60-73 in `NeighbourPlacements.cpp`)

cqc-melf (approved):

PR #233, opened 2022-02-17
Remove `.gitattributes`; normalize symengine conanfile line endings on Windows

cqc-alec:


.github/workflows/build_symengine.yml
```diff
@@ -146,6 +146,11 @@ jobs:
       run: pip install conan
     - name: create profile
       run: conan profile new tket --detect
+    - name: normalize line endings in conanfile
```
cqc-melf, line 149: Why are you doing this only for this one file?
The `.gitattributes` should have done that for the hole repo?

cqc-alec: Yes, but this is the only file it is necessary for. This file gets hashed by conan to determine the revision number.

cqc-melf (approved):

PR #242, opened 2022-02-19
Fix issues revealed by valgrind; add valgrind check on CI.

cqc-alec:


tket/src/Transformations/ControlledGates.cpp
```diff
@@ -432,22 +432,27 @@ static Circuit lemma71(
...
+    if (!bin.contains(v) && rep.get_OpType_from_Vertex(v) == OpType::CRy) {
       Expr v_angle = rep.get_Op_ptr_from_Vertex(v)->get_params()[0];
       Circuit cry_replacement = CircPool::CRy_using_CX(v_angle);
       Subcircuit sub{rep.get_in_edges(v), rep.get_all_out_edges(v), {v}};
-      rep.substitute(cry_replacement, sub, Circuit::VertexDeletion::Yes);
+      rep.substitute(cry_replacement, sub, Circuit::VertexDeletion::No);
+      bin.insert(v);
     }
-    if (rep.get_OpType_from_Vertex(v) == OpType::CU1) {
+    if (!bin.contains(v) && rep.get_OpType_from_Vertex(v) == OpType::CU1) {
```
yao-cqc, line 446: Would changing this to `else if` solve the pointer referencing problem?

cqc-alec: Yes the code could definitely be better structured...

cqc-alec: Done.

yao-cqc (approved):
Nice, thanks!

PR #245, opened 2022-02-21

cqc-melf:


cqc-melf:
Hi @sjdilkes,
I am not sure which of these changes should be merged to RV3.1. They might increase the runtime a bit, but it will take care that the optypes matches the number of qubits somehow. What do you think?

Took quiet some time to get all single qubit cases to work, I am not sure if not would not be better to check the single qubit placement as well to make sure that they are all placed at the end? (that would require changes in the routing process)

tket/src/Architecture/Architecture.cpp
```diff
@@ -25,33 +25,78 @@
...
+    }
   }
+
+  // Barriers are allways allowed
+  if (op->get_type() == OpType::Barrier) {
+    return true;
+  }
+
+  // allow two qubit gates only for placed and connected nodes
+  if (is_multi_qubit_type(op->get_type())) {
```
sjdilkes, line 78: It looks like you have some repeated code? I'm guessing a copy and paste gone wrong

sjdilkes (changes requested):
To address your comment here - I like the safe checking of the OpTypes and whether they are single_qubit/multi_qubit here. However i'm wondering whether we've cursed ourselves with this name and are trying to over fit the solution, in a scenario where the performance hampering could be too large (as you point out).

Sorry if this undoes work. The largest problem here is that due to its name and being an Architecture method, the correct approach to writing this method (which I think is what you have done) is too much for Mapping. Hence I now suggest we do the following two things:
1) Make Make a new MappingFrontier method, with similar name (valid_boundary_operation or something). Retain some of the same logic with these differences: a) return true if uids.size() == 1 always (remove single_qubit_type check), b) r emove multi_qubit_type check, but keep  how uids.size() ==2  and uids.size() == 3 checks works
2) Keep this Architecture::valid_operation as is and expose it to the pytket layer as an Architecture method. It will be useful for anybody making their own python level heuristics. Add python tests.

Does this sound reasonable to you?

tket/src/Architecture/Architecture.cpp
cqc-melf: > To address your comment here - I like the safe checking of the OpTypes and whether they are single_qubit/multi_qubit here. However i'm wondering whether we've cursed ourselves with this name and are trying to over fit the solution, in a scenario where the performance hampering could be too large (as you point out).
> 
> Sorry if this undoes work. The largest problem here is that due to its name and being an Architecture method, the correct approach to writing this method (which I think is what you have done) is too much for Mapping. Hence I now suggest we do the following two things:
> 
> 1. Make Make a new MappingFrontier method, with similar name (valid_boundary_operation or something). Retain some of the same logic with these differences: a) return true if uids.size() == 1 always (remove single_qubit_type check), b) r emove multi_qubit_type check, but keep  how uids.size() ==2  and uids.size() == 3 checks works
> 2. Keep this Architecture::valid_operation as is and expose it to the pytket layer as an Architecture method. It will be useful for anybody making their own python level heuristics. Add python tests.
> 
> Does this sound reasonable to you?

Yes that sound reasonable. As allready mention I was not really happy with the longer runtime, the function is used very often so it could have a big impact at the end increasing the runtime only a bit.

I think it would be much more consistent to remove the box and conditional checks from the architecture valid command in there, but that would require major changes.

I will continue as suggested.

cqc-melf:
> To address your comment here - I like the safe checking of the OpTypes and whether they are single_qubit/multi_qubit here. However i'm wondering whether we've cursed ourselves with this name and are trying to over fit the solution, in a scenario where the performance hampering could be too large (as you point out).
> 
> Sorry if this undoes work. The largest problem here is that due to its name and being an Architecture method, the correct approach to writing this method (which I think is what you have done) is too much for Mapping. Hence I now suggest we do the following two things:
> 
>     1. Make Make a new MappingFrontier method, with similar name (valid_boundary_operation or something). Retain some of the same logic with these differences: a) return true if uids.size() == 1 always (remove single_qubit_type check), b) r emove multi_qubit_type check, but keep  how uids.size() ==2  and uids.size() == 3 checks works
> 
>     2. Keep this Architecture::valid_operation as is and expose it to the pytket layer as an Architecture method. It will be useful for anybody making their own python level heuristics. Add python tests.
> 
> 
> Does this sound reasonable to you?

Yes that sound reasonable. As allready mention I was not really happy with the longer runtime, the function is used very often so it could have a big impact at the end increasing the runtime only a bit.

I think it would be much more consistent to remove the box and conditional checks from the architecture valid command in there, but that would require major changes.

I will continue as suggested.

tket/src/Architecture/Architecture.cpp
cqc-melf: One of the cases is handling the case of an conditional gate, so it is not fully identical

cqc-melf:
Replaced by https://github.com/CQCL/tket/pull/248

PR #246, opened 2022-02-21
Add register accessors to Circuit

yao-cqc:


ss2165 (changes requested):
Nice, can we also add methods to get all the registers?

ss2165 (approved):

PR #247, opened 2022-02-22
Add StandardPass for basic qubit relabelling

sjdilkes:
Via the NaivePlacement class and NaiveMappingPass.

Current Placement pass will relabel all qubits (whether they are already an architecture qubit or not) and frequently will leave some as "unplaced", prompting Routing to label it later.

Mapping confirms that all two-qubit constraints are satisfied -> if a Qubit is unlabelled and only has single qubit gates it will remain unlabelled.

Compilation should not leave these qubits unlabelled -> this PR adds a placement method for naively assigning any unlabelled qubit to any unassigned architecture qubit. Its use in `FullMappingPass` is reasonable as we can assume that any unlabelled qubits have only single qubit gates and so this assignment is fine.

### DIFF
--- a/pytket/binders/passes.cpp
+++ b/pytket/binders/passes.cpp
@@ -495,6 +495,14 @@ PYBIND11_MODULE(passes, m) {
       "\n:return: a pass to relabel :py:class:`Circuit` Qubits to "
       ":py:class:`Architecture` Nodes",
       py::arg("placer"));
+    
+  m.def(
+      "NaivePlacementPass", &gen_naive_placement_pass,
+      ":param architecture: The Architecture used for relabelling."
+      "\n:return: a pass to relabel :py:class:`Circuit` Qubits to "
+      ":py:class:`Architecture` Nodes",
+      py::arg("arc"));
+
 
   m.def(
       "RenameQubitsPass", &gen_rename_qubits_pass, "Rename some or all qubits.",
--- a/pytket/tests/predicates_test.py
+++ b/pytket/tests/predicates_test.py
@@ -32,6 +32,7 @@ from pytket.passes import (  # type: ignore
     RoutingPass,
     CXMappingPass,
     PlacementPass,
+    NaivePlacementPass,
     RenameQubitsPass,
     FullMappingPass,
     DefaultMappingPass,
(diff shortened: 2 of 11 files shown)


pytket/binders/passes.cpp
```diff
@@ -496,6 +496,13 @@ PYBIND11_MODULE(passes, m) {
       ":py:class:`Architecture` Nodes",
       py::arg("placer"));
 
+  m.def(
```
cqc-alec, line 499: Add note to changelog.

pytket/tests/predicates_test.py
```diff
@@ -660,31 +684,44 @@ def sq(a: float, b: float, c: float) -> Circuit:
     # FullMappingPass
     fm_pass = FullMappingPass(arc, placer, config=[LexiRouteRoutingMethod()])
     assert fm_pass.to_dict()["pass_class"] == "SequencePass"
+    # # for x in fm_pass.get_sequence():
```
cqc-alec, line 687: Remove commented-out lines.

tket/src/Placement/include/Placement/Placement.hpp
```diff
@@ -269,6 +269,16 @@ class Placement {
   Architecture arc_;
 };
 
+class NaivePlacement : public Placement {
```
cqc-alec, line 272: Add doxygen documentation for class.

tket/src/Placement/Placement.cpp
```diff
@@ -154,6 +154,45 @@ std::vector<qubit_mapping_t> Placement::get_all_placement_maps(
...
+  qubit_vector_t to_place;
+  std::vector<Node> placed;
+
+  // Find which/if any qubits need placing
+  for (const Qubit &q : circ_.all_qubits()) {
+    Node n(q);
+    if (!this->arc_.node_exists(n)) {
+      to_place.push_back(n);
+    }
+    if (this->arc_.node_exists(n)) {
```
cqc-alec, line 173:
```suggestion
    } else {
```

tket/src/Placement/Placement.cpp
```diff
@@ -154,6 +154,45 @@ std::vector<qubit_mapping_t> Placement::get_all_placement_maps(
...
+      placed.push_back(n);
+      // if already placed, make sure qubit retains placement
+      placement.insert({n, n});
+    }
+  }
```
cqc-alec, line 178: Maybe set a variable `unsigned n_placed = to_place.size();` here and use it in the places below where we call `to_place.size()`?

cqc-alec (changes requested):

pytket/binders/passes.cpp
sjdilkes: Done

tket/src/Placement/include/Placement/Placement.hpp
sjdilkes: done

tket/src/Placement/Placement.cpp
sjdilkes: ha yeah, done

tket/src/Placement/Placement.cpp
sjdilkes: done

cqc-alec (approved):

PR #248, opened 2022-02-23

cqc-melf:
As discussed in this PR:
1 Make Make a new MappingFrontier method, with similar name (valid_boundary_operation or something). Retain some of the same logic with these differences: a) return true if uids.size() == 1 always (remove single_qubit_type check), b) r emove multi_qubit_type check, but keep how uids.size() ==2 and uids.size() == 3 checks works
2 Keep this Architecture::valid_operation as is and expose it to the pytket layer as an Architecture method. It will be useful for anybody making their own python level heuristics. Add python tests.


sjdilkes (changes requested):
Just a single change

tket/src/Mapping/MappingFrontier.cpp
```diff
@@ -585,4 +586,53 @@ void MappingFrontier::merge_ancilla(
   this->bimaps_->final.left.erase(merge);
 }
 
+bool MappingFrontier::valid_boundary_operation(
+    const ArchitecturePtr& architecture, const Op_ptr& op,
+    const std::vector<Node>& uids) const {
+  // boxes are never allowed
+  if (is_box_type(op->get_type())) {
```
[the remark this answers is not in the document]
cqc-melf: done

tket/src/Architecture/Architecture.cpp
```diff
@@ -18,37 +18,26 @@
...
-    const Op_ptr& op, const std::vector<Node>& uids) const {
-  if (op->get_desc().is_box() ||
-      (op->get_type() == OpType::Conditional &&
-       static_cast<const Conditional&>(*op).get_op()->get_desc().is_box())) {
-    return false;
+bool Architecture::valid_operation(const std::vector<Node>& uids) const {
+  for (Node n : uids) {
+    if (!this->node_exists(Node(n))) return false;
   }
   if (uids.size() == 1) {
```
sjdilkes, line 31: Now this is just for Nodes and not OpTypes, and is not used in Routing, I guess this should check if the node is in the architecutre before returning true? its a different circumstance than to during mapping

tket/src/Architecture/Architecture.cpp
```diff
@@ -18,37 +18,26 @@
...
-  } else if (op->get_type() == OpType::Barrier) {
     return true;
   } else if (uids.size() == 2) {
-    if (this->node_exists(uids[0]) && this->node_exists(uids[1]) &&
-        this->bidirectional_edge_exists(uids[0], uids[1])) {
+    if (this->bidirectional_edge_exists(uids[0], uids[1])) {
       return true;
     }
-  } else if (uids.size() == 3 && op->get_type() == OpType::BRIDGE) {
+  } else if (uids.size() == 3) {
```
sjdilkes, line 37: similalry, architectures in general dont support 3 qubit gates, so we should probably just return false if there are more than 2 uids

tket/src/Architecture/CMakeLists.txt
```diff
@@ -27,11 +27,7 @@ add_library(tket-${COMP}
     NeighboursFromArchitecture.cpp)
 
 list(APPEND DEPS_${COMP}
-    Circuit
```
sjdilkes, line 30: sweeet

sjdilkes (changes requested):
A couple of small things.

tket/src/Architecture/Architecture.cpp
cqc-melf: Yes it should, that is already added in the for loop at the beginning of the function.

tket/src/Architecture/Architecture.cpp
cqc-melf: will remove that

sjdilkes (approved):

PR #249, opened 2022-02-23
Add method to get all opgroup names

yao-cqc:


tket/tests/Circuit/test_Circ.cpp
```diff
@@ -2618,6 +2624,7 @@ SCENARIO("Named operation groups") {
 
     Circuit c1 = c;
     REQUIRE(c == c1);
+    REQUIRE(c.get_opgroups() == opgroups2);
```
cqc-alec, line 2627: Should this test be about `c1.get_opgroups()`?

cqc-alec: Maybe do both then?

cqc-alec (approved):

PR #255, opened 2022-02-25
Merge develop into feature/RV3.1

sjdilkes:
Via PR as changes probably need review.


cqc-alec (changes requested):
The file `TokenSwappingWithArch/CMakeLists.txt` should have been abolished with the merge.

Apart from that, the merge looks OK to me. I will have a few more comments that I will save for the RV3.1-->develop PR.

cqc-alec:
Actually I may as well put these here, though they aren't actually about the merge, just as a reminder:
* Obviously the modifications to `build_and_test.yml` should be deleted, as should `kyriakos.py`.
* I don't think `Architecture` should depend on `Circuit` or `Op` -- these should be kept orthogonal. I think the dependency has been introduced just for one method (`Architecture::valid_operation()`), but I'd argue it doesn't belong as an `Architecture` method if it depends on `Circuit` and should be implemented higher up the stack.

cqc-melf:
> Actually I may as well put these here, though they aren't actually about the merge, just as a reminder:
> 
>     * Obviously the modifications to `build_and_test.yml` should be deleted, as should `kyriakos.py`.
> 
>     * I don't think `Architecture` should depend on `Circuit` or `Op` -- these should be kept orthogonal. I think the dependency has been introduced just for one method (`Architecture::valid_operation()`), but I'd argue it doesn't belong as an `Architecture` method if it depends on `Circuit` and should be implemented higher up the stack.
> 
>     * Header files should use `#pragma once`, not include guards.
> 
>     * Copyright notice missing on some new files.

About the dependency of OpType in Architecture: there are changes planed so no need to think about it now. I will keep that in mind.

sjdilkes:
> Actually I may as well put these here, though they aren't actually about the merge, just as a reminder:
> 
> * Obviously the modifications to `build_and_test.yml` should be deleted, as should `kyriakos.py`.
> * I don't think `Architecture` should depend on `Circuit` or `Op` -- these should be kept orthogonal. I think the dependency has been introduced just for one method (`Architecture::valid_operation()`), but I'd argue it doesn't belong as an `Architecture` method if it depends on `Circuit` and should be implemented higher up the stack.
> * Header files should use `#pragma once`, not include guards.
> * Copyright notice missing on some new files.

I've modifed build_and_test.yml, removed kyriakos.py, fixed the failing test case, removed TokenSwappingWithArch/CMakeLists.txt and added copyright/pragma once where I spotted it.

As Melf says, the OpType dependency is being removed in another PR so will be fixed once PR to develop is made.

sjdilkes:
> > Actually I may as well put these here, though they aren't actually about the merge, just as a reminder:
> > 
> > * Obviously the modifications to `build_and_test.yml` should be deleted, as should `kyriakos.py`.
> > * I don't think `Architecture` should depend on `Circuit` or `Op` -- these should be kept orthogonal. I think the dependency has been introduced just for one method (`Architecture::valid_operation()`), but I'd argue it doesn't belong as an `Architecture` method if it depends on `Circuit` and should be implemented higher up the stack.
> > * Header files should use `#pragma once`, not include guards.
> > * Copyright notice missing on some new files.
> 
> I've modifed build_and_test.yml, removed kyriakos.py, fixed the failing test case, removed TokenSwappingWithArch/CMakeLists.txt and added copyright/pragma once where I spotted it.
> 
> As Melf says, the OpType dependency is being removed in another PR so will be fixed once PR to develop is made.

I may have updated the workflow file one commit too early ha.

cqc-alec:
> I may have updated the workflow file one commit too early ha.

Oops, yes.

cqc-alec (approved):

PR #274, opened 2022-03-04
Remove `tket` command-line tool.

cqc-alec:
It's unused, undocumented, unmaintained and untested.


ss2165 (approved):
till we see you again

PR #278, opened 2022-03-07
add documentation of single qubit problem

cqc-melf:
Hi @yao-cqc ,

Could you please take a look on this changes to the documentation and let me know if you understand what the issue is on this point and if it is clear how to avoid it?

### DIFF
--- a/pytket/binders/mapping.cpp
+++ b/pytket/binders/mapping.cpp
@@ -105,7 +105,10 @@ PYBIND11_MODULE(mapping, m) {
       LexiLabellingMethod, std::shared_ptr<LexiLabellingMethod>, RoutingMethod>(
       m, "LexiLabellingMethod",
       "Defines a RoutingMethod for labelling Qubits that uses the "
-      "Lexicographical Comparison approach outlined in arXiv:1902.08091.")
+      "Lexicographical Comparison approach outlined in arXiv:1902.08091.\n"
+      "This method is not labelling qubits that only have single qubits gates "
+      "or not gates at all on them. To get them labelled as well this method "
+      "can be used after applying a placer first.")
       .def(py::init<>(), "LexiLabellingMethod constructor.");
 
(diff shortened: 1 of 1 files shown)


pytket/binders/mapping.cpp
```diff
@@ -105,7 +105,10 @@ PYBIND11_MODULE(mapping, m) {
       LexiLabellingMethod, std::shared_ptr<LexiLabellingMethod>, RoutingMethod>(
       m, "LexiLabellingMethod",
       "Defines a RoutingMethod for labelling Qubits that uses the "
-      "Lexicographical Comparison approach outlined in arXiv:1902.08091.")
+      "Lexicographical Comparison approach outlined in arXiv:1902.08091.\n"
+      "This method is not labelling qubits that only have single qubits gates "
+      "or not gates at all on them. To get them labelled as well this method "
+      "can be used after applying a placer first.")
```
yao-cqc, line 111: Can we simply say that this method only labels interacting qubits?

pytket/binders/mapping.cpp
```diff
@@ -105,7 +105,10 @@ PYBIND11_MODULE(mapping, m) {
       LexiLabellingMethod, std::shared_ptr<LexiLabellingMethod>, RoutingMethod>(
       m, "LexiLabellingMethod",
       "Defines a RoutingMethod for labelling Qubits that uses the "
-      "Lexicographical Comparison approach outlined in arXiv:1902.08091.")
+      "Lexicographical Comparison approach outlined in arXiv:1902.08091.\n"
+      "This method is only labelling interacting qubits that have multi qubit "
+      "gates on them. To get qubits with only single qubits gates or not gates "
```
cqc-alec, line 110:
```suggestion
      "gates on them. To get qubits with only single qubits gates or no gates "
```

yao-cqc (approved):

PR #279, opened 2022-03-07
[bugfix] fix develop issue with sympy

cqc-melf:
try to fix develop test run on python


cqc-melf:
This is only a a fix for the very near future, the possibility to work with sympy 1.10 should be added back in before the release.

see https://github.com/CQCL/tket/issues/280

yao-cqc (approved):
Thanks!

PR #281, opened 2022-03-08
Bugfix/add sympy 1.10 support

cqc-melf:
fixes problem with sympy 1.10

### DIFF
--- a/pytket/mypy.ini
+++ b/pytket/mypy.ini
@@ -26,3 +26,7 @@ ignore_errors = True
 [mypy-lark.*]
 ignore_missing_imports = True
 ignore_errors = True
+
+[mypy-sympy.*]
+ignore_missing_imports = True
+ignore_errors = True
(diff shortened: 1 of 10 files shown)


cqc-alec (approved):

cqc-melf:
Very simple solution to all the mypy errors, if you want something better please let me know.

pytket/mypy.ini
```diff
@@ -26,3 +26,7 @@ ignore_errors = True
 [mypy-lark.*]
 ignore_missing_imports = True
 ignore_errors = True
+
+[mypy-sympy]
+ignore_missing_imports = True
+ignore_errors = True
```
cqc-alec, line 32: Is this still necessary when we have the "type:  ignore" comments there?

cqc-melf: Thank you, it is not. Removed it.

cqc-alec (approved):
Thanks, I will make a new issue to try and fix up the sympy type annotations, but let's go with this for now.

PR #284, opened 2022-03-08
[feature] Enhance circuit statuses

johnchildren:
Extend CircuitStatus to support a few new fields, such as timestamps for
changes in status, queue position and detailed error information.


cqc-alec (commented):
Could you add a line about this to the changelog (`pytket/docs/changelog.rst`) please?

cqc-alec (approved):

PR #285, opened 2022-03-08
Bugfix/singlequbitlabelling

cqc-melf:
fix issue with unlabelled qubits only containing single qubit gates

### DIFF
--- a/pytket/binders/mapping.cpp
+++ b/pytket/binders/mapping.cpp
@@ -150,6 +150,6 @@ PYBIND11_MODULE(mapping, m) {
           "\n:param routing_methods: Ranked methods to use for routing "
           "subcircuits. In given order, each method is sequentially checked "
           "for viability, with the first viable method being used.",
-          py::arg("circuit"), py::arg("routing_methods"));
+          py::arg("circuit"), py::arg("routing_methods"), py::arg("label_single_qubits") = true);
 }
 }  // namespace tket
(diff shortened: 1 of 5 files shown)


pytket/binders/mapping.cpp
```diff
@@ -149,7 +149,10 @@ PYBIND11_MODULE(mapping, m) {
           "qubits.\n\n:param circuit: pytket circuit to be mapped"
           "\n:param routing_methods: Ranked methods to use for routing "
           "subcircuits. In given order, each method is sequentially checked "
-          "for viability, with the first viable method being used.",
-          py::arg("circuit"), py::arg("routing_methods"));
+          "for viability, with the first viable method being used."
+          "\n:param label_single_qubits: will not label qubits without gates "
```
cqc-alec, line 153: Maybe `label_isolated_qubits` would be a better name?

tket/src/Mapping/MappingManager.cpp
```diff
@@ -100,6 +101,56 @@ bool MappingManager::route_circuit_with_maps(
...
+
+    node_vector_t nodes_vec = this->architecture_->get_all_nodes_vec();
+
+    // place all unplaced qubits
+
+    for (Qubit q : q_vec) {
+      if (!this->architecture_->node_exists(Node(q))) {
+        // found unplaced qubit
+        // other checks could be added here to avoid placing unused qubits or
+        // qubits that are not in an ppb
```
cqc-alec, line 138: ppb?

pytket/binders/mapping.cpp
cqc-melf: done

tket/src/Mapping/MappingManager.cpp
cqc-melf: old comment, removed it

PR #287, opened 2022-03-14
Release/1.0.0

cqc-alec:


.github/disabled-workflows/build_and_test_v2.yml
```diff
@@ -0,0 +1,436 @@
+name: Build and test v2
```
sjdilkes, line 1: Do we want to preserve these (I assume now redundant) workflows, or can we remove them?

sjdilkes (changes requested):
Last 3 commits look fine. One question about outdated workflows.

.github/disabled-workflows/build_and_test_v2.yml
cqc-alec: We want to keep these, but they are disabled for now until we start active development on the `develop2` branch.

sjdilkes: I misunderstood their use! Got it now, thanks.

sjdilkes (approved):

PR #297, opened 2022-03-18
Update to pybind11_json 0.2.12.

cqc-alec:


cqc-alec:
Sorry, I don't understand why the build is failing on MacOS (I reran the build and it failed again so it seems a non-transient problem).
Will close for now.

yao-cqc (approved):

PR #298, opened 2022-03-18
[TKET-1913] Refactor long running tests

ferbetanzo:
This PR is a refactoring of the current tests and CI so those tests that have a running time >= 1 second (on a regular modern laptop) are executed only once a week. It includes the following changes:
- The long tests have the `"[.long]"` tag, which means that they are by default _hidden_, and can be run by specifying the `"[long]"` tag to the test_tket binary. **NOTE:** making a scenario hidden has the side-effect that its tags are not reported if you run the catch2 binary with `-t` to list the tags. 
- Remove the `TKET_TESTS_FULL` compilation flag so now all the tests are built at once.
- Refactor some of the tests so only the long running `GIVEN` cases are assigned the `"[.long]"` flag.
- Add the `tket-tests:long` option to run only the long tests.
- Change the coverage.yml to generate and check 2 different coverage reports, one for the _normal_ tests, and one for the _full suite_ tests.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -98,24 +98,41 @@ jobs:
     - name: Install gcovr
       run: pip install gcovr
     - name: Build coverage report
-      if: github.event_name == 'pull_request' || github.event_name == 'push'
       run: |
         mkdir test-coverage
         gcovr --print-summary --html --html-details -r ./tket/src/ --exclude-lines-by-pattern '.*\bTKET_ASSERT\(.*\);' --object-directory ${PWD}/build/tket/ -o test-coverage/index.html > test-coverage/summary.txt
-    - name: Build coverage report (full suite)
-      if: github.event_name == 'schedule'
-      run: |
-        mkdir test-coverage
-        gcovr --print-summary --html --html-details -r ./tket/src/ --exclude-lines-by-pattern '.*\bTKET_ASSERT\(.*\);' --object-directory ${PWD}/build/tket/ -o test-coverage/index.html > test-coverage/summary-full.txt
     - name: Upload artefact
       uses: actions/upload-artifact@v2
       with:
         name: test_coverage
         path: test-coverage/
 
+  check_coverage:
+    name: Check coverage
+    needs: generate_coverage
+    if: (github.event_name == 'push' && needs.changes.outputs.tket == 'true') || github.event_name == 'schedule'
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v2
+    - name: Download artefact
+      uses: actions/download-artifact@v2
+      with:
+        name: test_coverage
+        path: test-coverage/
+    - name: Compare with latest report from develop (short tests)
+      if: github.event_name == 'push'
+      run: |
+        wget https://cqcl.github.io/tket/tket/test-coverage-short/summary.txt
+        ./.github/workflows/compare-coverage summary.txt test-coverage/summary.txt
+    - name: Compare with latest report from develop (full suite)
+      if: github.event_name == 'schedule'
+      run: |
+        wget https://cqcl.github.io/tket/tket/test-coverage/summary.txt
+        ./.github/workflows/compare-coverage summary-full.txt test-coverage/summary-full.txt
+
   publish_coverage:
     name: Publish coverage
-    needs: generate_coverage
+    needs: check_coverage
     concurrency: gh_pages
     if: (github.event_name == 'push' && needs.changes.outputs.tket == 'true') || github.event_name == 'schedule'
     runs-on: ubuntu-20.04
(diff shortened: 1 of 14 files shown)


.github/workflows/coverage.yml
```diff
@@ -120,6 +133,7 @@ jobs:
         git config --global user.email "tket-bot@cambridgequantum.com"
         git config --global user.name  "«$GITHUB_WORKFLOW» github action"
     - name: Remove old report
+      if: github.event_name == 'push'
       run: git rm -r docs/tket/test-coverage
     - name: Add report to repository
```
cqc-alec, line 138: Need to be a bit careful here. Looks like we are pushing both the long and non-long coverage to the same place, so they will overwrite each other. I suggest that the existing directory should be used for the long version, and a different directory used for the non-long version, so that https://cqcl.github.io/tket/tket/test-coverage/ points to the long version.

cqc-alec (changes requested):

.github/workflows/coverage.yml
ferbetanzo: I've changed it so that the short tests directory is `test-coverage-short`, and I've also commited the required directory and summary.txt file in th `gh-pages` branch.

recipes/tket-tests/conanfile.py
```diff
@@ -43,7 +47,6 @@ def validate(self):
     def _configure_cmake(self):
         if self._cmake is None:
             self._cmake = CMake(self)
-            self._cmake.definitions["TESTS_FULL"] = self.options.full
```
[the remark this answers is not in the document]
ferbetanzo: Done :D

.github/workflows/coverage.yml
```diff
@@ -104,7 +117,7 @@ jobs:
     name: Publish coverage
     needs: generate_coverage
     concurrency: gh_pages
-    if: github.event_name == 'push' && needs.changes.outputs.tket == 'true'
+    if: (github.event_name == 'push' && needs.changes.outputs.tket == 'true') || github.event_name == 'schedule'
```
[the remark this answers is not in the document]
ferbetanzo: The publish job now depends on the check job.

.github/workflows/coverage.yml
```diff
@@ -100,48 +107,63 @@ jobs:
...
         path: test-coverage/
 
-  publish_coverage:
-    name: Publish coverage
+  check_coverage:
+    name: Check coverage
     needs: generate_coverage
-    concurrency: gh_pages
-    if: github.event_name == 'push' && needs.changes.outputs.tket == 'true'
+    if: (github.event_name == 'push' && needs.changes.outputs.tket == 'true') || github.event_name == 'schedule'
```
cqc-alec, line 113: We want to run this job on `pull_request`, not `push`. (Coverage reduction should be caught at PR stage.)

On the other hand, the "publish" job should run on `push`. So the dependencies need tweaking a bit. (I think `publish_coverage` needs `generate_coverage` but not `check_coverage`?)

cqc-alec (changes requested):

cqc-alec:
Thanks, one question: on "schedule" runs, do we want to publish if the check fails?

- Reason to do so: the published report will correspond to the state of develop.
- Reason not to do so: subsequent checks will pass, meaning we could easily ignore the failure and coverage will go down.

Also, "Add report to repository (short tests)" and "Add report to repository (full suite)" should be qualified with "if" clauses I think.

cqc-alec:
> @cqc-alec that's a good question. I think there's definitely a risk of ignoring a decrease in the coverage of the full suite if we always publish the report regardless of the check. So, I'd prefer to only update it if the check succeeds. On the other hand, I'm not sure if there exist any other stuff that depends on the report being updated every week?

No, there is nothing else that depends on it. It is only useful as a reference for developers -- so that we can see where coverage is missing.
One option would be to publish it to a different locations -- say, `test-coverage-develop` -- which we can still use as a reference but which won't be used by this check in future runs.

ferbetanzo:
> Also, "Add report to repository (short tests)" and "Add report to repository (full suite)" should be qualified with "if" clauses I think.
You're totally right. I've fixed it now.

ferbetanzo:
> No, there is nothing else that depends on it. It is only useful as a reference for developers -- so that we can see where coverage is missing. One option would be to publish it to a different location -- say, `test-coverage-develop` -- which we can still use as a reference but which won't be used by this check in future runs.

Mmm, since we already have 2 different coverage reports (the short and the full one), I'm worried that introducing another one can be more harmful than useful. I'm not entirely sure about the use case and when would you exactly want to see the report of the coverage, but everytime the workflow is run the generated coverage report is saved as an artifact that you can download from the GH Actions UI (even if it was not published), and then you can compare the current with the obtained one... would this be sufficient enough?

If so, then I think I'll put the requirement that for the `'schedule'` publishing must be done only after a successful check.

.github/workflows/coverage.yml
ferbetanzo: Addressed :)

ferbetanzo:
@cqc-alec I've reflected our discussion on when to publish the report in my last PR. I ended up creating 2 different publishing jobs since they have a somewhat different logic. I think this was the only unresolved review feedback left.

cqc-alec (commented):
Thanks! Sorry, one more question: on scheduled runs, what will be the value of `needs.changes.outputs.tket`? We want to make sure we run `generate_coverage` at least if there have been any changes on `develop` since the last scheduled run; does the current logic ensure that?

ferbetanzo:
> Thanks! Sorry, one more question: on scheduled runs, what will be the value of needs.changes.outputs.tket? We want to make sure we run generate_coverage at least if there have been any changes on develop since the last scheduled run; does the current logic ensure that?

That's a very good point. Reading at the `dorny/paths-filter@v2` documentation, I'm not sure if the `steps.filter.outputs.tket` would be set to true if the workflow is not initiated by a push. I have added it explicitly to the `if` in the `generate_coverage`. That means that, even if there are no changes pushed to develop, we would always run the workflow every Saturday morning. Would that work?

cqc-alec:
> That's a very good point. Reading the `dorny/paths-filter@v2` documentation, I'm not sure if the `steps.filter.outputs.tket` would be set to true if the workflow is not initiated by a push. I have added it explicitly to the `if` in the `generate_coverage`. That means that, even if there are no changes pushed to develop, we would always run the workflow every Saturday morning. Would that work?

Sounds good, yes, thanks.

cqc-alec (approved):

PR #302, opened 2022-03-23
Feature/serialise measurement setup

yao-cqc:


tket/src/MeasurementSetup/include/MeasurementSetup/MeasurementSetup.hpp
```diff
@@ -37,6 +38,7 @@ namespace tket {
 class MeasurementSetup {
  public:
   struct MeasurementBitMap {
+    MeasurementBitMap(){};
```
cqc-alec, line 41:
```suggestion
    MeasurementBitMap(){}
```

cqc-alec, line 41: Can we add a test for serialization/deserialization of a default-constructed `MeasurementBitMap`?

tket/src/MeasurementSetup/MeasurementSetup.cpp
```diff
@@ -107,4 +107,50 @@ std::string MeasurementSetup::to_str() const {
...
+      map_list;
+  for (const std::pair<
+           const QubitPauliString,
+           std::vector<MeasurementSetup::MeasurementBitMap>> &tensor_map :
+       setup.get_result_map()) {
+    map_list.push_back(tensor_map);
+  }
+  // sort the list for consistent serialisation
+  std::sort(map_list.begin(), map_list.end(), [](auto pair1, auto pair2) {
+    return pair1.first < pair2.first;
```
cqc-alec, line 136: Is it possible for two elements of the list to have the same first member? (In which case we'd also need to compare the second elements.)

cqc-alec (changes requested):

tket/src/MeasurementSetup/MeasurementSetup.cpp
yao-cqc: It's shouldn't be possible because the first elements are the keys in the result map.

tket/src/MeasurementSetup/include/MeasurementSetup/MeasurementSetup.hpp
```diff
@@ -37,6 +38,7 @@ namespace tket {
 class MeasurementSetup {
  public:
   struct MeasurementBitMap {
+    MeasurementBitMap() {}
```
cqc-alec, line 41: Shoul explicitly initialize `_circ_index` and `_invert` here.

cqc-alec (approved):

PR #303, opened 2022-03-23
Feature/serialization for predicate

yao-cqc:


cqc-alec:
Thanks, one question: on "schedule" runs, do we want to publish if the check fails?
Reason to do so: the published report will correspond to the state of develop.
Reason not to do so: subsequent checks will pass, meaning we could easily ignore the failure and coverage will go down.
At the moment I think it is indeterministic which happens, because `check_coverage` and `publish_coverage` can be run concurrently.

pytket/binders/predicates.cpp
```diff
@@ -82,7 +83,20 @@ PYBIND11_MODULE(predicates, m) {
...
       .def("__str__", [](const Predicate &) { return "<tket::Predicate>"; })
-      .def("__repr__", &Predicate::to_string);
+      .def("__repr__", &Predicate::to_string)
+      .def(
+          "to_dict",
+          [](const PredicatePtr &predicate) {
+            return nlohmann::json(predicate);
+          },
+          "Return a JSON serializable dict representation of "
+          "the Predicate.\n"
```
cqc-alec, line 93:
```suggestion
          "the Predicate.\n\n"
```

cqc-alec (changes requested):

cqc-alec (approved):

PR #307, opened 2022-03-25
Pin jinja2 and gcovr versions.

cqc-alec:
Works around incompatibility between gcovr 5.0 and jinja2 3.1.0 (see https://github.com/CQCL/tket/runs/5688235924 for error).


ferbetanzo (approved):
Thanks for the change. LGTM.

PR #308, opened 2022-03-25
Work around issues arising from symbolic squashing.

### LINKED ISSUE #304: Symbolic substitution after compilation yields operations with parameters containing `nan`
```python
from sympy import Symbol
from pytket.passes import SynthesiseTket

a, b = Symbol("a"), Symbol("b")
c = Circuit(1)
c.Rx(a, 0).Ry(b, 0)
SynthesiseTket().apply(c)
c.symbol_substitution({a: 0, b: 0})
print(c.get_commands())
```
Output:
```
[TK1(1/2 - (1/4 + nan*pi**(-1)), 0, -1/2 - (-1/4 + nan*pi**(-1))) q[0];]
```

cqc-alec:
Fixes #304.

### DIFF
--- a/tket/src/Gate/Gate.cpp
+++ b/tket/src/Gate/Gate.cpp
@@ -21,8 +21,12 @@
 #include "OpPtrFunctions.hpp"
 #include "OpType/OpType.hpp"
 #include "OpType/OpTypeInfo.hpp"
+#include "Utils/Expression.hpp"
+#include "Utils/RNG.hpp"
+#include "symengine/eval_double.h"
 
 namespace tket {
+using std::stringstream;
 
 Op_ptr Gate::dagger() const {
   OpType optype = get_type();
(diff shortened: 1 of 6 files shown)


tket/src/Gate/Gate.cpp
```diff
@@ -221,13 +225,79 @@ Op_ptr Gate::transpose() const {
...
+    ss << e;
+    if (std::regex_search(ss.str(), nan_regex)) return true;
+  }
+  return false;
+}
+
+static double random_perturbation() {
+  static RNG rng;
+  int a = rng.get_size_t(20);
+  return (a - 10) * EPS;
```
willsimmons1465, line 241: We might want to consider dropping the range down a bit. In the ideal world, our compilation has deterministic results, but even in this case we should aim for angles to not differ by more than `ERR_EPS` between runs. Whilst `ERR_EPS` is 10*`EPS`, these permutations could be added to multiple symbols in the expression and I don't know whether the trig will blow this up to a larger error. Probably a bit much to do the calculations by hand to see what a viable range would be, but if we can get away with halving the range that might be good

tket/src/Gate/Gate.cpp
```diff
@@ -221,13 +225,79 @@ Op_ptr Gate::transpose() const {
...
+  // we fail after 1000 attempts, give up.
+  for (unsigned i = 0; i < 1000; i++) {
+    SymEngine::map_basic_basic new_sub_map;
+    for (const auto& pair : sub_map) {
+      new_sub_map[pair.first] = Expr(pair.second) + random_perturbation();
+    }
+    std::vector<Expr> new_params_1 =
+        subs_all_params(this->params_, new_sub_map, true);
+    if (!params_contain_nan(new_params_1)) {
+      return get_op_ptr(this->type_, new_params_1, this->n_qubits_);
```
willsimmons1465, line 294: I presume that forcing the reduction to a double here is how we enforce any `atan2` to decide whether to reduce to a `nan` or not. Since this will only happen when the full expression contains no symbols, I don't think this will always catch it when multiple symbols are instantiated one at a time. For example, if the expression is of the form `alpha + ... atan2(... beta ..., ...) ...` and `beta` is instantiated first, it could be set to a value that would yield a `nan` from the `atan2`. However, we would only detect this at the later point when `alpha` is instantiated, at which point the permutations would only affect `alpha` and so wouldn't resolve the `nan`. Perhaps we just let this case hit the exception? The other alternative I can think of is replacing the `eval_double` call with a more expensive call to the generic simplification procedure so we spot the `nan` at the point where a relevant symbol is instantiated. Which do you think would work better?

tket/src/Gate/Gate.cpp
```diff
@@ -221,13 +225,79 @@ Op_ptr Gate::transpose() const {
...
+    }
+  }
+
+  // Something really is fishy.
+  throw SymEngine::NotImplementedError(
+      "Invalid substitution in symbolic expression");
```
willsimmons1465, line 300: I think this error message could be more specific to give the user an indication of something to try and resolve it themselves. Also, I don't think we should use SymEngine's error classes for exceptions thrown in our own code. That could cause confusion when debugging this in future

willsimmons1465 (changes requested):
I'm mostly happy with the changes. A couple of points to discuss further

tket/src/Gate/Gate.cpp
cqc-alec: Done.

tket/src/Gate/Gate.cpp
cqc-alec: I think this can be an exceptional case. It seems to be quite hard to trigger! (I tried.)
Another thing that can happen is infinities, which are different, and represented in symengine expressions by `zoo` etc. For example these arise when dividing by zero. We could expand the regex to catch these as well, but I think they will be even more rare in practice.

willsimmons1465: Wouldn't the following example trigger it?
```
circ = Circuit(1)
circ.Rz(gamma, 0)
circ.Rx(alpha, 0)
circ.Ry(beta, 0)
circ.Rz(gamma, 0)
SquashTK1().apply(circ)
circ.symbol_substitution({alpha : 0., beta : 0.})
circ.symbol_substitution({gamma : 0.})
```
In between the two substitutions, the command is `TK1(1/2 - (3/4 - (1/2 + gamma) + nan*pi**(-1)), 0, -1/2 - (1/4 - (1/2 + gamma) + nan*pi**(-1))) q[0];` where both expressions containing `nan` also contain `gamma`.

But yeah, I'm happy for this to fall through to the exception. We shouldn't be generating the infinite cases within our compilation so it makes sense for that kind of thing to also hit the exception as well, since it should be traceable back to a user input error

cqc-alec: Thanks for the example, yes it should do!

PR #316, opened 2022-04-05
Add `DefaultRoutingPass`

sjdilkes:
Currently there is no option to make a pass for only routing circuits with choice over which `RoutingMethod` are used. This PR adds a `DefaultRoutingPass` which returns a pass which does routing with [`LexiLabellingMethod()`, `LexiRouteRoutingMethod()`] and changes `RoutingPass` to take `List[RoutingMethod]` as an argument.


yao-cqc (approved):
Looks good!

PR #320, opened 2022-04-07
Use prebuilt symengine for release builds

cqc-alec:
Note, packages have been uploaded already, by [this](https://github.com/CQCL/tket/actions/runs/2109259738) test run.
Release workflow using prebuilt package: https://github.com/CQCL/tket/actions/runs/2109353901

### DIFF
--- a/.github/workflows/build_symengine.yml
+++ b/.github/workflows/build_symengine.yml
@@ -167,3 +167,29 @@ jobs:
     - name: upload package
       if: github.event_name == 'push'
       run: conan upload symengine/${{ env.SYMENGINE_VER }}@tket/stable --all -r=tket-conan
+  manylinux:
+    name: build symengine (manylinux)
+    needs: changes
+    if: needs.changes.outputs.recipes_symengine == 'true'
+    runs-on: ubuntu-20.04
+    env:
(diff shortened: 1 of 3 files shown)


.github/workflows/build_symengine.yml
```diff
@@ -167,3 +167,29 @@ jobs:
     - name: upload package
       if: github.event_name == 'push'
       run: conan upload symengine/${{ env.SYMENGINE_VER }}@tket/stable --all -r=tket-conan
+  manylinux:
+    name: build symengine (manylinux)
+    needs: changes
+    if: needs.changes.outputs.recipes_symengine == 'true'
+    runs-on: ubuntu-20.04
+    env:
+      UPLOAD_PACKAGE: "NO"
```
cqc-melf, line 176: I am not fully convinced that this is "working as expected", but if this is not working it will not effect this workflow, an empty variable is still not yes.

(I think this is only working within this file, not in the bash script. )

cqc-alec: It is passed to the docker image via the file `env-vars` below.

cqc-melf (approved):

PR #322, opened 2022-04-07
Upload spdlog package to tket/stable

cqc-alec:
Follows the pattern of the symengine workflow.

This version of `spdlog` differs from the version on conan-center in that it does not have `fmt` as an external dependency.

In a [follow-up PR](https://github.com/CQCL/tket/pull/323) I will make tket depend on this version.

### DIFF
--- a/.github/workflows/build_spdlog.yml
+++ b/.github/workflows/build_spdlog.yml
@@ -1,0 +1,193 @@
+name: build spdlog
+on:
+  push:
+    branches:
+      - develop
+  pull_request:
+    branches:
+      - develop
+jobs:
+  changes:
+    runs-on: ubuntu-20.04
+    outputs:
+      recipes_spdlog: ${{ steps.filter.outputs.recipes_spdlog }}
+    steps:
+    - uses: actions/checkout@v2
+    - uses: dorny/paths-filter@v2
+      id: filter
+      with:
+        base: ${{ github.ref }}
+        filters: |
+          recipes_spdlog:
+            - 'recipes/spdlog/**'
+  linux:
+    name: build spdlog (linux)
+    needs: changes
+    if: needs.changes.outputs.recipes_spdlog == 'true'
+    runs-on: ubuntu-20.04
+    strategy:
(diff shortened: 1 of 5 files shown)


.github/workflows/build_spdlog.yml
```diff
@@ -0,0 +1,193 @@
...
+      run: |
+        $spdlog_ver = conan inspect --raw version recipes/spdlog/
+        echo "SPDLOG_VER=${spdlog_ver}" >> $env:GITHUB_ENV
+    - name: upload package (dry run)
+      if: github.event_name == 'pull_request'
+      run: conan upload spdlog/${{ env.SPDLOG_VER }}@tket/stable --all -r=tket-conan --skip-upload
+    - name: upload package
+      if: github.event_name == 'push'
+      run: conan upload spdlog/${{ env.SPDLOG_VER }}@tket/stable --all -r=tket-conan
+  manylinux:
```
cqc-melf, line 168: Why is the linux and the manylinux build needed?

recipes/spdlog/conandata.yml
```diff
@@ -0,0 +1,40 @@
+sources:
+  "1.10.0":
+    url: "https://github.com/gabime/spdlog/archive/v1.10.0.tar.gz"
+    sha256: "697f91700237dbae2326b90469be32b876b2b44888302afbc7aceb68bcfe8224"
```
cqc-melf, line 4: Should I check the hash values?

.github/workflows/build_spdlog.yml
cqc-alec: The manylinux build is for use by the release builds. It differs in that it uses a different libcpp.

recipes/spdlog/conandata.yml
cqc-alec: This is copied from https://github.com/conan-io/conan-center-index/blob/master/recipes/spdlog/all/conandata.yml

cqc-melf (approved):

PR #325, opened 2022-04-08
Revert to using spdlog from conan-center.

cqc-alec:


cqc-melf (approved):
maybe change the cache as well?

cqc-melf (approved):

PR #326, opened 2022-04-09
Remove dependency on spdlog.

cqc-alec:
This has the effect of removing the `pthread_cond_clockwait` dependency, and so fixing the issue with incompatibility with older versions of `glibc`.

The only loss from this is in being able to format strings passed to logging functions using format strings. There were only a few places where we were doing this. Instead, we have to construct them as strings and pass those. When all compilers have support for C++20 `std::format` we can probably use that.

(Although the new code is fully covered by tests, the branch coverage has gone down. Apparently this is because gcovr's branch coverage includes compiler-generated branches, so it isn't easy to control. There is a `--decisions` flag in gcovr 5.1 which is supposed to help with this, but it causes errors to be thrown. So I opted to remove branch coverage from the comparison and just base it on line coverage. Will raise a new issue to look into this.)


tket/tests/Utils/test_Logging.cpp
```diff
@@ -0,0 +1,42 @@
...
+  tket_log()->set_level(LogLevel::Off);
+
+  tket_log()->trace("Test trace (no output).");
+  tket_log()->debug("Test debug (no output).");
+  tket_log()->info("Test info (no output).");
+  tket_log()->warn("Test warn (no output).");
+  tket_log()->error("Test error (no output).");
+  tket_log()->critical("Test critical (no output).");
+
+  tket_log()->set_level(LogLevel::Err);
```
ss2165, line 39: is there anyway to capture the stdout stream and check correctness?

cqc-alec: Not that I could find, unfortunately: https://github.com/catchorg/Catch2/tree/devel/docs

ss2165: https://stackoverflow.com/questions/4810516/c-redirecting-stdout can we use this?

cqc-alec: It seems a lot easier to do this in pytest: I could add some tests there, along the lines of [this one](https://github.com/CQCL/tket/blob/a18e64f664a58c8d5b6d3bb8fe84b7a4ce9a9530/pytket/tests/measurement_setup_test.py#L68)?

cqc-alec: Alternatively: give the logging functions an optional argument representing an ostream, defaulting to `std::cout`; then the tests can use this. I might try that.

ss2165: yes I'm happy with either option, though the second sounds more thorough

ss2165 (approved):

PR #332, opened 2022-04-12
Bitwise operations

Roland-djee:
This PR addresses the addition of classical bitwise ops: AND, OR and XOR to the QIR generation process.

### DIFF
--- a/pytket/pytket/qir/qir.py
+++ b/pytket/pytket/qir/qir.py
@@ -66,18 +66,66 @@ QUANTINUUM_GATES = GateSet(
     name="Quantinuum",
     template=Template("__quantinuum__${optype}__${opname}__body"),
     gateset={
-        "h": CustomPyQIRGate(functions=[types.QUBIT]),
-        "x": CustomPyQIRGate(functions=[types.QUBIT]),
-        "y": CustomPyQIRGate(functions=[types.QUBIT]),
-        "z": CustomPyQIRGate(functions=[types.QUBIT]),
-        "rx": CustomPyQIRGate(functions=[types.DOUBLE, types.QUBIT]),
-        "ry": CustomPyQIRGate(functions=[types.DOUBLE, types.QUBIT]),
-        "rz": CustomPyQIRGate(functions=[types.DOUBLE, types.QUBIT]),
-        "phx": CustomPyQIRGate(functions=[types.DOUBLE, types.DOUBLE, types.QUBIT]),
-        "cnot": CustomPyQIRGate(functions=[types.QUBIT, types.QUBIT]),
-        "zzmax": CustomPyQIRGate(functions=[types.QUBIT, types.QUBIT]),
-        "zzph": CustomPyQIRGate(functions=[types.DOUBLE, types.QUBIT, types.QUBIT]),
-        "mz": CustomPyQIRGate(functions=[types.QUBIT, types.RESULT]),
+        "h": CustomPyQIRGate(
+            optype="qis",
+            functions=[types.QUBIT]
+        ),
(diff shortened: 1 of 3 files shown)


pytket/pytket/qir/qir.py
```diff
@@ -19,11 +19,16 @@
...
 from pytket import Circuit, OpType, Bit, Qubit
 from pytket.circuit import Op  # type: ignore
+from pytket.circuit.logic_exp import BitWiseOp
 
 
 CustomPyQIRGate = NamedTuple(
     "CustomPyQIRGate",
     [
+        (
+            "optype",
```
ss2165, line 29: I'd suggest calling it something other than "optype" since that term is very overloaded in tket

pytket/tests/qir_test.py
```diff
@@ -151,34 +159,16 @@ def test_qir_from_pytket_circuit_and_pyqir_gateset(
...
-#     assert call in em_ir_str
+def test_bitwise_ops_for_quantinuum_gateset(
+    circuit_bitwise_ops: Circuit, bitwise_file: str
+) -> None:
+    with open(bitwise_file, "r") as input:
+        data = input.read()
+    print(data)
+    call_and = f"call void @__quantinuum__cis__and__body(%Result* %zero, %Result* %zero1, %Result* %zero2)"
+    call_or = f"call void @__quantinuum__cis__or__body(%Result* %zero3, %Result* %zero4, %Result* %zero5)"
+    call_xor = f"call void @__quantinuum__cis__xor__body(%Result* %zero6, %Result* %zero7, %Result* %zero8)"
```
ss2165, line 170: yeah I think we're going to have to make sure we have SSA calls with set outputs

pytket/pytket/qir/qir.py
Roland-djee: Definitely. I had the same reflection. Should be a bit more creative.

pytket/tests/qir_test.py
Roland-djee: Sure. This is something I am not entirely clear with atm. Need to think through it a little.

pytket/pytket/qir/qir.py
Roland-djee: In [07951c5](https://github.com/CQCL/tket/pull/332/commits/07951c562d645be16af5be0739bac9434fa6ee36)

pytket/tests/qir_test.py
Roland-djee: I guess, one thing that I can do here is to reflect better the DAG structure. If I understand correctly, the output of an `AND` operation should be an SSA variable of the same type. Then, there must be input/output bindings to be made at the start and end of each SSA block (using SORE/READ). Concretely, that would mean there should be some return type for the function calls.

ss2165: yes, though I think you only need one STORE/READ per Bit of the Circuit at the start and end of the dag, and just use SSA variables in the middle

Roland-djee: OK, it looks like the `add_external_function` only works for functions defined to return a `types.VOID`. I've tracked the Python binding and it finally calls an `emit_void_call` [here](https://github.com/qir-alliance/pyqir/blob/dd4f5fe56c8f8dc27483f96a40d63c572c37ce72/qirlib/src/generation/qir/instructions.rs#L183). There is also an `emit_call_with_return` but it only works for calls to `Measure`. Not sure there is a direct call to that method from Python. Will investigate.

Roland-djee:
Closing now for non-disclosure purposes.

PR #337, opened 2022-04-19
Add OpType::TK2

cqc-alec:


tket/src/Gate/Gate.cpp
```diff
@@ -356,6 +359,17 @@ std::optional<double> Gate::is_identity() const {
       } else
         return notid;
     }
+    case OpType::TK2: {
+      int phase = 0;
+      for (const Expr& a : params) {
+        if (equiv_0(a + 2, 4)) {
+          phase ^= 1;
```
lmondada, line 366: You could use `bool pi_phase` instead of `int phase`.

lmondada (changes requested):
One small suggestion.

tket/src/Gate/Gate.cpp
cqc-alec: Done.

lmondada (approved):

PR #338, opened 2022-04-19
[TKET-1823] add wasm op

cqc-melf:


cqc-melf:
Adding the wasm op to tket.

PR #343, opened 2022-04-20
Always uninstall pytket before installing.

cqc-alec:
This avoids problems with conflicts when we update dependency versions.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -300,6 +300,7 @@ jobs:
         pyenv shell tket-3.8
         OPENBLAS="$(brew --prefix openblas)" pip install -U scipy
         cd pytket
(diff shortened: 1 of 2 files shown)


cqc-melf (approved):
WE should think about keeping .github/disabled-workflows up to date or remove them until needed ...?

cqc-alec:
> WE should think about keeping .github/disabled-workflows up to date or remove them until needed ...?

They are already rather out of date; I think there is little point maintaining them until needed. Yes, perhaps the best thing is to remove them for now.

cqc-melf:
> > WE should think about keeping .github/disabled-workflows up to date or remove them until needed ...?
> 
> They are already rather out of date; I think there is little point maintaining them until needed. Yes, perhaps the best thing is to remove them for now.

I think removing them is the right thing to do.

PR #344, opened 2022-04-20
Remove disabled workflows.

cqc-alec:
They are already out of date, but we can always recover them from git history to use as a basis for TKET 2 workflows when needed.


cqc-melf (approved):

cqc-melf:
Can you remove the directory as well?

cqc-alec:
> Can you remove the directory as well?

I think that is neither necessary nor possible: directories are not really a thing in git, AIUI. But when you pull from the repo there will not be an empty directory there.

PR #349, opened 2022-05-03
Construct a new `OptVertex` when populating `vertices`.

cqc-alec:
I think this should fix the valgrind issue.


lmondada:
Thank you so much Alec 🙏

lmondada (approved):

PR #352, opened 2022-05-06
Create symengine 0.9.0.1 requiring boost 1.79.0.

cqc-alec:
Will follow-up with a TKET PR to use this version of symengine and boost, which will allow us to drop one of our build-time patches (already tested locally).

### DIFF
--- a/recipes/symengine/conanfile.py
+++ b/recipes/symengine/conanfile.py
@@ -20,7 +20,8 @@ required_conan_version = ">=1.33.0"
 
 class SymengineConan(ConanFile):
     name = "symengine"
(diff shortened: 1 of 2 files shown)


recipes/symengine/conanfile.py
```diff
@@ -20,7 +20,8 @@
 
 class SymengineConan(ConanFile):
     name = "symengine"
-    version = "0.9.0"
+    version = "0.9.0.1"
+    upstream_version = "0.9.0"
```
cqc-melf, line 24: What is the upstream version doing?

cqc-melf (approved):

PR #353, opened 2022-05-06
Update to boost 1.79.0

cqc-alec:
This allows us to remove one of our build-time patches.
Also update to symengine 0.9.0.1 which uses the same version of boost.
Also update to latest catch2 version while we're about it.


recipes/tket/patches/adjacency_list.diff
```diff
@@ -1,24 +0,0 @@
---- a/boost/graph/detail/adjacency_list.hpp
-+++ b/boost/graph/detail/adjacency_list.hpp
```
cqc-melf, line 2: I am a little confused: from my understanding of this PR this file should be deleted, but some lines are green?

cqc-alec: The file is deleted. I think that is just how github displays diff files.

cqc-melf: Ah, yes, thank you.

cqc-melf (approved):

PR #363, opened 2022-05-17
Cache diameter returned by `Architecture::get_diameter()`

sjdilkes:
In its underlying graph classes. The graph classes finds the graph diameter by getting the distance between every pair of nodes in the graph, tracking which is largest and then returning it. This value isn't cached, meaning every time `get_diameter` is called this process is repeated, something which is done liberally in routing and a cause of some nasty overhead. This small fix massively improves routing runtime, especially on larger connectivity graphs.


yao-cqc (approved):
LGTM

PR #364, opened 2022-05-17
Fix manylinux builds.

cqc-alec:
Tested [here](https://github.com/CQCL/tket/runs/6468932093).
Cf. failing build [here](https://github.com/CQCL/tket/runs/6468838914).

### DIFF
--- a/.github/workflows/linuxbuildwheel
+++ b/.github/workflows/linuxbuildwheel
@@ -36,6 +36,9 @@ cd /tket/pytket
 mkdir wheelhouse
 mkdir audited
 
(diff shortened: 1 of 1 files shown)


.github/workflows/linuxbuildwheel
```diff
@@ -36,6 +36,9 @@ cd /tket/pytket
 mkdir wheelhouse
 mkdir audited
 
+# Work around https://github.com/pypa/manylinux/issues/1309
```
cqc-melf, line 39: Should we add an issue to remove this after the problem on pypa is solved?

cqc-melf (approved):

PR #368, opened 2022-05-19
[feature] TK2 gate decomposition

lmondada:


lmondada:
It's currently difficult to run proptests, as `DecomposeTK2` will fail unless the angles are normalised. I have reduced for now the generation of TK2 angles in the proptests to normalised angles. This can be removed once the `NormalisedTK2` predicate is implemented.

tket/src/Circuit/include/Circuit/CircPool.hpp
```diff
@@ -237,9 +237,35 @@ Circuit ZZPhase_using_TK2(Expr alpha);
...
 Circuit ZZPhase_using_CX(Expr alpha);
 
-/** Equivalent to TK2, using CX and single-qubit gates */
+/** Equivalent to XXPhase, using ZZPhase and H gates */
+Circuit XXPhase_using_ZZPhase(Expr alpha);
+
+/** Equivalent to YYPhase, using ZZPhase and V/Vdg gates */
+Circuit YYPhase_using_ZZPhase(Expr alpha);
+
+/** Equivalents to TK2, using CX and single-qubit gates, using 1, 2 or 3 CX.
```
cqc-alec, line 246: Please can each function have its own doxygen comment (even if this means some copy-pasting between them).
Same for below.

tket/src/Circuit/CircUtils.cpp
```diff
@@ -105,6 +105,13 @@ Eigen::Matrix4cd get_matrix_from_2qb_circ(const Circuit &circ) {
           }
           break;
         }
+        case OpType::TK2: {
+          auto params = o->get_params();
+          TKET_ASSERT(params.size() == 3);
```
cqc-alec, line 110: I don't fully understand this function, but it seems that the CX code has more complex logic (in terms of `uqb ==  0` and the endianness switch) than the TK2 code. Why is this?

tket/src/Circuit/CircUtils.cpp
```diff
@@ -230,18 +220,34 @@ static std::pair<Eigen::Matrix4cd, Complex> decompose_VD(
   return {V, z0};
 }
 
+void replace_TK2_2CX(Circuit &circ) {
```
cqc-alec, line 223: Should make this `static`.

tket/src/Predicates/PassGenerators.cpp
```diff
@@ -548,6 +548,22 @@ PassPtr KAKDecomposition(double cx_fidelity) {
...
+PassPtr DecomposeTK2(const Transforms::TwoQbFidelities& fid) {
+  Transform t = Transforms::decompose_TK2(fid);
+  const PredicatePtrMap no_precons;
+  PredicateClassGuarantees preserve_all;
+  PostConditions postcons{{}, preserve_all, Guarantee::Preserve};
+  nlohmann::json j;
+  j["name"] = "DecomposeTK2";
+  nlohmann::json fid_json;
+  fid_json["CX"] = fid.CX_fidelity;
+  fid_json["ZZPhase"] = "SERIALIZATION OF FUNCTIONS IS NOT YET SUPPORTED";
```
cqc-alec, line 561:
```suggestion
  fid_json["ZZPhase"] = "SERIALIZATION OF FUNCTIONS IS NOT SUPPORTED";
```
Not sure if it will ever be.

tket/src/Predicates/include/Predicates/PassGenerators.hpp
```diff
@@ -119,6 +120,29 @@ PassPtr gen_user_defined_swap_decomp_pass(const Circuit& replacement_circ);
  */
 PassPtr KAKDecomposition(double cx_fidelity = 1.);
 
+/**
+ * @brief Decomposes each TK2 gate into two-qubit gates.
```
cqc-alec, line 124: Please explain how the target gate is selected.

tket/src/Transformations/BasicOptimisation.cpp
```diff
@@ -273,7 +273,12 @@ static bool replace_two_qubit_interaction(
   Subcircuit sub = {in_edges, out_edges, i.vertices};
   Circuit subc = circ.subcircuit(sub);
   Eigen::Matrix4cd mat = get_matrix_from_2qb_circ(subc);
-  Circuit replacement = two_qubit_canonical(mat, cx_fidelity);
+  Circuit replacement = two_qubit_canonical(mat);
+  // TODO: for now we decompose all the way to CX. In the future, this pass
+  // should output TK2, and decompose to CX (or other gates) later if necessary.
+  TwoQbFidelities fid;
```
cqc-alec, line 279:
```suggestion
  TwoQbFidelities fid = {0,0, 0};
```
Not sure if these will be initialized by default.

cqc-alec (changes requested):
Sorry, not got all the way through before going on holiday, so publishing the comments I have so far...!

pytket/binders/circuit/library.cpp
```diff
@@ -36,6 +36,22 @@ void init_library(py::module &m) {
...
       "_CX_using_TK2", &CircPool::CX_using_TK2,
       "Equivalent to CX, using a TK2 and single-qubit gates");
+  library_m.def(
+      "_TK2_using_CX", &CircPool::TK2_using_CX,
+      "Equivalent to TK2, using up to 3 CX gates and single-qubit gates");
+  library_m.def(
+      "_approx_TK2_using_1xCX", &CircPool::approx_TK2_using_1xCX,
+      "Approximate equivalent to TK2, using 1 CX gates and single-qubit gates. "
+      "No TK2 gate angle is required for this approximation (it is assumed that"
+      " `NormalisedTK2Predicate` is satisfied).");
```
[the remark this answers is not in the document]
lmondada: Done.

tket/src/Circuit/CircPool.cpp
```diff
@@ -838,21 +838,91 @@ Circuit ZZPhase_using_CX(Expr alpha) {
...
-  c.add_op<unsigned>(OpType::Z, {0});
   c.add_op<unsigned>(OpType::Vdg, {0});
+  c.add_op<unsigned>(OpType::Vdg, {1});
+  c.add_op<unsigned>(OpType::ZZPhase, alpha, {0, 1});
+  c.add_op<unsigned>(OpType::V, {0});
   c.add_op<unsigned>(OpType::V, {1});
+  return c;
+}
+
+Circuit approx_TK2_using_1xCX() {
```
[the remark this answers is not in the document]
lmondada: I've improved docs in `CircPool.hpp`

tket/src/Circuit/include/Circuit/CircPool.hpp
lmondada: Done.

tket/src/Circuit/CircUtils.cpp
lmondada: Done.

tket/src/Predicates/PassGenerators.cpp
lmondada: Done.

tket/src/Transformations/Decomposition.cpp
```diff
@@ -419,15 +422,327 @@ Transform decompose_MolmerSorensen() {
...
+        Circuit sub = CircPool::XXPhase_using_ZZPhase(g->get_params()[0]);
+        circ.substitute(sub, v, Circuit::VertexDeletion::No);
+        bin.push_back(v);
+      } else if (circ.get_OpType_from_Vertex(v) == OpType::YYPhase) {
+        success = true;
+        const Op_ptr g = circ.get_Op_ptr_from_Vertex(v);
+        Circuit sub = CircPool::YYPhase_using_ZZPhase(g->get_params()[0]);
+        circ.substitute(sub, v, Circuit::VertexDeletion::No);
+        bin.push_back(v);
       }
```
drzenharper, line 742: I think it should be a `switch` statement on `circ.get_OpType_from_Vertex(v)` instead of `if` clauses.

(_This would also solve the problem of `circ.get_OpType_from_Vertex(v)` being recalculated multiple times, which is trivial, but unnecessary_).

tket/src/Utils/MatrixAnalysis.cpp
```diff
@@ -292,173 +292,55 @@ Eigen::VectorXcd apply_qubit_permutation(
...
 
+// How close TK2(a,b,c) is from the identity
+double trace_fidelity(double a, double b, double c) {
+  constexpr double g = PI / 2;
+  a *= g;
+  b *= g;
+  c *= g;
+  double trace_sq = 16. * (pow(cos(a) * cos(b) * cos(c), 2) +
+                           pow(sin(a) * sin(b) * sin(c), 2));
+  return (4. + trace_sq) / 20.;
```
drzenharper, line 303: I'd prefer writing `0.2 + 0.8 * ...` directly, although it's not important

tket/src/Utils/MatrixAnalysis.cpp
```diff
@@ -292,173 +292,55 @@ Eigen::VectorXcd apply_qubit_permutation(
...
   // gate fidelity achievable with 0,...,3 cnots
   // this is fully determined by the information content k and is optimal
   // see PhysRevA 71.062331 (2005) for more details on this
   switch (nb_cx) {
     case 0:
-      return fidelity(a, b, c);
+      return trace_fidelity(a, b, c);
     case 1:
-      return fidelity(PI / 4 - a, b, c);
+      return trace_fidelity(0.5 - a, b, c);
```
drzenharper, line 321: Is this doing the same as before? Before it was multiplying by `-pi/2`, now you are multiplying by `+pi/2`. That doesn't matter for `a,b,c` of course because `sin` etc. are squared, but you are also modifying `a`.

I worry if it should be `0.5+a` instead of `0.5-a`. I assume there are tests somewhere else with specific values to check this? (_With non-symmetric "random" values like `a=0.12345` etc._) We should not rely on algebra alone!

tket/src/Utils/MatrixAnalysis.cpp
```diff
@@ -292,173 +292,55 @@ Eigen::VectorXcd apply_qubit_permutation(
...
     case 2:
-      return fidelity(0, 0, c);
+      return trace_fidelity(0, 0, c);
     case 3:
       return 1.;
   }
   // never happens
   return -1.;
```
drzenharper, line 328: I'd prefer to delete this, and change `case 3:` to `default:` above; you already have `TKET_ASSERT(nb_cx < 4);` above so I think it's acceptable.

tket/src/Utils/MatrixAnalysis.cpp
```diff
@@ -525,13 +407,13 @@ get_information_content(const Eigen::Matrix4cd &X) {
...
+      (-.5 * basis_change * thetas / PI).head<3>().unaryExpr([](double d) {
+        return mod(d, 4);
       });
 
-  // move k into Weyl chamber ie pi/4 >= k_x >= k_y >= |k_z|
+  // move k into Weyl chamber ie 1/2 >= k_x >= k_y >= |k_z|
   //   1. permutate ks
-  //   2. modulo pi/2 and pi/4
+  //   2. modulo 1. and 1/2
   std::vector<int> ind_order{0, 1, 2};
```
drzenharper, line 417: Not your code, but looks to me like `ind_order` could be a `std::array` instead

tket/tests/test_TwoQubitCanonical.cpp
```diff
@@ -43,6 +44,33 @@ static void check_get_information_content(const Eigen::Matrix4cd &U) {
...
+    c.add_op<unsigned>(OpType::Vdg, {0});
+    c.add_op<unsigned>(OpType::H, {1});
+    c.add_op<unsigned>(OpType::Tdg, {1});
+    c.add_op<unsigned>(OpType::CX, {1, 0});
+  }
+
+  THEN("The unitaries from tket_sim and get_matrix are identical") {
+    auto u1 = tket_sim::get_unitary(c);
+    auto u2 = get_matrix_from_2qb_circ(c);
+    REQUIRE(u1.isApprox(u2));
```
drzenharper, line 70: You're not testing that the unitaries are "identical", only that they are "approximately equal".

Which do you actually mean? If one of the functions `tket_sim::get_unitary` or `get_matrix_from_2qb_circ` is a wrapper around the other, then "identical" would be worth testing (_i.e., exact equality of doubles, normally very naughty but justified in rare cases_); but if they're calculated using different but algebraically equivalent formulae then approximations as here are the correct thing to test, since there could always be roundoff.

tket/tests/test_TwoQubitCanonical.cpp
```diff
@@ -129,7 +157,7 @@ SCENARIO("Testing two-qubit canonical forms") {
     circ.add_op<unsigned>(tket::OpType::Rz, 0.5, {1});
     circ.add_op<unsigned>(tket::OpType::Rz, 0.5, {0});
     circ.add_op<unsigned>(tket::OpType::Rx, 1.2, {1});
-    Eigen::Matrix4cd U = get_matrix_from_2qb_circ(circ);
+    Eigen::Matrix4cd U = tket_sim::get_unitary(circ);
```
drzenharper, line 160: This is actually changing the test (a different function). Does the function `get_matrix_from_2qb_circ` still work? Should do, surely! If so, could you add a check that the unitaries produced are almost identical?

drzenharper (changes requested):
Sorry, just some minor nitpicks

tket/src/Circuit/include/Circuit/CircPool.hpp
```diff
@@ -237,33 +237,115 @@ Circuit ZZPhase_using_TK2(Expr alpha);
...
 Circuit TK2_using_3xCX(Expr alpha, Expr beta, Expr gamma);
-// Choose which of the above decompositions to use. Decomposition is exact
+
+/**
+ * @brief Equivalent to TK2(α, β, γ) with minimal number of CX gates.
+ *
+ * A TK2-equivalent circuit with as few CX gates as possible (0, 1, 2 or 3 CX).
+
+ * Decomposition is exact. The parameters must be normalised to the Weyl
+ * chamber, i.e. it must hold 0.5 ≥ 𝛼 ≥ 𝛽 ≥ |𝛾|.
```
drzenharper, line 298: Do we care if these inequalities fail due to roundoff? Maybe not worth worrying about, but what if e.g. `alpha = beta` algebraically, but due to roundoff error we get `alpha = 0.1` and `beta = 0.10000000000001` ?

pytket/binders/passes.cpp
```diff
@@ -33,6 +34,26 @@ using json = nlohmann::json;
 
 namespace tket {
 
+// given keyword arguments for DecomposeTK2, return a TwoQbFidelities struct
+Transforms::TwoQbFidelities get_fidelities(const py::kwargs &kwargs) {
+  Transforms::TwoQbFidelities fid;
+  for (const auto kwarg : kwargs) {
```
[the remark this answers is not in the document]
lmondada: No you're right. Done.

tket/src/Circuit/include/Circuit/CircPool.hpp
```diff
@@ -237,9 +237,35 @@ Circuit ZZPhase_using_TK2(Expr alpha);
 /** Equivalent to ZZPhase, using CX and Rz gates */
 Circuit ZZPhase_using_CX(Expr alpha);
 
-/** Equivalent to TK2, using CX and single-qubit gates */
+/** Equivalent to XXPhase, using ZZPhase and H gates */
+Circuit XXPhase_using_ZZPhase(Expr alpha);
```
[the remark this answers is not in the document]
lmondada: You're right, I've done a nice find&replace all.

tket/src/Transformations/Decomposition.cpp
```diff
@@ -419,15 +422,327 @@ Transform decompose_MolmerSorensen() {
   });
 }
 
+static double get_ZZPhase_fidelity(
+    const std::array<double, 3> &k, unsigned nb_cx) {
+  auto [a, b, c] = k;
```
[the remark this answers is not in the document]
lmondada: Done.

tket/src/Transformations/Decomposition.cpp
```diff
@@ -419,15 +422,327 @@ Transform decompose_MolmerSorensen() {
...
+    default:
+      return 1.;
+  }
+}
+
+// Try to decompose a TK2 gate using different gate sets, find the one with
+// the highest fidelity.
+// If no fidelities are provided, (best_optype, n_gates) is left unchanged.
+static void best_noise_aware_decomposition(
+    std::array<double, 3> angles, TwoQbFidelities fid, OpType &best_optype,
```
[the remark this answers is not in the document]
lmondada: Done.

tket/src/Transformations/Decomposition.cpp
```diff
@@ -419,15 +422,327 @@ Transform decompose_MolmerSorensen() {
...
+  bool is_symbolic = true;
+  double last_val = .5;
+  for (unsigned i = 0; i < k.size(); ++i) {
+    std::optional<double> eval = eval_expr_mod(k[i], 4);
+    if (eval) {
+      is_symbolic = false;
+      double abs_eval = std::min(*eval, -(*eval) + 4);
+      if (i + 1 == k.size()) {
+        if (abs_eval > last_val) {
+          return false;
```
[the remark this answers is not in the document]
lmondada: Sure thing!

tket/src/Transformations/Decomposition.cpp
```diff
@@ -419,15 +422,327 @@ Transform decompose_MolmerSorensen() {
...
+      n_gates > 1) {
+    n_gates = 1;
+    best_optype = OpType::ZZPhase;
+  }
+}
+
+// Whether the TK2 angles are in the Weyl chamber.
+//
+// For numerical values: 1/2 >= k_x >= k_y >= |k_z|
+// Further, we demand that all symbolic values come before numerical values.
```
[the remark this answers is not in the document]
lmondada: Yes you are right. I changed it to `Whether the TK2 angles are normalised`, and then rewrote the details in the main body of the documentation.

tket/src/Utils/MatrixAnalysis.cpp
cqc-alec, line 303: Bit confused by this. The result is always at least 0.2, even when a=b=c=0, which equates to the identity?
And if say cos(a)=0 and sin(b)=0, the result is exactly 0.2, independent of c -- which also seems wrong.

tket/src/Utils/MatrixAnalysis.cpp
```diff
@@ -570,41 +452,40 @@ get_information_content(const Eigen::Matrix4cd &X) {
       Eigen::kroneckerProduct(PauliX, Eigen::Matrix2cd::Identity());
   const Eigen::Matrix4cd s_ix =
       Eigen::kroneckerProduct(Eigen::Matrix2cd::Identity(), PauliX);
-  if (k(0) > PI / 2) {
-    k(0) -= 1.5 * PI;
-    K1 *= -i_ * s_xx;
+  if (k(0) > 1.) {
+    k(0) -= 3.;
+    K1 *= i_ * s_xx;
```
cqc-alec, line 457: Why did the sign of `K1` change here (and in a few other places below)?

tket/src/Transformations/include/Transformations/Decomposition.hpp
```diff
@@ -92,8 +92,51 @@ Transform decompose_ZX_to_HQS1();
...
 Transform decompose_MolmerSorensen();
 
-// finds all ZZ gates which are in the form of CX and Rz gates and converts
-// into ZZ gates Expects: CX, Rz Produces: ZZ, CX, Rz
+/**
+ * @brief A simple struct to store some two-qubit gate fidelities.
+ *
+ * We currently support CX, ZZMax and ZZPhase.
+ *
+ * In the Python layer, this struct is replaced by keyword arguments.
```
cqc-alec, line 100: I'd delete this. (The C++ layer shouldn't document what the Python layer does.)

cqc-alec (changes requested):

tket/src/Transformations/Decomposition.cpp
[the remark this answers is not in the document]
lmondada: `best_optype` and `n_gates` are passed by reference because they are modified.

tket/src/Transformations/Decomposition.cpp
```diff
@@ -419,15 +422,327 @@ Transform decompose_MolmerSorensen() {
...
+    VertexList bin;
+    BGL_FORALL_VERTICES(v, circ.dag, DAG) {
+      if (circ.get_OpType_from_Vertex(v) != OpType::TK2) continue;
+
+      success = true;
+      auto params = circ.get_Op_ptr_from_Vertex(v)->get_params();
+      TKET_ASSERT(params.size() == 3);
+      std::array<Expr, 3> angles{params[0], params[1], params[2]};
+
+      Circuit sub = TK2_replacement(angles, fid);
```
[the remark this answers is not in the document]
lmondada: Done.

tket/src/Transformations/Decomposition.cpp
```diff
@@ -419,15 +422,327 @@ Transform decompose_MolmerSorensen() {
...
     bool success = decompose_PhaseGadgets().apply(circ);
+    VertexList bin;
     BGL_FORALL_VERTICES(v, circ.dag, DAG) {
       if (circ.get_OpType_from_Vertex(v) == OpType::PhaseGadget) {
         const Op_ptr g = circ.get_Op_ptr_from_Vertex(v);
         circ.dag[v] = {get_op_ptr(OpType::ZZPhase, g->get_params()[0])};
+      } else if (circ.get_OpType_from_Vertex(v) == OpType::XXPhase) {
+        success = true;
+        const Op_ptr g = circ.get_Op_ptr_from_Vertex(v);
+        Circuit sub = CircPool::XXPhase_using_ZZPhase(g->get_params()[0]);
```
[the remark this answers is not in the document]
lmondada: I've added an assertion.

tket/src/Transformations/Decomposition.cpp
lmondada: Done!

tket/src/Utils/MatrixAnalysis.cpp
lmondada: Yes, we're safe. But good question. Basically, after the sign swap we project `a` back into the `[0, 0.5]` interval. Now `-0.5` happens to be projected to `0.5`, so this is correct.

tket/src/Utils/MatrixAnalysis.cpp
lmondada: Done.

tket/src/Utils/MatrixAnalysis.cpp
lmondada: Done.

tket/tests/test_TwoQubitCanonical.cpp
lmondada: No, they are different methods to calculate the same. I considered replacing `get_matrix_from_2qb_circ` with `tket_sim`, but it would create a dependency on `Simulation` for basically every tket module. And probably some circular dependencies as a consequence too.

tket/tests/test_TwoQubitCanonical.cpp
lmondada: My thinking was to use `tket_sim::get_unitary` everywhere possible in the tests, for consistence. That `get_matrix_from_2qb_circ` behaves correctly is already checked in the other tests I added (lines 47-72).

tket/tests/test_TwoQubitCanonical.cpp
```diff
@@ -158,7 +186,7 @@ SCENARIO("Testing two-qubit canonical forms") {
     circ.add_op<unsigned>(tket::OpType::Rz, 0.5, {1});
     circ.add_op<unsigned>(tket::OpType::CX, {1, 0});
     circ.add_op<unsigned>(tket::OpType::Rz, 1.2, {0});
-    Eigen::Matrix4cd U = get_matrix_from_2qb_circ(circ);
+    Eigen::Matrix4cd U = tket_sim::get_unitary(circ);
```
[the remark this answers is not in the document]
lmondada: Same remark as above.

tket/tests/test_TwoQubitCanonical.cpp
```diff
@@ -217,117 +245,20 @@ SCENARIO("Testing two-qubit canonical forms") {
...
-      phase += angles_q0.back() + angles_q1.back();
-      if (it + 1 != gates.end()) {
-        circ_out.add_op<unsigned>(OpType::CX, {0, 1});
-      }
-    }
-    Eigen::Matrix4cd out =
-        exp(i_ * phase * PI) * get_matrix_from_2qb_circ(circ_out);
-    REQUIRE(out.isApprox(U));
-  }
-
```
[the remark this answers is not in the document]
lmondada: These tests really are useless. The way I've refactored the code, these tests would either test nothing at all or test exactly the same as some of the other tests I've written.

tket/src/Utils/MatrixAnalysis.cpp
lmondada: This formula comes straight from https://arxiv.org/pdf/1811.12926.pdf. See equations B3 and B5.

Regarding your two specific examples:

- If `a=b=c=0`, then `cos(a)cos(b)cos(c) = 1` and hence the fidelity is `1.0` as it should be.
- `a, b, c` must be in the Weyl chamber. `sin(b) = 0` ⇒ `b = 0` ⇒ `c = 0`. Hence the fidelity doesn't need to depend on `c`.
 
I have added a comment specifying that a,b,c must be in the Weyl chamber. Is there anything else that I am missing?

tket/src/Utils/MatrixAnalysis.cpp
lmondada: Because the new `k(0)` has an additional sign flip compared to the old `k(0)` (the change of sign of the matrix `P` higher up (line 27)). Hence projecting `k(0)` into the Weyl chamber requires a "move" in the opposite direction.

I spent a lot of time getting these operations right, I'm fairly confident they are correct now.

tket/src/Transformations/include/Transformations/Decomposition.hpp
lmondada: Done.

pytket/binders/transform.cpp
```diff
@@ -185,6 +205,35 @@ PYBIND11_MODULE(transform, m) {
...
+          },
+          "Decompose each TK2 gate into two-qubit gates."
+          "\n\nWe currently support CX, ZZMax and ZZPhase."
+          "\n\nIf one or more gate fidelities are provided, the two-qubit gate "
+          "type achieving the highest fidelity will be chosen for the "
+          "decomposition, as measured using squared trace fidelity. "
+          "If no fidelities are provided, the TK2 gates will be decomposed "
+          "exactly using CX gates.\n\n"
+          "All TK2(α, β, γ) gates must be normalised to the Weyl chamber, i.e. "
+          "0.5 ≥ 𝛼 ≥ 𝛽 ≥ |𝛾|."
```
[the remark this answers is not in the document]
lmondada: Should be fixed.

tket/src/Transformations/Decomposition.cpp
```diff
@@ -419,15 +422,327 @@ Transform decompose_MolmerSorensen() {
...
+ * provided.
+ *
+ * Symbolic parameters are supported. In that case, decompositions are exact.
+ *
+ * @param angles The TK2 parameters
+ * @param fid The two-qubit gate fidelities
+ * @return Circuit TK2-equivalent circuit
+ */
+static Circuit TK2_replacement(
+    std::array<Expr, 3> angles, const TwoQbFidelities &fid) {
```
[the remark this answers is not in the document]
lmondada: Done.

pytket/tests/transform_test.py
```diff
@@ -695,7 +695,7 @@ def test_symbol_pauli_squash_1() -> None:
         u1 = c1.get_unitary()
         # PauliSquash does not preserve global phase.
         v = u @ u1.conjugate().transpose()
-        assert np.allclose(v, v[0, 0] * np.eye(4, 4, dtype=complex))
+        assert np.allclose(v, v[0, 0] * np.eye(4, 4, dtype=complex), atol=1e-7)
```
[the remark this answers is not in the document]
lmondada: This has to do (as so often) with exploding SymEngine symbolic expressions. It is indeed a rounding error. I'm trying to get to the root cause of this.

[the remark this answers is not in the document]
lmondada: Should be fixed now.

tket/src/Utils/MatrixAnalysis.cpp
lmondada: Sorry, I didn't explain myself properly: "I'm confident they are correct _now_" refers to earlier as I was developing this. It took a fair bit of algebra + testing to make sure every case was covered properly. This is what all the tests on lines 1773 in `test_Synthesis.cpp` and below are testing. If you look at the coverage, you will indeed see that all the cases are covered.

lmondada:
I've added a `minus_times` function that commutes `-expr` and tries to simplify with `SymEngine::expand` when possible. I hope this doesn't make anyone unhappy.

It helps a lot with shortening expressions of the type `1/2 - (1/2 - (... - (1/2 - alpha))))` that would creep up from `squash_1qb_to_tk1`, because it would repeatedly `.dagger()` and un-`.dagger()` expressions.

This wasn't what actually solved the rounding error I had. It was adding `squash_1qb_to_tk1` at the end of `decompose_TK2` that did it. But I figured I might as well solve the other problem that was apparent from the test cases I ran.

tket/src/Utils/MatrixAnalysis.cpp
cqc-alec: You're right, I was confused by trying to read it as distance from the identity, which is what the comment above makes it sound like. Perhaps that comment could be clarified?

tket/src/Utils/Expression.cpp
```diff
@@ -117,6 +117,16 @@ Expr sin_halfpi_times(const Expr& e) {
...
 }
 
+Expr minus_times(const Expr& e) {
+  Expr e1 = -e;
+  Expr e2 = SymEngine::expand(e1);
+
+  unsigned e1_size = e1.get_basic()->dumps().size();
+  unsigned e2_size = e2.get_basic()->dumps().size();
+
+  return e2_size < e1_size ? e2 : e1;
```
cqc-alec, line 127: Are there cases where `e2_size > e1_size` or can we just return `e2`?

cqc-alec (changes requested):

tket/src/Utils/Expression.cpp
lmondada: You can definitely construct a lot of cases where not `expand`ing would keep the expressions shorter, e.g. `Rx((a + 1)^n).dagger()`.

We could however decide nonetheless to always expand, if that makes things more consistent. I personally don't think it's necessary, but I'm happy to do it if you would prefer that.

cqc-alec: OK, let's leave it!

tket/src/Utils/MatrixAnalysis.cpp
lmondada: Yes, I've removed that comment. I've expanded the doxygen docs in the `hpp`.

lmondada:
We should be good to go. Thanks @cqc-alec and @drzenharper for the reviews.

pytket/binders/circuit/library.cpp
```diff
@@ -190,6 +214,17 @@ void init_library(py::module &m) {
   library_m.def(
       "_ZZPhase_using_CX", &CircPool::ZZPhase_using_CX,
       "Equivalent to ZZPhase, using CX and Rz gates");
+  library_m.def(
+      "_TK2_using_ZZPhase", &CircPool::TK2_using_ZZPhase,
+      "Equivalent to TK2, using 3 ZZPhase gates");
+  library_m.def(
+      "_approx_TK2_using_1xZZPhase", &CircPool::approx_TK2_using_1xZZPhase,
+      "Approximate equivalent to TK2, using 1 ZZPhase gate and single-qubit "
+      "gates. Only requires the first angle of the TK2 gate.");
```
cqc-alec, line 223: I think it is still worth saying that all TK2 angles should be normalized -- just descriptively, without reference to the predicate. (Same below.)

cqc-alec (changes requested):

pytket/binders/circuit/library.cpp
lmondada: Actually, for the decompositions using ZZPhase, it works no matter what the normalisation is.

cqc-alec: Ah OK

cqc-alec (approved):

drzenharper (approved):

PR #369, opened 2022-05-19
Add failsafe to `LexiRouteRoutingMethod`

sjdilkes:
Very rarely `LexiRouteRoutingMethod` (once thousands) can return the same SWAP as best choice repeatedly without realising any two-qubit gates in between, i.e. _infinite routing_ (seemingly for "dense" circuits only). On finding a SWAP, this PR updates `LexiRouteRoutingMethod` to check whether for the pair of edges the SWAP would be wired into, that the source vertex for both these edges is 1) not the same, and 2) if it is the same does not have OpType::SWAP. This condition implies that the same SWAP gate is being executed on the same pair of qubits sequentially without either qubit completing some other two qubit gate, i.e. a redundant SWAP that could lead to infinite routing.

If these conditions are true `MappingFrontier::add_swap` returns `false` and `LexiRoute` finds a path of SWAP gates to SWAP along that guarantees to realise some two-qubit gate. Any proposed SWAP will always have at least one Node with a program qubit in two-qubit gate at the "frontier" being routed on. As this case is so rare, the solution is to find a path between one of these Nodes and the Node the two-qubit program gate is between and then SWAP along a path between them naively.


tket/src/Mapping/LexiRoute.cpp
```diff
@@ -521,7 +521,50 @@ bool LexiRoute::solve(unsigned lookahead) {
...
+        // adds a SWAP between each pair of adjacent nodes on path
+        while (path_it_0 != path.end()) {
+          this->mapping_frontier_->add_swap(*path_it_0, *path_it_1);
+          ++path_it_0;
+          ++path_it_1;
+        }
+      } else {
+        it = this->interacting_uids_.find(chosen_swap.second);
+        if (it == this->interacting_uids_.end()) {
+          throw LexiRouteError("Neither Node in SWAP in interaction.");
```
cqc-alec, line 553: Can this happen?
(If it should be impossible, it should be a `TKET_ASSERT`.)

tket/src/Mapping/LexiRoute.cpp
```diff
@@ -521,7 +521,50 @@ bool LexiRoute::solve(unsigned lookahead) {
...
+        }
+        auto path =
```
cqc-alec, line 555: This code is the same as above so I feel should be refactored.

tket/src/Mapping/LexiRoute.cpp
```diff
@@ -521,7 +521,50 @@ bool LexiRoute::solve(unsigned lookahead) {
   if (!check.first && !check.second) {
     // update circuit with new swap
     // final_labelling is initial labelling permuted by single swap
-    this->mapping_frontier_->add_swap(chosen_swap.first, chosen_swap.second);
+
+    // if false, SWAP is identical to last SWAP added without any gates realised
+    if (!this->mapping_frontier_->add_swap(
```
cqc-alec, line 526: Are there tests that exercise this? (I guess not since it was getting stuck before?) Would be good to add one.

cqc-alec (changes requested):

tket/src/Mapping/LexiRoute.cpp
sjdilkes: I'll add a lambda function

tket/src/Mapping/LexiRoute.cpp
sjdilkes: Will change

sjdilkes: (No it can't)

tket/src/Mapping/LexiRoute.cpp
sjdilkes: The scenario is extremely hard to replicate "naturally": it happened once in the thousands of circuits benchmarked very deep into a very large circuit. I'm fairly happy with the logic that if this is false we have a known workaround and that we know it calls false under the right circumstance (what 526 covers).

sjdilkes: As I'm still running benchmarks I'll generate and benchmark a number of small circuits with what I consider to be the potential to clock this case (i.e. just very dense circuits) and if one hits it that is somewhat small I'll a) double check it would have failed previously and then b) add it as a test case here with the criteria being that it successfully finishes routing.

cqc-alec: Sounds good; you can tag the test as "long" if it takes a long time (> 1s) to run.

tket/src/Mapping/LexiRoute.cpp
sjdilkes: done

tket/src/Mapping/LexiRoute.cpp
sjdilkes: done

tket/src/Mapping/LexiRoute.cpp
sjdilkes: "not done" <- still churning through circuits

cqc-alec: OK. Am keen to have at least one verified run of the new code before merging. Thanks!

cqc-alec (approved):

PR #373, opened 2022-05-19
Extend range of operations recognized as Clifford.

cqc-alec:


tket/src/Transformations/Decomposition.cpp
```diff
@@ -597,12 +597,12 @@ Transform decompose_cliffords_std() {
...
-      if (opT == OpType::TK1 || opT == OpType::U3 || opT == OpType::U2 ||
-          opT == OpType::U1 || opT == OpType::Rx || opT == OpType::Ry ||
-          opT == OpType::Rz || opT == OpType::PhasedX) {
-        const Op_ptr g = circ.get_Op_ptr_from_Vertex(v);
-        std::vector<Expr> tk1_param_exprs = as_gate_ptr(g)->get_tk1_angles();
+      Op_ptr op = circ.get_Op_ptr_from_Vertex(v);
+      OpType type = op->get_type();
+      if (type != OpType::V && type != OpType::S && type != OpType::X &&
+          type != OpType::Z && is_single_qubit_unitary_type(type) &&
+          op->is_clifford()) {
```
yao-cqc, line 604: Do you know how this will affect the `CliffordSimp` pass? (e.g. `H` gates now get converted using `VSXZ`); I think we also need to update the doc strings for `decompose_cliffords_std` and `RebaseToCliffordSingles`.

cqc-alec: You're right that it now changes e.g. the H gate which it didn't before (though that wasn't documented). Yes, I will correct the documentation of this function and its Python binding.

yao-cqc (approved):

PR #378, opened 2022-05-21
Fixes to compile with gcc-11

1tnguyen:
Changes:

(1) Since `tket` is compiled with `-Wall -Werror`, it will fail to compile Eigen due to this [bug](https://gitlab.com/libeigen/eigen/-/issues/2304). Hence, adding `-Wno-maybe-uninitialized` when gcc 11+ is used.

(2) gcc-11 only has changes in its [header dependency](https://gcc.gnu.org/gcc-11/porting_to.html). In particular, `<limits>` (for `std::numeric_limits`) is used less widely and needs to be included explicitly.

Tested by: compiling `tket` following the [instructions](https://github.com/CQCL/tket#how-to-build-tket-and-pytket).

### DIFF
--- a/tket/src/CMakeLists.txt
+++ b/tket/src/CMakeLists.txt
@@ -40,6 +40,10 @@ endif()
 
 set(PROFILE_COVERAGE no CACHE BOOL "Build library with profiling for test coverage")
 IF (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+    # Bug in Eigen with gcc 11: https://gitlab.com/libeigen/eigen/-/issues/2304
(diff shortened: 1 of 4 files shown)


cqc-alec:
Thanks for this! Apart from the minor formatting issue (please install `clang-format` and run the `do-clang-format` script), I just have one comment: we prefer to isolate eigen-specific compilation flags as `#pragma` directives in [this header file](https://github.com/CQCL/tket/blob/develop/tket/src/Utils/include/Utils/EigenConfig.hpp), rather than in the top-level `CMakeLists,txt`. This file is included by any source file that uses eigen.

1tnguyen:
Thanks for the comments. I have fixed the code formatting issue and used the `pragma` directive as suggested.

cqc-alec (approved):

PR #386, opened 2022-05-25
redo changes to placement, without merging dev first

drzenharper:


drzenharper:
The Mac test failure is due to a timeout on one problem. I will increase the timeout parameter a bit (on the main branch) - it can just be ignored here.

PR #387, opened 2022-05-25
Feature/update circuit rendering

DNA386:
Update the circuit rendering to use the new pytket-circuit-render module. At the moment it is set to load the latest v0.1.x, so there should be no need to update it every time the circuit render gets a minor change. 
The api to call the renderer is mostly the same except there is no longer a need to specify render options in advance as they are dealt with by the display itself.


cqc-alec (approved):
LGTM!

PR #388, opened 2022-05-25
change CustomGate is_equal function; check and throw for null input p…

drzenharper:
…tr in constructor


tket/src/Circuit/Boxes.cpp
```diff
@@ -285,14 +285,37 @@ bool CompositeGateDef::operator==(const CompositeGateDef &other) const {
...
-    : Box(OpType::CustomGate, gate->signature()), gate_(gate), params_(params) {
+    : Box(OpType::CustomGate, get_non_null_ptr(gate)->signature()),
+      gate_(gate),
+      params_(params) {
   if (params_.size() != gate_->n_args()) throw InvalidParameterCount();
 }
 
 CustomGate::CustomGate(const CustomGate &other)
-    : Box(other), gate_(other.gate_), params_(other.params_) {}
+    // Should be impossible to get a null gate ptr, but check anyway.
```
cqc-alec, line 306: If it should be impossible then we should `TKET_ASSERT`, not throw. (But I'm not sure even that is worthwhile.)

That also means we only call `get_non_null_ptr` in one place, so it doesn't need to be a separate function and the check should just be integrated into the other constructor.

cqc-alec (changes requested):

tket/src/Circuit/Boxes.cpp
drzenharper: It's impossible currently, but as soon as someone writes a new constructor and forgets to ensure that gate_ is not null, it will become possible again (admittedly unlikely, since I hope they would look at the current constructor and see the problem).

Perhaps we should consider using a not_null template?

cqc-alec: Let's cross that bridge when we come to it. In any case the copy constructor should not have to repeat validation of the data.

tket/src/Circuit/Boxes.cpp
```diff
@@ -285,15 +285,35 @@ bool CompositeGateDef::operator==(const CompositeGateDef &other) const {
   return this->get_def()->circuit_equality(*other.get_def(), {}, false);
 }
 
+static const composite_def_ptr_t &get_non_null_ptr(
```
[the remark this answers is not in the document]
drzenharper: It could be. Currently the `CustomGate` constructor calls the `Box` constructor, which requires a `signature` coming from the `gate->signature()` function.

Thus we'd have to call the `Box` constructor without a  `signature` object, THEN do the non-null check on `gate` and set the `op_signature_t signature_` to `gate->signature()` (which is just a vector) within the body of the `CustomGate` constructor.

This might slow it down (although negligibly), because a useless empty `signature` vector is constructed to pass into the `Box` constructor (although, if we're lucky, the compiler would optimise it away).

However I note that `Box` is already effectively doing that because `signature = {}` is a default argument `Box` constructor, so already a useless empty `signature` vector must be constructed (unless the compiler optimises it away).

So, I'll do that if you're happy with the possible performance hit.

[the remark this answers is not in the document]
cqc-alec: I think that's fine.

Also note that you will need to do a merge from `develop` to get the Linux tests to pass.

cqc-alec (dismissed):

tket/tests/Circuit/test_Boxes.cpp
```diff
@@ -855,5 +855,51 @@ SCENARIO("Checking equality", "[boxes]") {
...
+    composite_def_ptr_t def = CompositeGateDef::define_gate(prefix, setup, {a});
+    CustomGate g(def, {0.4444});
+    const std::string name = g.get_name();
+    CHECK(prefix != name);
+    CHECK_THAT(name, Catch::Matchers::StartsWith(prefix));
+
+    // Of course a bit naughty, it relies on the default precision
+    // not having too many decimal places (as 0.4444 is NOT exactly
+    // represented by a double!)
+    CHECK(name == "gate with params(0.4444)");
```
cqc-alec, line 882: How about just
```
CHECK_THAT(name, Catch::Members::StartsWith(prefix.append("(0.444"))));
```
combining both tests and sidestepping the floating-point issue.

tket/tests/Circuit/test_Boxes.cpp
```diff
@@ -855,5 +855,51 @@ SCENARIO("Checking equality", "[boxes]") {
...
+    Expr eb(b);
+    Expr ec(c);
+    setup.add_op<unsigned>(OpType::TK1, {ea, eb, ec}, {0});
+    const std::string prefix("gate with 3 params");
+    composite_def_ptr_t def =
+        CompositeGateDef::define_gate(prefix, setup, {a, b, c});
+    CustomGate g(def, {0.1111, 0.2222, 0.4444});
+    const std::string name = g.get_name();
+    CHECK(prefix != name);
+    CHECK_THAT(name, Catch::Matchers::StartsWith(prefix));
```
cqc-alec, line 899: These two checks are implied by the one below.

cqc-alec (commented):
Couple of redundant tests to delete still?

PR #391, opened 2022-06-07
Verify metadata before uploading to pypi.

cqc-alec:


.github/workflows/release.yml
```diff
@@ -394,8 +394,6 @@ jobs:
         user: __token__
         password: ${{ secrets.PYPI_PYTKET_API_TOKEN }}
         verbose: true
-        # TODO Remove this line when long_description added:
-        verify_metadata: false
```
cqc-melf, line 398: I assume that this is automatically set to true?

cqc-melf (approved):
Is this also checked for all the extensions?

cqc-alec:
> Is this also checked for all the extensions?

Yes, the extensions have this check.

.github/workflows/release.yml
cqc-alec: Yes: https://github.com/pypa/gh-action-pypi-publish#disabling-metadata-verification

PR #396, opened 2022-06-09
Update test suite to catch2 3.0.1

cqc-alec:
Version 3 of catch2 is no longer header-only, so we link against the static library.

Instead of including `catch.hpp`, just include `catch_test_macros.hpp` and any other headers needed.

Also update the version of rapidcheck used for the proptests.


cqc-alec:
> Shame it's no longer header only!

Should improve build times a bit though.

PR #398, opened 2022-06-09
[bugfix] Leftover graphviz output in tests

lmondada:


cqc-alec (dismissed):

lmondada:
I've had to improve test coverage (because the stray `graphviz` function happened to not be tested).

cqc-alec (approved):

PR #407, opened 2022-06-20
Replace boost vf2 with WSM module in GraphPlacement

sjdilkes:
This commit preserves the current edge breaking routine, but replaces the boost vf2 call with a call to the new weighted subgraph monomorphism module instead.


tket/src/ArchAwareSynth/Path.cpp
```diff
@@ -165,24 +151,25 @@ std::list<unsigned> PathHandler::find_path(unsigned i, unsigned j) {
...
+  Architecture::UndirectedConnGraph undirected_target =
+      arch.get_undirected_connectivity();
+  QubitGraph::UndirectedConnGraph undirected_pattern =
+      q_graph.get_undirected_connectivity();
+  std::vector<qubit_bimap_t> all_maps = get_unweighted_subgraph_monomorphisms(
+      undirected_pattern, undirected_target, 1, timeout);
 
   /* Architecture has no hampath, sad. */
-  if (!found_monomorphism) return {};
+  if (all_maps.size() == 0) return {};
```
cqc-alec, line 172:
```suggestion
  if (all_maps.empty()) return {};
```

schemas/compiler_pass_v1.json
```diff
@@ -834,13 +834,13 @@
         "timeout": {
           "type": "integer",
           "minimum": 0,
-          "description": "Timeout for vf2 search, in seconds."
+          "description": "Timeout for monomorphism search, in milliseconds."
```
cqc-alec, line 837: I assume this was a bug, we aren't changing it?

recipes/tket/conanfile.py
```diff
@@ -106,29 +102,9 @@ def configure(self):
...
-            tools.patch(base_path=boost_include_path, patch_file=patch_file)
         cmake = self._configure_cmake()
-        try:
-            print("Building")
-            cmake.build()
-        finally:
-            for filepath in patches.keys():
-                print("Restoring " + filepath)
-                shutil.move(filepath + ".original", filepath)
+        print("Building")
```
cqc-alec, line 106: I'd delete this print line as well as it isn't useful now.

cqc-alec (changes requested):

tket/src/ArchAwareSynth/Path.cpp
sjdilkes: Done

schemas/compiler_pass_v1.json
sjdilkes: Yeah the json was incorrect previously

recipes/tket/conanfile.py
sjdilkes: done

cqc-alec (approved):

PR #409, opened 2022-06-21
Coerce ZZPhase angles to (-pi, pi) 

sjdilkes:
Also adds a new pass ZZPhaseToRz to convert ZZPhase gates with angle equal to -pi or pi to two Rz gates with angle pi. This is a new restriction from Quantinuum devices.

Does coercion at point of producing qasm string as when accessing pytket.params it automatically reduces the parameters the the canonical range, meaning this information would be lost if completed earlier.


pytket/pytket/qasm/qasm.py
```diff
@@ -1042,7 +1043,17 @@ def circuit_to_qasm_io(
         elif header == "hqslib1" and optype == OpType.ZZPhase:
             # special handling for zzphase
             opstr = "RZZ"
-            params = op.params
+            param = op.params[0]
+            # as op.params returns reduced parameters, we can assume
```
cqc-melf, line 1047: How sure are we about this? Should we do an assert?

tket/src/Predicates/PassLibrary.cpp
```diff
@@ -419,4 +419,21 @@ const PassPtr &NormaliseTK2() {
...
+    Transform t = Transforms::ZZPhase_to_Rz();
+    // GateSetPredicate not preserved because ZZPhase gates may be converted to
+    // Rz gates
+    PredicateClassGuarantees g_postcons = {
+        {typeid(GateSetPredicate), Guarantee::Clear}};
+    PostConditions postcon = {{}, g_postcons, Guarantee::Preserve};
+    PredicatePtrMap precons;
+    // record pass config
+    nlohmann::json j;
+    j["name"] = "ZZPhaseToRz";
```
cqc-melf, line 433: Have you made a testcase checking this in the json generation?

pytket/pytket/qasm/qasm.py
```diff
@@ -1042,7 +1043,17 @@ def circuit_to_qasm_io(
         elif header == "hqslib1" and optype == OpType.ZZPhase:
             # special handling for zzphase
             opstr = "RZZ"
-            params = op.params
```
cqc-melf, line 1045: What happens to the other parameters which are not the first element of the `op.params` list? Or is that only one element anyway?

pytket/binders/passes.cpp
```diff
@@ -776,6 +776,11 @@ PYBIND11_MODULE(passes, m) {
       "transformed circuit (if omitted, an X gate is used)"
       "\n:return: a pass to perform the simplification",
       py::arg("allow_classical") = true, py::arg("xcirc") = nullptr);
+
+  m.def(
+      "ZZPhaseToRz", &ZZPhaseToRz,
```
cqc-melf, line 781: Could you add this to the changelog?

pytket/pytket/qasm/qasm.py
sjdilkes: It's certainly the case at the moment, if it does change then the test in qasm_test.py will fail so it can be corrected.

pytket/pytket/qasm/qasm.py
```diff
@@ -1042,7 +1043,17 @@ def circuit_to_qasm_io(
...
+            param = op.params[0]
+            # as op.params returns reduced parameters, we can assume
+            # that 0 <= param < 4
+            if param > 1:
```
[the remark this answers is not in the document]
sjdilkes: done

tket/src/Predicates/PassLibrary.cpp
sjdilkes: Though as its a standard pass with no special arguments do we need one? It seems that not all passes do have one

cqc-melf: Maybe we don't need one in that case, thought there are testcases for all passes.

pytket/pytket/qasm/qasm.py
cqc-melf: Okay, then keep it as it is

cqc-melf (dismissed):

tket/src/Predicates/PassLibrary.cpp
sjdilkes: and tbf i forgot to add it to compiler_pass_v1.json so probably a good reason to have the tests ha

sjdilkes: added tests + fixed

cqc-melf (approved):

cqc-melf (approved):

PR #411, opened 2022-06-22
Feature/add linear depth cnu decomposition

yao-cqc:
This PR

1. Implements the linear depth CnU decomposition method introduced in https://arxiv.org/abs/2203.11882.
2. Improves the default CnX decomposition.

The next task should be improving CnRy and general QControlBox decomposition. However, the current method supports decomposing symbolic circuits while the new method doesn't. If we don't want to break the API then we might need to modify this to support symbolic matrices.


tket/src/Transformations/ControlledGates.cpp
```diff
@@ -805,6 +805,94 @@ Circuit cnx_gray_decomp(unsigned n) {
...
+    Op_ptr op = std::make_shared<Unitary1qBox>(ub);
+    QControlBox qcb(op);
+    circ.add_box(qcb, {i - 1, n});
+  }
+}
+
+// Add an incrementer without toggling the least significant bit
+// Apply to qubits {0,...,n-1}
+static void add_qn(Circuit& circ, unsigned n) {
+  for (unsigned i = n - 1; i > 1; i--) {
```
cqc-alec, line 845: This breaks horribly if `n == 0`; should `incrementer_linear_depth()` forbid that case or have special handling? (We should be able to handle a CnX gate with 0 controls.)

tket/src/Transformations/ControlledGates.cpp
```diff
@@ -805,6 +805,94 @@ Circuit cnx_gray_decomp(unsigned n) {
...
+    add_pn(circ, i, true);
+  }
+}
+
+// https://arxiv.org/abs/2203.11882 Equation 5
+Circuit incrementer_linear_depth(unsigned n, bool lsb) {
+  Circuit circ(n);
+  add_qn(circ, n);
+  if (lsb) {
+    circ.add_op<unsigned>(OpType::Rx, 1, {0});
```
cqc-alec, line 861: Any reason to use an `Rx(1)` rather than an `X` (with phase adjustment)?

tket/src/Transformations/ControlledGates.cpp
```diff
@@ -805,6 +805,94 @@ Circuit cnx_gray_decomp(unsigned n) {
...
+Circuit cnu_linear_depth_decomp(unsigned n, const Eigen::Matrix2cd& u) {
+  if (!is_unitary(u)) {
+    throw CircuitInvalidity(
+        "Matrix for the controlled operation must be unitary");
+  }
+  Circuit circ(n + 1);
+  // Add pn(u) to qubits {1,...,n}
+  add_pn_unitary(circ, u, n, false);
+  // Add CU to {0, n}
+  Eigen::Matrix2cd m = nth_root(u, 1 << (n - 1));
```
cqc-alec, line 876: Think about handling of `n == 0`?

cqc-alec (changes requested):

tket/src/Transformations/Replacement.cpp
```diff
@@ -40,10 +41,12 @@ Circuit TK2_circ_from_multiq(const Op_ptr op) {
       return c;
     }
     case OpType::CnX:
-      if (n_qubits >= 6 && n_qubits <= 8) {
+      if (n_qubits >= 6 && n_qubits <= 50) {
         // TODO We should be able to do better than this.
```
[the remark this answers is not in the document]
yao-cqc: What did you have in mind for the TODO?

tket/tests/test_ControlDecomp.cpp
```diff
@@ -84,24 +87,106 @@ static bool check_incrementer_borrow_1_qubit(const unsigned n) {
...
+  for (unsigned i = 0; i < correct_out_bits.size(); i++) {
+    if (correct_out_bits[i] == 1) {
+      sv_set_idx = sv_set_idx + (1 << (n - i - 1));
+    }
+  }
+  const StateVector sv = tket_sim::get_statevector(circ);
+  bool correct = true;
+  for (unsigned i = 0; i < sv.size(); ++i) {
+    if (i == sv_set_idx) {
+      correct &= (std::abs(sv[i]) - 1 < ERR_EPS);
```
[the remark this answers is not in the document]
yao-cqc: Ahh, yes, you are right.

tket/src/Transformations/Replacement.cpp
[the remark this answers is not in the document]
cqc-alec: Ah I remember now, this is part of `TK2_circ_from_multiq` and the TODO was about more efficient decompositions to TK2 without going via CX. So it should probably stay.

tket/tests/test_ControlDecomp.cpp
```diff
@@ -397,6 +515,45 @@ SCENARIO("Test a CnX is decomposed correctly when bootstrapped", "[.long]") {
   }
 }
 
+SCENARIO("Test a CnX is decomposed correctly using the linear depth method") {
```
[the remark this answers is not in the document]
yao-cqc: You are right, it takes 2.5s on a M1 chip. I'll mark it as "long". The test below (checking controlled random unitaries) only takes 0.4s so I'll leave it as it is.

tket/src/Transformations/ControlledGates.cpp
yao-cqc: There isn't any specific reason. Is there any advantage in using `X`?

cqc-alec: Just a general preference for unparametrized gates over parametrized ones. There may still be some passes that handle them better; and it makes the circuit more "readable" (at least to me). But I don't feel particularly strongly about this.

cqc-alec (approved):
I leave the decision about Rx/X up to you.

cqc-alec (approved):
Thanks!

tket/src/Transformations/include/Transformations/ControlledGates.hpp
```diff
@@ -30,6 +30,27 @@ Circuit incrementer_borrow_1_qubit(unsigned n);
...
 Circuit incrementer_borrow_n_qubits(unsigned n);
 
+/**
+ * @brief Get an n-qubit incrementer circuit with linear depth and O(n^2) gate
+ * count. There exists a global phase difference
+ * https://arxiv.org/abs/2203.11882
+ *
+ * @param n number of qubits
+ * @param lsb set to false if we don't want to toggle the least siginificant bit
+ * @return Circuit CRx, Rx(1)
```
cqc-alec, line 40: Oh,  this needs changing, sorry.

tket/src/Transformations/include/Transformations/ControlledGates.hpp
```diff
@@ -30,6 +30,27 @@ Circuit incrementer_borrow_1_qubit(unsigned n);
...
+ */
+Circuit incrementer_linear_depth(unsigned n, bool lsb = true);
+
+/**
+ * @brief Implement CnU gate with linear depth and O(n^2) gate count.
+ * https://arxiv.org/abs/2203.11882
+ *
+ * @param n number of controls
+ * @param u the controlled 2x2 unitary matrix
+ * @return Circuit CRx, Rx(1), and controlled unitary boxes
```
cqc-alec, line 50: And this.

cqc-alec (changes requested):

tket/src/Transformations/include/Transformations/ControlledGates.hpp
yao-cqc: Thanks for spotting this. This one actually won't produce Rx(1) or X. It's updated now.

cqc-alec (approved):

PR #413, opened 2022-06-22
Remove "long" tag from tktokenswap tests.

cqc-alec:
Otherwise they won't run. They actually add very little to the running time of the tests.


ferbetanzo:
> Otherwise they won't run. They actually add very little to the running time of the tests.

Sorry, what do you mean with "they won'r run"? That we want to run them more frequently than weekly (as the other long running tests)?

cqc-alec:
> > Otherwise they won't run. They actually add very little to the running time of the tests.
> 
> Sorry, what do you mean with "they won'r run"? That we want to run them more frequently than weekly (as the other long running tests)?

This relates to https://github.com/CQCL/tket/pull/406 . These tests have been moved from the main test suite into a library-specific test suite. They are always run on the CI without specifying "long", so these ones won't be run at all if they have this tag. I don't think it's (yet) worth separating out long-running tests for the libraries, which will not not usually be run by PRs.

ferbetanzo (approved):
LGTM.

PR #421, opened 2022-06-27
Make WSM component into a versioned library

cqc-alec:
This is the obvious next candidate for modularization: it's big and self-contained with its own tests.
The coverage test failure is expected, since a chunk of code and corresponding tests has been moved.


libs/tkwsm/test/SolvingProblems/test_SnakeIntoSquareGrid.cpp
```diff
@@ -201,15 +201,13 @@ SCENARIO("embedding paths into square grids - quicker problems, fewer tests") {
...
-    "embedding paths into square grids - quicker problems, more tests",
-    "[.long]") {
+SCENARIO("embedding paths into square grids - quicker problems, more tests") {
   // We test only quicker problems, but MORE of them;
   // so this is a "long" test.
   test(true, 1048, 100, 1);
 }
 
-SCENARIO("embedding paths into square grids - all slower problems", "[.long]") {
+SCENARIO("embedding paths into square grids - all slower problems") {
```
cqc-melf, line 210: I assume that this change is because we don't do long runs for the libs?

cqc-alec: Yes.

cqc-melf (approved):
From my side this can be merged now, happy to ignore the coverage issue as for the others.

PR #425, opened 2022-06-28
Better exceptions

### LINKED ISSUE #422: Improve exception naming and use
TKET defines a few custom exception types in `Utils/Exceptions.hpp` which are not always used according to their documentation. In particular, `NotImplemented` is not only used for operation types, and `NotValid` is not always used for operations.

These should be renamed to make their meaning clearer, and other exception types used when they are not appropriate.

(This header file should probably be eliminated and exceptions defined in the components where they make sense.)

cqc-alec:
Use either standard exceptions or custom exceptions with descriptive names, instead of the `NotImplemented` and `NotValid` exceptions that were used in all sorts of contexts.

Closes #422 .


sjdilkes (dismissed):

cqc-alec:
Thanks @sjdilkes ; sorry I just had to merge again from `develop` to resolve a conflict in `MetaOp.cpp`.

sjdilkes (approved):

PR #428, opened 2022-06-30
[test] Use --decisions flag for gcovr while generating tket coverage report

ferbetanzo:
The latest release of [gcovr](https://github.com/gcovr/gcovr/releases/tag/5.2) fixes the issue observed when using the `--decisions` flag in the tket code base. Also, it seems the line coverage is now correctly calculated. This PR then enables the use of the `--decisions` flag with `gcovr` for creating the coverage report, and the check done in `compare-coverage` against the branches score reported.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -30,6 +30,7 @@ jobs:
         filters: |
           tket:
             - 'tket/**'
+            - '.github/workflows/coverage.yml'
 
   generate_coverage:
     name: Generate coverage report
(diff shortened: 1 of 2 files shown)


.github/workflows/coverage.yml
```diff
@@ -44,6 +45,7 @@ jobs:
     - uses: actions/checkout@v3
       with:
         fetch-depth: '0'
+        persist-credentials: false
```
cqc-alec, line 48: What is the intent of this change?

ferbetanzo: Oh, I added that because I was fetching the `gcovr` repo as the fix was not yet released. It's not needed anymore, so I'll remove it.

cqc-alec (approved):

cqc-alec:
This looks great, can we do the same in `test_libs.xml` for the libraries (perhaps in a separate PR)?

ferbetanzo:
> This looks great, can we do the same in `test_libs.yml` for the libraries (perhaps in a separate PR)?

Sure, will take care of it :)

PR #430, opened 2022-07-02
Use symengine from conan-center

cqc-alec:
And use default gmp for multi-precision arithmetic instead of boost.
Simplify a few cmake files by linking to all conan libraries instead of explicit lists.


sjdilkes (approved):
Nice!

PR #431, opened 2022-07-04
Add missing quote to readme conan command

yao-cqc:


yao-cqc:
> Is this required on mac? Works on Linux like before?

Yes, it doesn't work on my Mac without the quotes. I have tested that it works on Linux with or without the quotes. They use quotes in their official examples for `conan search` command. https://docs.conan.io/en/1.46/reference/commands/consumer/search.html

cqc-melf:
> > Is this required on mac? Works on Linux like before?
> 
> Yes, it doesn't work on my Mac without the quotes. I have tested that it works on Linux with or without the quotes. They use quotes in their official examples for `conan search` command. https://docs.conan.io/en/1.46/reference/commands/consumer/search.html

Great!

cqc-melf (approved):

PR #435, opened 2022-07-07
Catch tket exceptions

Roland-djee:
This PR adds functionality to catch `tket` exceptions in a more meaningful way in `pytket`.

Currently, exceptions in `tket` are `std::exception` that `pybind11` translates into `RuntimeError`. 
There is a larger spectrum of errors defined in individual modules. For instance [here](https://github.com/CQCL/tket/blob/df1b966350e151d31b2f66a4503db0dade75379a/tket/src/Circuit/include/Circuit/Circuit.hpp#L188) in the `Circuit` module.

### DIFF
--- a/pytket/binders/circuit/main.cpp
+++ b/pytket/binders/circuit/main.cpp
@@ -19,6 +19,7 @@
 #include <sstream>
 
 #include "Circuit/Command.hpp"
+#include "Circuit/Circuit.hpp"
 #include "Gate/Gate.hpp"
 #include "Gate/OpPtrFunctions.hpp"
 #include "Gate/SymTable.hpp"
@@ -44,6 +45,12 @@ void init_library(py::module &m);
 
 PYBIND11_MODULE(circuit, m) {
   init_unitid(m);
+  py::register_exception<CircuitInequality>(m, "CircuitInequality");
+  py::register_exception<CircuitInvalidity>(m, "CircuitInvalidity");
+  py::register_exception<Unsupported>(m, "Unsupported");
+  py::register_exception<MissingVertex>(m, "MissingVertex");
+  py::register_exception<MissingEdge>(m, "MissingEdge");
+  py::register_exception<SimpleOnly>(m, "SimpleOnly");
   py::class_<Op, std::shared_ptr<Op>>(
       m, "Op", "Encapsulates operation information")
       .def_static(
--- a/pytket/tests/circuit_test.py
+++ b/pytket/tests/circuit_test.py
@@ -19,6 +19,12 @@ import pickle
 
 from pytket.circuit import (  # type: ignore
     Circuit,
+    CircuitInequality,
+    CircuitInvalidity,
+    Unsupported,
+    MissingVertex,
+    MissingEdge,
+    SimpleOnly,
     Op,
     OpType,
     Command,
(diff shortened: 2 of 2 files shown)


pytket/binders/circuit/Circuit/add_op.cpp
```diff
@@ -511,7 +511,11 @@ void init_circuit_add_op(py::class_<Circuit, std::shared_ptr<Circuit>> &c) {
             } else {
               ancilla_ = Qubit(ancilla.value());
             }
-            circ->add_assertion(box, qubits_, ancilla_, name);
+            try {
+              circ->add_assertion(box, qubits_, ancilla_, name);
+            } catch (const std::exception &e) {
+              throw CircuitInvalidity(e.what());
```
cqc-alec, line 517: It should not be necessary to catch and rethrow exceptions manually like this! The point of registering the exception with pybind11 was that this should happen automatically. Does it not work?

In particular, we don't want to translate _all_ exceptions to `CircuitInvalidity` here: only those that are actually `CircuitInvalidity` exceptions.

cqc-alec (changes requested):

pytket/binders/circuit/Circuit/add_op.cpp
Roland-djee: No. It doesn't work automatically and more worryingly, it can depend on the compiler. For instance [9ad9a1c](https://github.com/CQCL/tket/pull/435/commits/9ad9a1c2889483f8e08caff04c47571411bf1cc3) where the `RuntimeError` is actually caught correctly by compilers on Linux (gcc I believe ?) and WIndows (MSVC ?) whereas Clang seems to let it pass. Moreover, seemingly identical tests are registered differently using Clang. For example, in [that test](https://github.com/CQCL/tket/blob/b7f131e68aa289eeb0e96089c788c91ac1eb3a1a/pytket/tests/circuit_test.py#L865), the first context manager picks up the `RuntimeError` as a `CircuitInvalidity` but not the second. Overall, this looks like bugs in Clang when compiling binders. I know, the solution is inelegant, but that's one that works for all compilers.

cqc-alec: OK, thanks for investigating this. I don't think we should merge it like this as it will be a nightmare to maintain and will mistranslate a lot of exceptions. I think we should close this for now, but it would be nice to produce a minimal example that illustrates what goes wrong, and raise an issue on the relevant repo (pybind11? clang?).

Roland-djee: Argh, it took me 3 days to go to the bottom of this.

Anyway, I fully understand this and moreover, it will be difficult to extend if we subtype exceptions. 

About an example, I have the feeling this is partially due to the inherent layers of abstraction and possible successive catches and rethrows of exceptions in tket/pytket. But if an example from tket/pytket is fine, I can see and try to extract something.

Roland-djee: The other thing about this is that it _has_ to be caught as `std::exception`. Trying to catch it as a subtype of `std::logic_error` won't work.

Roland-djee: One possible explanation, while working with Will on catching the `ZXError` is that the `typeid(ZXError).hash_code()` in the binder is different from the ones in the core tket library. And to be fair, we had no clue how to resolve this.

cqc-alec:
@Roland-djee Can we close this PR?

PR #441, opened 2022-07-13
Extend reg comparison

Roland-djee:
This PR addresses the extension of the previous `ConstPredicate` to a more general `RegPredicate` where the operands can be either logic expressions, bits, bitregisters or constants.

### DIFF
--- a/pytket/pytket/circuit/add_condition.py
+++ b/pytket/pytket/circuit/add_condition.py
@@ -46,7 +46,7 @@ def _add_condition(
     elif isinstance(condition, Predicate):
         pred_exp, pred_val = condition.args
         # Predicate constructor should ensure arg order
-        if isinstance(pred_val, Constant):
+        if not isinstance(pred_val, Constant):
             raise NotImplementedError
     elif isinstance(condition, BitLogicExp):
         pred_val = 1
--- a/pytket/pytket/circuit/add_condition.py
+++ b/pytket/pytket/circuit/add_condition.py
@@ -52,7 +52,8 @@ def _add_condition(
         # Predicate constructor should ensure arg order
         if not isinstance(pred_val, Constant):
             raise NonConstError(
-                "Condition expressions must be of type `Predicate` with a constant second operand."
+                "Condition expressions must be of type `Predicate`\
+                with a constant second operand."
             )
     elif isinstance(condition, BitLogicExp):
         pred_val = 1
--- a/pytket/pytket/circuit/add_condition.py
+++ b/pytket/pytket/circuit/add_condition.py
@@ -52,8 +52,7 @@ def _add_condition(
         pred_exp = condition
     else:
         raise ValueError(
-            f"Condition {condition} must be of type Bit, "
-            "BitLogicExp or RegPredicate"
+            f"Condition {condition} must be of type Bit, " "BitLogicExp or RegPredicate"
         )
 
     next_index = (
--- a/pytket/pytket/circuit/add_condition.py
+++ b/pytket/pytket/circuit/add_condition.py
@@ -35,6 +35,10 @@ from pytket.circuit.logic_exp import (
 )
 
 
+class NonConstError(Exception):
+    """A custom exception class for non constant predicate argument."""
+
+
 def _add_condition(
     circ: Circuit, condition: Union[Predicate, Bit, BitLogicExp]
 ) -> Tuple[Bit, bool]:
@@ -47,7 +51,9 @@ def _add_condition(
         pred_exp, pred_val = condition.args
         # Predicate constructor should ensure arg order
         if not isinstance(pred_val, Constant):
-            raise NotImplementedError
+            raise NonConstError(
+                "Condition expressions must be of type `Predicate` with a constant second operand."
+            )
     elif isinstance(condition, BitLogicExp):
         pred_val = 1
         pred_exp = condition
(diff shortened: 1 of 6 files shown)


pytket/pytket/circuit/logic_exp.py
```diff
@@ -350,22 +350,26 @@ class RegRsh(BinaryOp, RegLogicExp):
     op = RegWiseOp.RSH
 
 
-class ConstPredicate(BinaryOp):
-    """A binary predicate where at least one of the arguments is constant."""
+class RegPredicate(BinaryOp):
```
ss2165, line 353: It can be bit comparisons too so better to call just `Predicate`

```suggestion
class Predicate(BinaryOp):
```

pytket/pytket/circuit/logic_exp.py
```diff
@@ -350,22 +350,26 @@ class RegRsh(BinaryOp, RegLogicExp):
...
-        assert isinstance(exp, (LogicExp, Bit, BitRegister))
-        assert isinstance(const, Constant)
+    def __init__(
+        self,
+        lhs: Union[LogicExp, Variable, Constant],
+        rhs: Union[LogicExp, Variable, Constant],
+    ) -> None:
+        super().__init__(lhs, rhs)
+        assert isinstance(lhs, (LogicExp, Bit, BitRegister, Constant))
+        assert isinstance(rhs, (LogicExp, Bit, BitRegister, Constant))
```
ss2165, line 363: these assertions are now unecessary and can be removed

ss2165: In fact the whole init function can therefore be removed

pytket/pytket/circuit/add_condition.py
```diff
@@ -36,24 +36,23 @@
...
     """Add a condition expression to a circuit using classical expression boxes,
     rangepredicates and conditionals. Return predicate bit and value of said bit.
     """
     if isinstance(condition, Bit):
         return condition, True
-    elif isinstance(condition, ConstPredicate):
+    elif isinstance(condition, RegPredicate):
         pred_exp, pred_val = condition.args
-        # ConstPredicate constructor should ensure arg order
+        # RegPredicate constructor should ensure arg order
```
ss2165, line 48: this comment is no longer true, you will need to add an exception here if the `pred_val` is not a constant (or handle it by adding the expression directly)

pytket/pytket/circuit/logic_exp.py
```diff
@@ -350,22 +350,26 @@ class RegRsh(BinaryOp, RegLogicExp):
     op = RegWiseOp.RSH
 
 
-class ConstPredicate(BinaryOp):
-    """A binary predicate where at least one of the arguments is constant."""
+class RegPredicate(BinaryOp):
+    """A binary predicate where the arguments are either registers or constants."""
```
ss2165, line 354:
```suggestion
    """A binary predicate where the arguments are either Bits, BitRegisters, or constants."""
```

ss2165 (changes requested):

pytket/pytket/circuit/logic_exp.py
Roland-djee: OK, I just pass it ?

ss2165: the whole function definition can be removed, it will call the super init automatically

Roland-djee: Yes, sure but what I meant is that:
```
class Predicate(BinaryOp):
"""A binary predicate where the arguments are either Bits, BitRegisters, or constants."""
 ...
```

Roland-djee: Oh really ? OK didn't know about that. Thanks.

Roland-djee: Done.

pytket/pytket/circuit/add_condition.py
Roland-djee: hm, shouldn't we simply remove the assertion ? Or, do you mean that further down th e`bool(pred_val )` will fail ?

ss2165: yes exactly, the rest of this function only applies if the value is a bool, so we either need to raise an exception or handle it, for now I think raising an exception is the right thing to do (instead of the assert).

Roland-djee: Oh, I see. I was confused because now, `pred_val` can be something else than a `Constant` i.e. a register and I didn't see how you could get a bool from it. But what you mean is that this `add_condition` method must work identically to what it was defined for: with the second argument being a `Constant` despite the extension of types for the predicate.

Roland-djee: Done.

ss2165: `NotImplemented` is not very descriptive, ideally use a custom exception like `NonConstError` with a helpful message like "Condition expressions must be Predicates which have a constant as their second operand".

ss2165 (changes requested):

pytket/pytket/circuit/add_condition.py
Roland-djee: Sure thing. Done.

ss2165 (approved):
LGTM

Roland-djee:
> LGTM

Thanks.

Roland-djee:
> LGTM



> LGTM

Hi @ss2165 could you please approve it ? It won't let me merge it otherwise. Thanks.

ss2165 (approved):

Roland-djee:
Thanks.

PR #442, opened 2022-07-15
[feature] Use implicit swaps in DecomposeTK2

lmondada:


tket/src/Transformations/Decomposition.cpp
```diff
@@ -563,6 +565,45 @@ static void best_exact_decomposition(
   }
 }
 
+// Given a 2-qb circuit with a single TK2, returns a tuple (pre, angles, post)
+// where pre/post are the parts of the circuit that come before/after the TK2
+// gate and angles are the 3 angles of the TK2 gate
+static auto get_TK2_angles(const Circuit &circ) {
```
cqc-alec, line 571: Since this function is essentially "undoing" what `CircPool::TK2_using_normalised_TK2()` did, it could perhaps be a bit more efficient if we refactored so that this function just took three angles, and was called both from this file and from the `CircPool` function. But I don't know if it is worth it.

tket/src/Predicates/PassGenerators.cpp
```diff
@@ -592,7 +596,7 @@ PassPtr DecomposeTK2(const Transforms::TwoQbFidelities& fid) {
 }
 
 PassPtr ThreeQubitSquash(bool allow_swaps) {
-  Transform t = Transforms::two_qubit_squash() >>
+  Transform t = Transforms::two_qubit_squash(allow_swaps) >>
```
cqc-alec, line 599: Does this change mean that the maximum number of CX gates in a 2-qubit circuit is less than 20 now, or are there still cases when it is 20?

cqc-alec (changes requested):
Thanks!

tket/src/Transformations/Decomposition.cpp
lmondada: That was a very good idea. It made the code nicer.

tket/src/Predicates/PassGenerators.cpp
lmondada: No. `two_qubit_squash` will still require up to three CX in the worst case.

cqc-alec (changes requested):
Nice.

cqc-alec:
Oops, now has conflicts in the changelog again -- and needs clang-formatting. Otherwise good to go.

PR #444, opened 2022-07-21
Feature/improve q control box decomp

yao-cqc:
This PR includes the following changes:

1. All the controlled decomposition methods are moved from ``Transformations`` to ``CircPool``
2. Generalise lemma 7.9 to decompose any multi-controlled SU(2) gates. This method is more efficient than ``CnU_linear_depth_decomp`` when n>9
3. Generalise the CnX Gray code decomposition method for decomposing any multi-controlled unitaries.
4. New handling for decomposing non-symbolic QControlBox.


tket/src/Circuit/CircUtils.cpp
```diff
@@ -705,6 +696,325 @@ Circuit with_controls(const Circuit &c, unsigned n_controls) {
...
+        replacement = CircPool::CnU_linear_depth_decomp(total_controls, m);
+      } else {
+        // Compute the SU(2) angles from the TK1 angles
+        std::vector<double> angles = tk1_angles_from_unitary(m);
+        if (equiv_val(angles[3], 1., 2)) {
+          // if the phase is odd, it can be absorbed into the first Rz rotation
+          angles[0] = angles[0] + 2;
+        } else if (!equiv_0(angles[3], 2)) {
+          // because it's SU(2), the phase must be integers
+          throw std::logic_error("tk1 angles don't imply SU(2)");
```
cqc-alec, line 984: Could this be a `TKET_ASSERT`?

tket/src/Circuit/CircUtils.cpp
```diff
@@ -705,6 +696,325 @@ Circuit with_controls(const Circuit &c, unsigned n_controls) {
...
+        }
+        // convert tk1 angles to zyz angles
+        std::vector<double> zyz_angles = {
+            angles[0] - 0.5, angles[1], angles[2] + 0.5};
+        replacement = CircPool::CnSU2_linear_decomp(
+            total_controls, zyz_angles[0], zyz_angles[1], zyz_angles[2]);
+      }
+    } else {
+      if (total_controls == 3 || total_controls == 4) {
```
cqc-alec, line 993: Add comment explaining this choice?

cqc-alec (changes requested):
Thank you, looking good!

tket/src/Circuit/CircUtils.cpp
```diff
@@ -717,6 +710,331 @@ Circuit with_controls(const Circuit &c, unsigned n_controls) {
...
+    unit_vector_t args = command.get_args();
+    TKET_ASSERT(!args.empty());
+    for (unsigned i = 0; i < args.size() - 1; i++) {
+      control_qubits.insert(args[i].index()[0]);
+    }
+    target_qubit = args.back().index()[0];
+    is_symmetric =
+        (op->get_type() == OpType::CZ || op->get_type() == OpType::CU1);
+    color = as_gate_ptr(op)->commuting_basis(args.size() - 1);
+    TKET_ASSERT(color != Pauli::I);
```
cqc-alec, line 772: I'm not sure we can actually assert this, because `commuting_basis()` can return `Pauli::I`. Should we throw an exception in this case, or should we handle it?

cqc-alec (commented):
Thanks! Just one thing I'm not sure about.

tket/src/Circuit/CircUtils.cpp
yao-cqc: As far as I know only ``BRIDGE`` and ``noop`` have ``Pauli::I`` ports. Since ``CnGateBlock`` is only being used in ``with_controls_numerical``,  bridges should have been rebased into ``CXs`` and ``noop `` should have been ignored. But I think an exception is more suitable here.

cqc-alec (approved):

PR #447, opened 2022-07-24
Implement `CustomPass` accepting a user-supplied circuit transformation function

cqc-alec:
…on function.


ss2165 (commented):
1. If we exposed circuit mutation methods, would it be possible to have the python function mutate the circuit directly?
2. Why not have the inner function requirement be that it returns `(Citcuit, bool)` and use the second argument in the return of the transform?

cqc-alec:
> 1. If we exposed circuit mutation methods, would it be possible to have the python function mutate the circuit directly?

Yes. That would be a lot of extra work though, whereas this allows passes to be defined with the API we have now.

>     2. Why not have the inner function requirement be that it returns `(Citcuit, bool)` and use the second argument in the return of the transform?

@yao-cqc made the same suggestion and I think it is a good one. Alternatively we could have the C++ method check if the transformed circuit is equal to the original and return `true` or `false` accordingly, which would be more reliable (for a small runtime cost).

ss2165:
1. Agreed
2. I like that, we could have both and most users would just want the automatically checked one?

cqc-alec:
>     2. I like that, we could have both and most users would just want the automatically checked one?

I think I prefer to keep it simple for now and just have the automatically-checked one.
Will also add an optional `label` parameter for more useful serialization.

tket/src/Predicates/PassGenerators.cpp
```diff
@@ -751,4 +752,20 @@ PassPtr GlobalisePhasedX(bool squash) {
   return std::make_shared<StandardPass>(precons, t, postcon, j);
 }
 
+PassPtr CustomPass(
+    std::function<Circuit(const Circuit&)> transform,
+    const std::string& label) {
+  Transform t{[transform](Circuit& circ) {
+    Circuit circ_in = circ;
+    circ = transform(circ);
+    return circ != circ_in;
```
ss2165, line 761: not sure what the performance implications are but this looks more natural to me, it might avoid a copy?
```suggestion
    Circuit circ_out = transform(circ);
    bool success circ != circ_out;
    circ = circ_out;
    return success
```

ss2165 (approved):
LGTM

yao-cqc (approved):

PR #448, opened 2022-07-25
Add additional `RebaseCustom` constructor that takes a TK2 conversion function

cqc-alec:


cqc-alec:
This will be more useful once we have a TK2 option for `FullPeepholeOptimise` (which I will work on next).

pytket/pytket/passes/auto_rebase.py
```diff
@@ -85,21 +107,29 @@ def auto_rebase_pass(gateset: Set[OpType]) -> RebaseCustom:
...
-    return RebaseCustom(
-        gateset, get_cx_decomposition(gateset), get_TK1_decomposition_function(gateset)
-    )
+    tk1 = get_TK1_decomposition_function(gateset)
+    try:
+        return RebaseCustom(gateset, tk1, get_tk2_decomposition(gateset))
+    except NoAutoRebase:
+        pass
+    try:
+        return RebaseCustom(gateset, get_cx_decomposition(gateset), tk1)
```
lmondada, line 128: The order of the parameters here is different. Is this a bug or an inconsistent API?

pytket/binders/passes.cpp
```diff
@@ -543,6 +544,28 @@ PYBIND11_MODULE(passes, m) {
...
+      "to each TK1(a,b,c)."
+      "\n\n"
+      ":param gateset: the allowed operations in the rebased circuit\n"
+      ":param tk1_replacement: a function which, given the parameters (a,b,c) "
+      "of an Rz(a)Rx(b)Rz(c) triple, returns an equivalent circuit in the "
+      "desired basis\n"
+      ":param tk2_replacement: a function which, given the parameters (a,b,c) "
+      "of an XXPhase(a)YYPhase(b)ZZPhase(c) triple, returns an equivalent "
+      "circuit in the desired basis\n"
+      ":return: a pass that rebases to the given gate set");
```
lmondada, line 567:
```suggestion
      ":param gateset: The allowed operations in the rebased circuit.\n"
      ":param tk1_replacement: A function which, given the parameters (a,b,c) "
      "of an Rz(a)Rx(b)Rz(c) triple, returns an equivalent circuit in the "
      "desired basis.\n"
      ":param tk2_replacement: A function which, given the parameters (a,b,c) "
      "of an XXPhase(a)YYPhase(b)ZZPhase(c) triple, returns an equivalent "
      "circuit in the desired basis.\n"
      ":return: A pass that rebases to the given gate set.");
```

tket/src/Circuit/CircPool.cpp
```diff
@@ -910,6 +914,23 @@ Circuit TK2_using_CX(const Expr &alpha, const Expr &beta, const Expr &gamma) {
   }
 }
 
+Circuit TK2_using_CX_optimal(
```
lmondada, line 917: I would suggest renaming this function to `TK2_using_CX` and the other `TK2_using_CX` to `normalised_TK2_using_CX`. This way we can keep consistency between the names here and in `pytket.circuit.library`.

lmondada (changes requested):

pytket/binders/passes.cpp
```diff
@@ -543,6 +544,28 @@ PYBIND11_MODULE(passes, m) {
...
+      "RebaseCustom", &gen_rebase_pass_via_tk2,
+      "Construct a custom rebase pass, given user-defined rebases for TK1 and "
+      "TK2. This pass:"
+      "\n\n"
+      "1. decomposes multi-qubit gates not in the set of gate types `gateset` "
+      "to TK2 gates;\n"
+      "2. if TK2 is not in `gateset`, replaces TK2(a,b,c) gates via the "
+      "`tk2_replacement` function;\n"
+      "3. converts any single-qubit gates not in the gate type set to TK1;\n"
+      "4. if TK2 is not in `gateset`. applies the `tk1_replacement` function "
```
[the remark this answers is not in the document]
cqc-alec: The docs formatter doesn't recognise the (1) style (just checked on the public docs page). I will change the one above to use this style.

pytket/pytket/passes/auto_rebase.py
cqc-alec: It's an inconsistent API. I thought it more natural to have TK1 before TK2 but perhaps it is better to keep consistency and have the TK1 replacement function last.

pytket/binders/passes.cpp
cqc-alec: We are inconsistent in this regard. I think I prefer uncapitalized with no full point, when it is a description and not a sentence

tket/src/Circuit/CircPool.cpp
cqc-alec: Good idea.

pytket/binders/passes.cpp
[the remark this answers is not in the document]
cqc-alec: (Or perhaps the reason was the lack of double `\n` in the docstring above. Anyway, will fix it.)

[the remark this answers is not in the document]
cqc-alec: Done.

pytket/pytket/passes/auto_rebase.py
cqc-alec: Swapped/

tket/src/Circuit/CircPool.cpp
cqc-alec: Done.

pytket/pytket/passes/auto_rebase.py
lmondada: I agree with you, but I think the change is worth it for the sake of consistency.

pytket/binders/passes.cpp
lmondada: Fine by me :)

Could you use this style for the other overload of `RebaseCustom`, so that we are at least consistent within the function?

lmondada (changes requested):

lmondada (approved):
Thanks!

PR #449, opened 2022-07-25
Fix/json metaop data

cqc-melf:
This PR tries to add the data option for the meta op, not sure if this is the right place


ss2165 (changes requested):
"data" should be added as a property of "operation", see "n_qb" as an example here: https://github.com/CQCL/tket/blob/4df582ce1c1995abbd401f0911679cd205737670/schemas/circuit_v1.json#L278

cqc-melf:
Thank you, done!

ss2165 (approved):

PR #454, opened 2022-07-26
Update from `main`

cqc-alec:


cqc-alec:
The pytket test coverage failure looks like a false alarm due to test randomization: clearly this does not change pytket test coverage in any way! I propose ignoring it.

cqc-melf (approved):

PR #457, opened 2022-07-26
Switch to quantinuum conan repo

cqc-alec:
Requires a small change to your conan configuration to build, as described in the README.

I have uploaded new versions of all the libraries in `libs` to the new repo, increasing the version numbers so that there can be no ambiguity if both repos are registered as remotes (though I'd recommend removing the old remote from config).

### DIFF
--- a/.github/workflows/build_libs.yml
+++ b/.github/workflows/build_libs.yml
@@ -79,11 +79,11 @@ jobs:
       if: matrix.os == 'ubuntu-20.04'
       run: conan profile update settings.compiler.libcxx=libstdc++11 tket
     - name: add remote
-      run: conan remote add tket-conan https://tket.jfrog.io/artifactory/api/conan/tket-conan
+      run: conan remote add tket-libs https://quantinuumsw.jfrog.io/artifactory/api/conan/tket1-libs
     - name: build ${{ matrix.lib }}
       run: conan create --profile=tket -s build_type=${{ matrix.build_type }} -o ${{ matrix.lib }}:shared=${{ matrix.shared }} libs/${{ matrix.lib }} tket/stable --build=missing
     - name: authenticate to repository
-      run: conan user -p ${{ secrets.JFROG_ARTIFACTORY_TOKEN_1 }} -r tket-conan ${{ secrets.JFROG_ARTIFACTORY_USER_1 }}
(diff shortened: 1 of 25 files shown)


.github/workflows/build_libs.yml
```diff
@@ -160,7 +160,7 @@ jobs:
         cat <<EOF > env-vars
         TKLIB=${{ matrix.lib }}
         UPLOAD_PACKAGE=${UPLOAD_PACKAGE}
-        JFROG_ARTIFACTORY_TOKEN_1=${{ secrets.JFROG_ARTIFACTORY_TOKEN_1 }}
-        JFROG_ARTIFACTORY_USER_1=${{ secrets.JFROG_ARTIFACTORY_USER_1 }}
+        JFROG_ARTIFACTORY_TOKEN_2=${{ secrets.JFROG_ARTIFACTORY_TOKEN_2 }}
+        JFROG_ARTIFACTORY_USER_2=${{ secrets.JFROG_ARTIFACTORY_USER_2 }}
```
cqc-melf, line 164: Why are we not using the secrets directly?

libs/tkwsm/test/SolvingProblems/test_RandomGraphs.cpp
```diff
@@ -376,11 +376,11 @@ SCENARIO(
     entry = -1;
   }
   TestParameters params;
-  params.timeout = 30000;
+  params.timeout = 60000;
```
cqc-melf, line 379: Why is this changed? Is the interaction longer?

.github/workflows/build_libs.yml
cqc-alec: Because they are not directly accessible from within the `linuxbuildlib` script: they need to be passed as environment variables.

cqc-melf (commented):
Only a few questions.

libs/tkwsm/test/SolvingProblems/test_RandomGraphs.cpp
cqc-alec: I found I needed to increase it because the coverage test was failing with a timeout here. (It was only slightly over 30000 but I changed it to 60000 to be on the safe side.)

cqc-alec: https://github.com/CQCL/tket/runs/7524375317

cqc-melf: Perfect, thank you!

cqc-melf (approved):

PR #467, opened 2022-08-02
[libs/tkwsm] WSM improvements including initial placement

cqc-alec:
This PR brings in the code from Zen's development branch (`feature/draft_initial_qubit_placement_with_wsm`).

The main sequence of commits (up to "Added comments to readme.txt (about the dynamic bitset refactor)") are directly from that branch, rebased to `develop`, with include paths and cmake scripts corrected.

Build fixes (allowing it to build on all supported platforms) follow. These include some changes to integer types to avoid potential loss of data when converting them, removal of some unused parameters and variables, and a few other things.

Then there are a couple of fixes to bugs that led to differing results on different platforms.

I then had to "fix" a small number of tests, whose asserted results differed from the actual results. I am not happy having so many tests whose correctness is difficult to verify (and is also dependent on the internal workings of the RNG), but this is what we have.

Finally, because the line coverage had decreased compared with the previous version, I removed a few unused debugging functions and shortened some untested error strings (without essentially changing the logic), in order to pass the coverage test.

Merging this PR will just create a new 0.2.0 version of tkwsm, but it won't update tket's dependency to that version. I have checked locally that all tket tests pass with the new version. Updating tket's version dependency will follow in a second PR.

Note that I have not made any effort to understand the code, beyond what was necessary to get it building and passing tests.


sjdilkes (approved):
I've removed a few unused bits. I've also run some basic benchmarks to confirm there is no major divergence in behaviour.

PR #470, opened 2022-08-04
Prevent Architecture from having an edge between the same node

sjdilkes:
Currently it's possible to have an Architecture with an edge between a node to itself.
This is not something methods are designed to deal with or that we expect a realistic quantum device to need to represent.
This PR adds checks when edges are added to a Architecture and throws an error if the initialisation wants to do this.

Additionally`qubit_lines` has been updated to return an empty list if the circuit has no edges in its interaction graph.

### DIFF
--- a/tket/src/Placement/Qubit_Placement.cpp
+++ b/tket/src/Placement/Qubit_Placement.cpp
@@ -166,6 +166,9 @@ QubitGraph generate_interaction_graph(
 
 QubitLineList qubit_lines(const Circuit& circ) {
   const QubitGraph q_graph = generate_interaction_graph(circ);
+  if (q_graph.get_all_edges_vec().size() == 0) {
+    return {};
+  }
   std::set<Qubit> all_qb;
   for (const Qubit& qb : circ.all_qubits()) {
     all_qb.insert(qb);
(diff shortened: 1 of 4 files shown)


tket/src/Placement/Qubit_Placement.cpp
```diff
@@ -166,6 +166,9 @@ QubitGraph generate_interaction_graph(
 
 QubitLineList qubit_lines(const Circuit& circ) {
   const QubitGraph q_graph = generate_interaction_graph(circ);
+  if (q_graph.get_all_edges_vec().size() == 0) {
```
cqc-alec, line 169:
```suggestion
  if (q_graph.get_all_edges_vec().empty()) {
```

cqc-alec (changes requested):
Can you please amend the docstring for the `DirectedGraph` class (further down in this file) to say: "DirectedGraph instances are loop-free directed graphs".

tket/src/Placement/Qubit_Placement.cpp
sjdilkes: Done

cqc-alec (approved):

PR #474, opened 2022-08-05
[feature] Smarter `commute_through_multis`

lmondada:
With this PR `commute_through_multis` supports conditionals, commuting through TK1 and others when possible (in edge cases) and looks nicer.

### DIFF
--- a/tket/src/Circuit/CircUtils.cpp
+++ b/tket/src/Circuit/CircUtils.cpp
@@ -707,6 +707,7 @@ static Circuit with_controls_symbolic(const Circuit &c, unsigned n_controls) {
     c2.append(cnu1);
   }
 
+  remove_noops(c2);
   return c2;
 }
 
@@ -1171,6 +1172,33 @@ std::tuple<Circuit, std::array<Expr, 3>, Circuit> normalise_TK2_angles(
   post = post.dagger();
 
   return {pre, {a, b, c}, post};
+}
+
+void remove_noops(Circuit &circ) {
+  auto all_zeros = [](const std::vector<Expr> &exprs) {
+    for (const Expr &e : exprs) {
+      if (!equiv_0(e, 4)) {
+        return false;
+      }
+    }
+    return true;
+  };
+
+  std::set<OpType> supported_ops{
+      OpType::U1, OpType::U2, OpType::U3, OpType::TK1, OpType::TK2};
+  VertexSet bin;
+  BGL_FORALL_VERTICES(v, circ.dag, DAG) {
+    Op_ptr op = circ.get_Op_ptr_from_Vertex(v);
+    if (supported_ops.contains(op->get_type())) {
+      if (all_zeros(op->get_params())) {
+        circ.remove_vertex(
+            v, Circuit::GraphRewiring::Yes, Circuit::VertexDeletion::No);
+        bin.insert(v);
+      }
+    }
+  }
+  circ.remove_vertices(
+      bin, Circuit::GraphRewiring::No, Circuit::VertexDeletion::Yes);
 }
 
 }  // namespace tket
--- a/tket/src/Circuit/CircUtils.cpp
+++ b/tket/src/Circuit/CircUtils.cpp
@@ -707,7 +707,7 @@ static Circuit with_controls_symbolic(const Circuit &c, unsigned n_controls) {
     c2.append(cnu1);
   }
 
-  remove_noops(c2);
+  c2.remove_noops();
   return c2;
 }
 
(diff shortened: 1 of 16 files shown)


tket/src/Circuit/macro_circ_info.cpp
```diff
@@ -143,9 +143,7 @@ Circuit Circuit::subcircuit(const Subcircuit& sc) const {
       port_t in_port = get_source_port(e);
       OpType type = sub.get_OpType_from_Vertex(sub_source);
       if (is_initial_q_type(type) || type == OpType::ClInput) {
-        // For Quantum and Classical edges, boundary_edge == *it;
-        // for Boolean, this gives the corresponding Classical
-        Edge boundary_edge = this->get_nth_out_edge(source, in_port);
+        Edge boundary_edge = get_linear_edge(e);
```
sjdilkes, line 146: What's wrong with the previous solution?

tket/tests/test_PhaseGadget.cpp
```diff
@@ -132,7 +132,7 @@ SCENARIO("Full optimise_via_PhaseGadget") {
     const auto s0 = tket_sim::get_unitary(circ);
     Transforms::optimise_via_PhaseGadget(CXConfigType::Tree).apply(circ);
     REQUIRE(circ.count_gates(OpType::CX) == 12);
-    REQUIRE(circ.depth() == 13);
+    REQUIRE(circ.depth() == 14);
```
sjdilkes, line 135: What's causing the extra depth?

tket/src/Circuit/setters_and_getters.cpp
```diff
@@ -425,6 +425,15 @@ std::vector<std::optional<Edge>> Circuit::get_linear_out_edges(
   return outedges;
 }
 
+Edge Circuit::get_linear_edge(const Edge &e) const {
```
sjdilkes, line 428: Please can you add a basic test for this, including the Boolean case

sjdilkes (changes requested):
Code implementation looks fab, just some requests for additional tests.

tket/src/Circuit/macro_circ_info.cpp
lmondada: Nothing, does the same thing. But I needed the same code elsewhere so I factored it out (and removed the comment because it was redundant).

tket/src/Circuit/CircUtils.cpp
```diff
@@ -1173,4 +1174,31 @@ std::tuple<Circuit, std::array<Expr, 3>, Circuit> normalise_TK2_angles(
   return {pre, {a, b, c}, post};
 }
 
+void remove_noops(Circuit &circ) {
```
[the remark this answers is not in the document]
lmondada: Done!

tket/src/Circuit/setters_and_getters.cpp
lmondada: Sure thing!

tket/tests/test_PhaseGadget.cpp
lmondada: This is because a TK1 gate gets commuted through a CX, whereas before it wasn't. In this particular case this impedes a previous merger of two TK1 gates. I don't think it's a big deal, just bad luck.

cqc-alec: Can we change the test from `== 14` to `<= 14`, so that future improvements don't fail it?

tket/src/Circuit/include/Circuit/CircUtils.hpp
```diff
@@ -197,4 +197,9 @@ Circuit with_controls(const Circuit& c, unsigned n_controls = 1);
 std::tuple<Circuit, std::array<Expr, 3>, Circuit> normalise_TK2_angles(
     Expr a, Expr b, Expr c);
 
+/**
+ * @brief Remove standard gates with 0 angle
+ */
+void remove_noops(Circuit& circ);
```
cqc-alec, line 203: Any reason not to make this a `Circuit` method?

tket/src/Circuit/CircUtils.cpp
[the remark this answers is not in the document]
lmondada: Yes, you're right! Wasn't aware of that, thank you.

tket/src/Circuit/include/Circuit/CircUtils.hpp
lmondada: Done.

tket/src/Transformations/PQPSquash.cpp
```diff
@@ -70,8 +70,8 @@ std::pair<Circuit, Gate_ptr> PQPSquasher::flush(
   OpType p = p_, q = q_;
 
   if (smart_squash_ && commutation_colour.has_value()) {
-    Gate P(p_, {0}, 1);
-    Gate Q(q_, {0}, 1);
+    Gate P(p_, {0.123}, 1);
+    Gate Q(q_, {0.123}, 1);
```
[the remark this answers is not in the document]
lmondada: The point is to have a non-zero angle, because it is otherwise detected as an identity. I am adding a comment in the code.

tket/src/Transformations/BasicOptimisation.cpp
```diff
@@ -206,33 +217,35 @@ static bool commute_singles_to_front(Circuit &circ) {
...
+          EdgeVec rewire_edges;
+          op_signature_t edge_types;
+          Edge rewire_edge = circ.get_last_edge(current_v, current_e);
+          for (const Edge &e : circ.get_in_edges(prev_v)) {
+            EdgeType type = circ.get_edgetype(e);
+            Edge boundary_edge;
+            if (type == EdgeType::Quantum) {
+              boundary_edge = rewire_edge;
+            } else {
+              // There should not be any Classical edges for conditional ops
```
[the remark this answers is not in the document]
lmondada: I thought so, but I might not be familiar with how every operation is implemented. Are there conceivably ops that are gate types and have one quantum wire as well as classical (i.e. non-boolean) input wires?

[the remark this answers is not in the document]
cqc-alec: Yes, `Measure` is a gate type according to our definition.

cqc-alec (commented):
Just one query, concerning a possible issue with conditional Measure gates in `commute_singles_to_front()`. I don't know if this is a real issue; perhaps a test could be added?

cqc-alec (approved):

sjdilkes (approved):

PR #476, opened 2022-08-11
[bugfix] FullPeephole in TK2 mode squashes correctly

lmondada:
An extra `two_qubit_squash` pass is necessary after `clifford_simp` when targetting TK2.

Also, implicit swaps are never required when targetting TK2 because any such swap can be swallowed into an existing TK2. Implicit swaps would be added in further down the compilation when TK2s are further decomposed into hardware primitives.


yao-cqc (approved):
Looks good!

PR #479, opened 2022-08-17
[feature] Better squash to tk1

lmondada:


tket/tests/test_Synthesis.cpp
```diff
@@ -2186,5 +2186,124 @@ SCENARIO("Restricting ZZPhase gate angles.") {
...
+    circ.add_op<unsigned>(OpType::T, {1});
+    circ.add_op<unsigned>(OpType::Sdg, {1});
+    circ.add_op<unsigned>(OpType::Rz, 2.434, {0});
+    circ.add_op<unsigned>(OpType::CZ, {0, 1});
+    circ.add_op<unsigned>(OpType::Rx, 22.22, {1});
+    circ.add_op<unsigned>(OpType::Rx, 3.334, {0});
+  }
+
+  auto u0 = tket_sim::get_unitary(circ);
+  Transforms::decompose_ZXZ_to_TK1().apply(circ);
```
yao-cqc, line 2243: Maybe require this to return true and also test that the rebased circuits only have tk1 and 2-q gates? Similar for the ZYZ test below.

yao-cqc (changes requested):
Can you add a PR comment explaining why this improves tk1 squash?

tket/tests/test_Synthesis.cpp
lmondada: Done.

lmondada:
I've added a comment in the code about why this rebase sequence makes sense. Apologies for all the trashy commits.

yao-cqc (approved):

PR #480, opened 2022-08-18
[bugfix] Valid TK2 rejected by `in_weyl_chamber`

lmondada:


yao-cqc (commented):
Can you add an entry to changelog? This looks good. Thanks

yao-cqc (approved):

PR #481, opened 2022-08-20
Circuit HTML - Remove blank lines in template.

aidanCQ:
When exporting jupyter notebooks as markdown, Jupyter (or pandoc presumably) preserves blank lines in the html it exports which isn't valid in a markdown file. I could try and write a fix on the jupyter repo or write a custom html parser but this is the easiest solution for now. This will help when embedding the circuit renderer on webpages when using a markdown export.

### DIFF
--- a/pytket/pytket/circuit/display/circuit.html
+++ b/pytket/pytket/circuit/display/circuit.html
@@ -5,7 +5,6 @@
 <html lang="en">
 <head>
     <meta charset="UTF-8">
-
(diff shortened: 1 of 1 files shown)


cqc-melf:
I think this is a good way to solve this. Are there any other changes required for the same display functions in the interaction with mushroom? @DNA386 Have you seen this?

DNA386 (approved):
Seems fine to me!

aidanCQ:
@cqc-melf Are we good to merge this?

cqc-melf:
Yes, I have merged it

PR #482, opened 2022-08-22
Feature/commute rzs through phased x

yao-cqc:


lmondada (changes requested):
Cool, thanks. Could you add `SquashRzPhasedX` to `auto_squash_pass`, so that whenever we’re squashing PhasedX and Rz we use this?

lmondada:
> Cool, thanks. Could you add `SquashRzPhasedX` to `auto_squash_pass`, so that whenever we’re squashing PhasedX and Rz we use this?

Could you also add this pass to `auto_squash_pass`?

pytket/pytket/passes/auto_rebase.py
```diff
@@ -141,4 +141,7 @@ def auto_squash_pass(gateset: Set[OpType]) -> SquashCustom:
     :return: Squash to target gateset
     :rtype: SquashCustom
     """
+    if gateset == {OpType.Rz, OpType.PhasedX}:
+        return SquashRzPhasedX()
```
lmondada, line 145: One more thing: can you change this test to check that {Rz, PhasedX} is included in gateset (instead of strictly equal)?

lmondada (changes requested):
Sorry for the late reply. Just one tiny thing and then we can merge it in.

lmondada (approved):

PR #483, opened 2022-08-22
Release/1.5.1

cqc-alec:


pytket/tests/transform_test.py
```diff
@@ -1120,7 +1120,7 @@ def test_tk2_decompositions() -> None:
         Path(__file__).resolve().parent / "qasm_test_files" / "test19.qasm"
     )
     FullPeepholeOptimise().apply(c)
-    assert c.depth() <= 29
+    assert c.depth() <= 30
```
cqc-melf, line 1123: This looks like we are allowing a result that is worse then before?

cqc-melf (commented):
This contains all the changes from develop?

pytket/tests/transform_test.py
yao-cqc: My local test suggests that the CX count has been reduced although we have an increase in depth. We probably don't have to worry about this for now.

cqc-melf (approved):

PR #487, opened 2022-08-23
Feature/tket upload.1

cqc-alec:


.github/workflows/build_and_test.yml
```diff
@@ -17,10 +17,54 @@ concurrency:
...
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-python@v3
+      with:
+        python-version: '3.10'
+    - uses: dorny/paths-filter@v2
+      id: filter
+      with:
+        base: ${{ github.ref }}
+        filters: |
```
cqc-melf, line 39: What about changes in `https://github.com/CQCL/tket/tree/develop/schemas`

.github/workflows/build_and_test.yml
```diff
@@ -17,10 +17,54 @@ concurrency:
...
+          tket:
+            - '{tket/src/**,recipes/tket/conanfile.py}'
+          tket_tests:
+            - '{tket/tests/**,recipes/tket-tests/conanfile.py}'
+          tket_proptests:
+            - '{tket/proptests/**,recipes/tket-proptests/conanfile.py}'
+          pytket:
+            - 'pytket/**'
```
cqc-melf, line 47: What about changes in `https://github.com/CQCL/tket/tree/develop/recipes/pybind11` ?

.github/workflows/build_and_test.yml
```diff
@@ -199,10 +255,13 @@ jobs:
...
-      run: conan create --profile=tket recipes/tket --build=missing
+      if: needs.check_changes.outputs.tket_changed == 'true'
+      run: conan create --profile=tket recipes/tket tket/stable --build=missing
     - name: Build and run tket tests
+      if: needs.check_changes.outputs.tket_changed == 'true' || needs.check_changes.outputs.tket_tests_changed == 'true'
       run: conan create --profile=tket recipes/tket-tests
     - name: Build and run tket proptests
+      if: needs.check_changes.outputs.tket_changed == 'true' || needs.check_changes.outputs.tket_proptests_changed == 'true'
       run: |
         conan install --profile=tket rapidcheck/cci.20210702@ --build=missing
```
cqc-melf, line 266: What is the flag `rapidcheck/cci.20210702@` doing here?

.github/workflows/build_and_test.yml
```diff
@@ -350,36 +418,24 @@ jobs:
         cd pytket/tests
         pip install -r requirements.txt
         pytest --ignore=simulator/ --doctest-modules
+    - name: Upload package
+      if: github.event_name == 'push' && github.event.push.ref == 'refs/heads/develop' && needs.check_changes.outputs.tket_changed == 'true'
+      run: |
+        conan user -p ${{ secrets.JFROG_ARTIFACTORY_TOKEN_2 }} -r tket-libs ${{ secrets.JFROG_ARTIFACTORY_USER_2 }}
+        conan upload tket/${{ needs.check_changes.outputs.tket_ver }}@tket/stable --all -r=tket-libs
+        conan user -c
```
cqc-melf, line 426: Is this `-c` a logout?

.github/workflows/build_and_test.yml
```diff
@@ -448,6 +514,11 @@ jobs:
         cd tests
         pip install -r requirements.txt
         pytest --ignore=simulator/ --doctest-modules
+    - name: Upload package
+      if: github.event_name == 'push' && github.event.push.ref == 'refs/heads/develop' && needs.check_changes.outputs.tket_changed == 'true'
+      run: |
+        conan user -p ${{ secrets.JFROG_ARTIFACTORY_TOKEN_2 }} -r tket-libs ${{ secrets.JFROG_ARTIFACTORY_USER_2 }}
+        conan upload tket/${{ needs.check_changes.outputs.tket_ver }}@tket/stable --all -r=tket-libs
```
cqc-melf, line 521: This is missing the logout ( `-c` ) ?

CONTRIBUTING.md
```diff
@@ -61,3 +61,11 @@ mypy --config-file=mypy.ini -p pytket -p tests
 We use [pylint](https://pypi.org/project/pylint/) on the CI to check compliance
 with a set of style requirements (listed in `pytket/.pylintrc`). You should run
 `pylint` over any changed files before submitting a PR, to catch any issues.
+
+## Version numbers
+
+If you make any changes in `thet/src`, you should bump the version number of
```
cqc-melf, line 67: Is there a way to make this automated on the CI?

cqc-melf: Are there defined changes that required which changes in the "part" of the version?

cqc-melf (changes requested):

.github/workflows/build_and_test.yml
cqc-alec: Those will have no impact on the tket conan package,

.github/workflows/build_and_test.yml
cqc-alec: I ended up not using `pytket_changed` at al. Even so, changes to the pybind11 recipe won't affect pytket unless the pytket version dependency is updated (which would be picked up by this check).

.github/workflows/build_and_test.yml
cqc-alec: This can probably be omitted if we put `--build=missing` in the following line (but I would do that in a separate PR).

.github/workflows/build_and_test.yml
cqc-alec: Yes.

.github/workflows/build_and_test.yml
cqc-alec: It's only necessary on the self-hosted M1 machine (I didn't want the login to persist there, but the cloud VMs just get torn down anyway).

CONTRIBUTING.md
cqc-alec: > Is there a way to make this automated on the CI?

Could be done with some scriptery, but might be more trouble than it's worth. Autogenerated conanfiles scare me a little.

cqc-alec: > Are there defined changes that required which changes in the "part" of the version?

Good question. I think it would be good to follow semver here. I'll add a note abut this to the README.

.github/workflows/build_and_test.yml
cqc-melf: Perfect!

.github/workflows/build_and_test.yml
cqc-melf: Okay, lets keep that in mind!

CONTRIBUTING.md
cqc-melf: > > Is there a way to make this automated on the CI?
> 
> Could be done with some scriptery, but might be more trouble than it's worth. Autogenerated conanfiles scare me a little.

Okay, yes I can understand that

cqc-melf: Are there any bad side effects expected if the version is not updated in a first CI run before a review?

cqc-melf:
Do we need any additional information about this in the readme for local development? It would be possible to use this locally if only working on python for example?

cqc-melf (commented):
see above

CONTRIBUTING.md
cqc-alec: > Are there any bad side effects expected if the version is not updated in a first CI run before a review?

I hope not. Packages only get uploaded to the artifactory on push to develop.

cqc-alec:
> Do we need any additional information about this in the readme for local development? It would be possible to use this locally if only working on python for example?

I will add a note to the README about this.

cqc-melf (approved):

PR #492, opened 2022-08-25
[bugfix] Prefer ZZPhase over ZZMax in ties

lmondada:
Currently `pytket-quantinuum` is still not using any `ZZPhase`. This is because the fidelities of the gates are unknown, and are thus set to 1. for both `ZZMax` and `ZZPhase`. As a consequence, TKET never chooses `ZZPhase` over `ZZMax`.

I suggest to solve this by choosing `ZZPhase` over `ZZMax` as a tie breaker when fidelities are equal.


cqc-alec (commented):
The change makes sense to me. Looks like some tests need updating.

lmondada:
I've changed the condition for using ZZPhase to: fidelity is equal AND it uses fewer gates. This sounds like a good default behaviour.

cqc-alec (approved):

PR #501, opened 2022-08-30
[feature] add opaque barrier gates to qasm conversion

cqc-melf:
draft for adding the opaque gate handling in the qasm parsing.


pytket/pytket/qasm/qasm.py
```diff
@@ -335,18 +335,84 @@ def mixedcall(self, tree: List) -> Iterator[CommandDict]:
         except StopIteration:
             args = next_tree
             pars = []
-        params = [f"({par})/pi" for par in pars]
+
+        params = [
+            f"({par})/pi" for par in pars
+        ]  # do we want to do this for all parameters?
+
+        treat_as_barrier = [
```
cqc-melf, line 343: I am not happy with this solution to have the list of the gates in here which we want to have converted to a barrier at the end. The problem is that the gates are all marked as opaque like others in the dev header file which we want to have handled differently, so I have not found a solution which is not having a hard coded list in here. I think the better idea would be to have pytket special gates for sleep, group and order like for the other opaque gates, but that would have been a bigger changes which we maybe should discuss in a bigger group?

Even with not being happy with this, I would suggest adding the for now as suggested in this PR.

pytket/pytket/qasm/qasm.py
```diff
@@ -335,18 +335,84 @@ def mixedcall(self, tree: List) -> Iterator[CommandDict]:
...
+                first_elem = True
+                for par in params:
+                    if not first_elem:
+                        param_sorted += ","
+                    first_elem = False
+                    param_sorted += par
+
+                op["data"] = f"{opstr}({param_sorted})"
+
+                args = list(args)
```
cqc-alec, line 402: Is this line necessary?

pytket/pytket/qasm/qasm.py
```diff
@@ -335,18 +335,84 @@ def mixedcall(self, tree: List) -> Iterator[CommandDict]:
...
-                "type": "CustomGate",
-                "id": str(uuid.uuid4()),
-                "gate": gdef,
-            }
-            box["params"] = params
-            op["box"] = box
-            params = []  # to stop duplication in to op
+            op: Dict[str, Any] = {}
+            if opstr in treat_as_barrier:
+                params = [p[1:-4] for p in params]
```
cqc-alec, line 390: Can you explain this please?

cqc-alec (changes requested):

pytket/pytket/qasm/qasm.py
```diff
@@ -335,18 +335,84 @@ def mixedcall(self, tree: List) -> Iterator[CommandDict]:
...
+                op["type"] = "Barrier"
+                param_sorted = ""
+                first_elem = True
+                for par in params:
+                    if not first_elem:
+                        param_sorted += ","
+                    first_elem = False
+                    param_sorted += par
```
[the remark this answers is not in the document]
cqc-melf: Yes

pytket/pytket/qasm/qasm.py
cqc-melf: Yes, it removes the `(` and `)/pi` added in https://github.com/CQCL/tket/blob/feature/opaque-barrier-gates-qasm/pytket/pytket/qasm/qasm.py#L340

The original `pars` variable is empty after L340. Could also solve this with removing `(` and `)/pi` from the `params` variable and adding a second variable somewhere including this again if you prefer this?

pytket/pytket/qasm/qasm.py
cqc-alec: But the line below would work with any iterator. Is it used anywhere else?

pytket/pytket/qasm/qasm.py
cqc-alec: How about just not adding the "(" and ")/pi" when opstr is in treat_as_barrier?

cqc-alec (changes requested):

pytket/pytket/qasm/qasm.py
cqc-melf: good point, done.

pytket/pytket/qasm/qasm.py
cqc-alec: Then can we do the conversion once unconditionally, before line 388, instead of twice conditionally (at lines 396 and 423)? Or are there some cases where we don't want to do the conversion? As it stands, at line 426 I don't know if `args` is a list or not, and this makes the code harder to understand.

cqc-alec (commented):
Look good, one small query.

cqc-alec:
There is this comment in `circuit_to_qasm_io()`:
```
                # TODO when opaque gates are supported, make sure they are not
                # unrolled here
```
Does this still apply? (See also the error message below it.)

pytket/pytket/qasm/qasm.py
cqc-melf: Yes, moved that out of the if block

cqc-alec (approved):

PR #505, opened 2022-09-06
Feature/add cn z and cn y

yao-cqc:


cqc-alec (changes requested):
Looks good, just a small comment.

cqc-alec (approved):

PR #507, opened 2022-09-06
Feature/gcc11

cqc-alec:


cqc-melf (approved):
Looks good!

Do we have plans to update this for the extensions, too?

cqc-alec:
> Do we have plans to update this for the extensions, too?

Yes, I intend to make this change on the extensions too. (Should be more straightforward as they are just Python.)

cqc-melf:
> > Do we have plans to update this for the extensions, too?
> 
> Yes, I intend to make this change on the extensions too. (Should be more straightforward as they are just Python.)

If you need help, let me know. I have scripts for generating branches and PRs on multiples repos in one step.

PR #510, opened 2022-09-07
Bugfix/measure ops should throw error in tksim

### LINKED ISSUE #490: `Circuit.get_unitary()` and `Circuit.get_statevector()` ignore measure operations in the circuit
I think they should throw an error in this case.

yao-cqc:
resolves #490

### DIFF
--- a/pytket/tests/backend_test.py
+++ b/pytket/tests/backend_test.py
@@ -69,7 +69,7 @@ def test_resulthandle() -> None:
 
 
 def test_check_handle_single() -> None:
-    b = TketSimBackend()
+    b = TketSimBackend(ignore_measures=True)
     c = Circuit(2).measure_all()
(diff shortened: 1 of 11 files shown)


pytket/tests/backend_test.py
```diff
@@ -69,7 +69,7 @@ def input_str_invalid(st: str) -> bool:
```
cqc-alec, line 69: Can we add a test that the simulator does throw an error as expected when measurements are included?

yao-cqc: I also implemented a simple Catch2 exception matcher to work with the `REQUIRE_THROWS_MATCHES` macro. I think that macro is potentially useful for many cases. However, I'm not sure if we already have a better way to do this.

cqc-alec (approved):

PR #512, opened 2022-09-09
Upload tket conan packages

cqc-alec:
This is a second attempt at #487 , which was reverted because on Linux pytket was not working when the tket package was downloaded from the repo insetad of being built locally. The reason was that the shared libraries contained hard-coded rpath information pointing to paths that were correct when they were built but are no longer correct at runtime.

The solution is to:

- move the tket libraries alongside the binder libraries in Linux builds of pytket, as we were already doing for MacOS and Windows;
- explicitly set the rpath to `${ORIGIN}` on Linux, which ensures that libraries look in the local directory for other libraries that they need to load.

### DIFF
--- a/tket/src/CMakeLists.txt
+++ b/tket/src/CMakeLists.txt
@@ -55,6 +55,9 @@ ELSEIF(APPLE)
     # set correct install_name
     set(CMAKE_INSTALL_NAME_DIR "@loader_path")
     set(CMAKE_BUILD_WITH_INSTALL_NAME_DIR ON)
+ELSEIF(UNIX)
+    set(CMAKE_BUILD_WITH_INSTALL_RPATH TRUE)
+    set(CMAKE_INSTALL_RPATH "\${ORIGIN}")
 ENDIF()
 
 # if you add new modules here make sure that it is added at the right position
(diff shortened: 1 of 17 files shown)


tket/src/CMakeLists.txt
```diff
@@ -55,6 +55,9 @@ ELSEIF(APPLE)
     # set correct install_name
     set(CMAKE_INSTALL_NAME_DIR "@loader_path")
     set(CMAKE_BUILD_WITH_INSTALL_NAME_DIR ON)
+ELSEIF(UNIX)
+    set(CMAKE_BUILD_WITH_INSTALL_RPATH TRUE)
+    set(CMAKE_INSTALL_RPATH "\${ORIGIN}")
```
cqc-melf, line 60: This is not required for the test package? or the other tket modules?

cqc-melf (commented):
have you done a test run on the Ci somewhere?

cqc-alec:
> have you done a test run on the Ci somewhere?

I have done several, including builds that actually [uploaded the packages](https://github.com/CQCL/tket/actions/runs/3021444191), others that [used them without building tket](https://github.com/CQCL/tket/actions/runs/3021858539), and a [test of the release workflow](https://github.com/CQCL/tket/actions/runs/3022663680).

tket/src/CMakeLists.txt
cqc-alec: Not required for the test package because nothing else uses it. This is only (currently) necessary for libraries that are packaged with pytket.

cqc-melf:
Perfect, thank you for sharing the runs.

cqc-melf (approved):

PR #520, opened 2022-09-12
Improvements to workflow and README

cqc-alec:
Forcing build of tket when `--build=missing` is used on MacOS prevents use of cached version. I think this could make a difference in cases where PRs get updated.


cqc-melf (approved):
Good point

PR #523, opened 2022-09-12
Fix bash.

cqc-alec:


cqc-alec:
@cqc-melf Sorry about the back and forth; I think this should be working now.

cqc-alec:
I think the MacOS M1 build failure is unrelated to this commit; it is also occurring on [this run](https://github.com/CQCL/tket/runs/8305962273?check_suite_focus=true) for example. Possibly related to the warning emitted in the pytket build.

cqc-melf (approved):
No problem.

From my side you can merge this without having to wait for fixing the issue in the test.

PR #537, opened 2022-09-15
Update boost version in libs.

cqc-alec:
I don't know why the coverage check failed; this shouldn't happen as there are no changes to the code being analysed; possibly the last report may have been uploaded in error, or gcovr has changed somehow. Anyway, I think this failure can be ignored.

PR updating tket to follow.

### DIFF
--- a/libs/tktokenswap/conanfile.py
+++ b/libs/tktokenswap/conanfile.py
@@ -18,7 +18,7 @@ from conans.errors import ConanInvalidConfiguration
 
 class TktokenswapConan(ConanFile):
     name = "tktokenswap"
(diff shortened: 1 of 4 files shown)


cqc-melf (approved):
Feel free to merge it without the coverage check passing

cqc-melf:
have you set the coverage check as not required or is that not required for all PRs? I think it should be required?

cqc-alec:
> have you set the coverage check as not required or is that not required for all PRs? I think it should be required?

I'd forgotten it wasn't required. Yes I agree it should be, will fix that.

PR #539, opened 2022-09-16
Fix mypy warning.

cqc-alec:
I was getting this warning locally, though not on the CI for some reason:
```
error: Returning Any from function declared to return "ndarray[Any, Any]"
```


cqc-alec:
```
error: Redundant cast to "ndarray[Any, Any]"
```

It seems I can't win.

PR #541, opened 2022-09-16
Handle python versions in matrix.

cqc-alec:
As discussed [here](https://github.com/CQCL/tket/pull/529#discussion_r970694550).

The price we pay is storing the contents of `~/.conan/data/tket` in an artefact, which is then downloaded by each pytket build. But this is still a net gain in time, as well as a useful refactor.

I also factored out the script to set up the conan profile and options.

(Will require a small tweak if #540 is merged first.)

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -61,8 +61,8 @@ jobs:
       if: steps.filter.outputs.tket == 'true' && steps.test_package_exists.outputs.tket_package_exists == 'true'
       run: exit 1
 
-  build_test:
-    name: Build and test
+  build_test_tket:
+    name: Build and test (tket)
     needs: check_changes
     strategy:
       matrix:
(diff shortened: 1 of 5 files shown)


cqc-alec:
Will modify "required" checks list after merge.

.github/workflows/build_and_test.yml
```diff
@@ -153,91 +148,79 @@ jobs:
...
+    - name: Test pytket with coverage
+      if: matrix.os == 'ubuntu-22.04' && matrix.pyver == '3.8'
       run: |
         cd pytket/tests
         pip install -r requirements.txt
         pytest --ignore=simulator/ --doctest-modules --cov=../pytket --cov-branch --cov-report=html --cov-report=xml:htmlcov/cov.xml
-    - name: Test pytket (3.8)
-      if: matrix.os != 'ubuntu-22.04' && github.event_name == 'push'
+    - name: Test pytket
+      if: matrix.os != 'ubuntu-22.04' || matrix.pyver != '3.8'
```
cqc-melf, line 201: Why are we doing the exclusion?

cqc-melf: is this the case where the coverage check is done?

cqc-alec: Yes, the tests are already run in that case using the coverage flags, so no need to run them again.

cqc-melf (approved):

PR #544, opened 2022-09-17
Switch to macos-12

cqc-alec:
Because it seems apple-clang 13.1 (macos-12) is not quite ABI-compatible with apple-clang 13.0 (macos-11), we have to build our own symengine package for this to work. I copied the recipe exactly from the conan-center, and uploaded the packages.

### DIFF
--- a/.github/workflows/build-symengine.yml
+++ b/.github/workflows/build-symengine.yml
@@ -1,0 +1,62 @@
+name: Build symengine
+
+on:
+  push:
+    branches:
+      - 'symengine-upload-new/**'
+
+jobs:
+  build:
+    name: Build
+    strategy:
+      matrix:
+        os: ['ubuntu-22.04', 'macos-12', 'windows-2022']
+    runs-on: ${{ matrix.os }}
+    env:
+      CONAN_REVISIONS_ENABLED: 1
+    steps:
+    - uses: actions/checkout@v3
+    - name: Install conan
+      uses: turtlebrowser/get-conan@v1.1
+    - name: Set up conan profile
+      run: conan profile new tket --detect --force
+    - name: Fix up libcxx in profile
+      if: matrix.os == 'ubuntu-22.04'
+      run: conan profile update settings.compiler.libcxx=libstdc++11 tket
+    - name: add remote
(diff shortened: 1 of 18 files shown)


cqc-melf:
The macos 11 check is still required but not running? I assume that is intended?

cqc-alec:
> The macos 11 check is still required but not running? I assume that is intended?

Yes, I will change it after merging this.

recipes/symengine/all/conandata.yml
```diff
@@ -0,0 +1,7 @@
+sources:
+  "0.8.1":
+    url: "https://github.com/symengine/symengine/releases/download/v0.8.1/symengine-0.8.1.tar.gz"
+    sha256: "41eb6ae6901c09e53d7f61f0758f9201e81fc534bfeecd4b2bd4b4e6f6768693"
+  "0.9.0":
```
cqc-melf, line 5: Why are we using two different versions?

.github/workflows/build-symengine.yml
```diff
@@ -0,0 +1,62 @@
+name: Build symengine
+
+on:
+  push:
+    branches:
+      - 'symengine-upload-new/**'
```
cqc-melf, line 6: do you want to have this on other branches, too?

recipes/symengine/all/conandata.yml
cqc-alec: We are only using 0.9.0 (as that is the version specified in the tket conanfile). We could delete the other option from here, but it does no harm (and is copied directly from the conan-center recipe). When new versions come out, we can just add them here and update tket retaining the ability to revert if there are problems.

.github/workflows/build-symengine.yml
cqc-alec: No, the idea is that in order to upload a new symengine version (which won't happen very often), one just has to create a branch with a name matching that pattern, modify the symengine recipe and push it.

cqc-melf: Sounds like a good idea! Thank you for explaining this.

cqc-melf (dismissed):

cqc-alec:
Just merged from develop and bumped the tket version again.

cqc-melf (approved):

PR #545, opened 2022-09-20
Initial PR

ChloeSilva:
Still some stuff to work on (const correctness, ceres polluting output...) but the basic functionality seems to be working :smile:


tket/src/NumericalOptimiser/NumericalOptimiser.cpp
```diff
@@ -0,0 +1,108 @@
...
+#include "NumericalOptimiser.hpp"
+
+namespace tket {
+
+Circuit optimise(Circuit &circ, Architecture &arch, unsigned int k) {
+  PartitionVec post_synthesis;
+  PartitionVec pre_synthesis = partition(circ, arch, k);
+
+  for (Partition partition : pre_synthesis) {
+    post_synthesis.insert(post_synthesis.begin(), synthesise(partition, arch));
```
cqc-alec, line 24: The order is reversed?

tket/src/NumericalOptimiser/NumericalOptimiser.cpp
```diff
@@ -0,0 +1,108 @@
...
+  for (Partition partition : post_synthesis) {
+    EdgeVec edges = {};
+    for (Qubit qubit : partition.second) {
+      edges.push_back(circ.get_nth_out_edge(circ.get_in(qubit), 0));
+    }
+    Subcircuit to_replace = {edges, edges};
+    circ.substitute(partition.first, to_replace);
+  }
+
+  return circ;
```
cqc-alec, line 36: Since this acts in place on `circ` (by reference) is there any reason to return it?
(At some point we may want to make this a `Transform`.)

tket/src/NumericalOptimiser/NumericalOptimiser.cpp
```diff
@@ -0,0 +1,108 @@
...
+  PartitionVec post_synthesis;
+  PartitionVec pre_synthesis = partition(circ, arch, k);
+
+  for (Partition partition : pre_synthesis) {
+    post_synthesis.insert(post_synthesis.begin(), synthesise(partition, arch));
+  }
+
+  for (Partition partition : post_synthesis) {
+    EdgeVec edges = {};
+    for (Qubit qubit : partition.second) {
```
cqc-alec, line 29: I'm a bit confused about this. It seems that the `to_replace` subcircuit is empty (the input edges are the same as the output edges, no gates in between); so we are just prepending the replacements to the original circuit, without removing anything. Am I misunderstanding?

tket/src/NumericalOptimiser/include/NumericalOptimiser/PartitionCircuit.hpp
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * @brief Partitions a circuit into subcircuits of maximum size k.
+ * The qubits in each partition form a connected subgraph of the given
+ * architecture to maximise the number of positions where a CX gate may
+ * be placed during resynthesis.
+ * 
+ * @param circ the circuit to be partitioned
+ * @param arch the target architecture the circuit will be run on
+ * @param k the maximum number of qubits per partition
+ * @return an ordered vector of the partitions
```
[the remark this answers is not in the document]
ChloeSilva: The circuits are in causal order

tket/src/NumericalOptimiser/NumericalOptimiser.cpp
ChloeSilva: Partition() has the side effect of removing each partition so the circuit is empty when it comes time to add the re-synthesised sub circuits

PR #547, opened 2022-09-20
Fix phase issue in SquashRzPhasedX

### LINKED ISSUE #535: `SquashRzPhasedX` does not preserve global phase
For example:

```python
from pytket import Circuit
from pytket.passes import SquashRzPhasedX

c = Circuit(1).H(0).H(0)

SquashRzPhasedX().apply(c)

print(c.get_commands())
print(c.phase)
```

Output:
```
[]
1.0
```

yao-cqc:
Fixes #535


cqc-alec:
The build failure looks like my fault. I am investigating.

PR #550, opened 2022-09-21
Further workflow improvements

cqc-alec:
A final (for now) round of improvements:
* Don't cache conan data on MacOS (not necessary any more, and saves time).
* Remove unused environment variable.
* Remove unnecessary fetches of whole repo history.
* Install ccache using external action.
* Update actions versions and pin marketplace actions to specific versions.


cqc-melf:
You might have missed https://github.com/CQCL/tket/blob/feature/workflow.1/.github/workflows/release.yml#L309
Or is that the newest version?

cqc-alec:
> You might have missed https://github.com/CQCL/tket/blob/feature/workflow.1/.github/workflows/release.yml#L309 Or is that the newest version?

Thanks, I had missed it.

cqc-melf (approved):
Looks good,I think we should update the actions on the other repos, too.

PR #552, opened 2022-09-21
[infra] Use `gcovr --decisions` for tket libs

ferbetanzo:
This PR enables the use of the `--decisions` flag for `gcovr` when generating the coverage report of the tket libs. It also changes the `test_libs.yml` workflow a bit so that all the libs are built and tested when changes to the yaml file itself are made.

### DIFF
--- a/.github/workflows/test_libs.yml
+++ b/.github/workflows/test_libs.yml
@@ -143,7 +143,7 @@ jobs:
     - name: build coverage report
       run: |
         mkdir ${{ matrix.lib }}-coverage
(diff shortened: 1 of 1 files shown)


.github/workflows/test_libs.yml
```diff
@@ -31,14 +33,37 @@ jobs:
...
+        run: |
+          echo "::set-output name=matrix::$LIBS_TO_TEST"
   test_libraries:
     name: test library
-    needs: changes
-    if: ${{ needs.changes.outputs.libs != '[]' && needs.changes.outputs.libs != '' }}
+    needs: set_libs_matrix
+    if: ${{ needs.set_libs_matrix.outputs.libs != '[]' && needs.set_libs_matrix.outputs.libs != '' }}
     strategy:
+      fail-fast: false
```
cqc-alec, line 63: We want the workflow to fail if any of these jobs fails. Is this still the case with this set to `false`?

ferbetanzo: Yes, that's still the case. The `fail-fast: false` only means that any matrix jobs still running will not be canceled if another job fails. At the end, the workflow is reported as failed if any of those jobs failed.

cqc-alec (approved):

PR #555, opened 2022-09-28
Refactor Placement directory

sjdilkes:
In pytket 1.4 `GraphPlacement` was updated to use the new TKET-WSM library instead of the vf2 algorithm available through boost when finding subgraph monomorphisms. This decision was based on internal benchmarking that showed the algorithm to generally find solutions quicker, find subgraph monomorphisms within set timeouts the boost algorithm could not, and also perform better for QAOA style problems which originally motivated the construction of TKET-WSM. This generally went in hand with better routing performance.

However, later benchmarking has shown that for some problems the routing performance was worse and that the final physically mapped `Circuit` used many more `Architecture` `Node`. The reason for this was due to the edge breaking routine still in place. In some cases the pattern graph a subgraph monomorphism was found for (constructed by looking at interacting `Qubit` in a `Circuit`) would have multiple disconnected subgraph, when the original circuit was one connected graph. TKET-WSM would in some cases map each of these disconnected subgraph to distant parts of the `Architecture` graph, leaving routing to use additional `Architecture` `Node` to map program `Qubit`. It's suspected that this was also a problem with the previous boost algorithm, but that it "happened" to map the disconnected subgraph to close subgraphs on the target graph (probably due to indexing, or luck).

Primarily, this pull request updates the `GraphPlacement` algorithm to not do this. In the previous algorithm a pattern graph was constructed from the two-qubit interactions in the `Circuit` to be mapped and then searched for a subgraph monomorphism on the target graph (just the `Architecture` graph). If no subgraph monomorphism was found, a single edge was removed from the pattern graph and then it searched for a subgraph monomorphism on the same target graph. This was repeated until one was found, at which point it was returned.

In this updated algorithm it first searches for a subgraph monomorphism of the full pattern graph on the full target graph. If none is found, it removes edges from the pattern graph based on out-degree and tries again. If this does not work it adds distance 2 edges on the `Architecture` graph as lower weighted edges on the target graph and tries again. This is repeated until a solution is found, emulating a partial solution. The extra weighted edges prevent badly mapped cases as previously described from occurring. Additionally this rewrite has speed improvements over the previous algorithm and seems to give minor improvements to routing on _average_.
It proceeds like this as TKET-WSM tool finds solutions by first removing all impossible cases, meaning it quickly returns `false` if no subgraph monomorphism can be found.

Additionally this pull request refactors the files in src/Placement, updates the base `Placement` logic to do a naive mapping (though I appreciate this could be considered a controversial decision so am happy to renege) and minorly changes the `NoiseAwarePlacement` logic. It also rewrites a series of tests, improving on prior meagre test coverage.


cqc-alec:
Just going to close and reopen this to trigger a new build. There was a bug in the tket version bump check which meant it wasn't failing when it should (and I think this should have!).

cqc-alec (changes requested):
Please bump tket version, and corresponding required version in tket tests and proptests and pytket.
(Was not failing before because of a bug in the workflow script, now fixed.)

sjdilkes:
> Please bump tket version, and corresponding required version in tket tests and proptests and pytket. (Was not failing before because of a bug in the workflow script, now fixed.)

Done.

tket/tests/tkettestsfiles.cmake
```diff
@@ -59,7 +59,9 @@ set(TEST_SOURCES
     ${TKET_TESTS_DIR}/test_PauliGraph.cpp
     ${TKET_TESTS_DIR}/test_Architectures.cpp
     ${TKET_TESTS_DIR}/test_ArchitectureAwareSynthesis.cpp
-    ${TKET_TESTS_DIR}/Placement/test_Monomorphism_Edge_Break.cpp
```
cqc-alec, line 62: I assume these tests are no longer applicable?
We should delete the file if so.

tket/tests/Placement/test_Placement.cpp
```diff
@@ -19,723 +19,135 @@
 #include "Placement/Placement.hpp"
 
 namespace tket {
-namespace test_Placement {
```
cqc-alec, line 22: Have all the tests deleted from here moved to other files?

tket/tests/Circuit/test_Boxes.cpp
```diff
@@ -149,6 +157,38 @@ SCENARIO("Using Boxes", "[boxes]") {
     // check it's the identity
     REQUIRE((d1m - Eigen::Matrix4cd::Identity()).cwiseAbs().sum() < ERR_EPS);
   }
+  GIVEN("Unitary Box Identity constructors") {
```
cqc-alec, line 160: I don't see what the circuit equality tests here are testing. The API doesn't promise that an empty UnitaryBox will decompose to these specific circuits, and a change of implementation could change that and so make this test fail. I think it's only valid to check unitary equivalence of the decomposed circuits.

(I also don't see what it has to do with the changes in this PR?)

tket/tests/Circuit/test_Boxes.cpp
```diff
@@ -161,6 +201,14 @@ SCENARIO("Using Boxes", "[boxes]") {
...
   }
   GIVEN("ExpBox manipulation") {
+    // empty
+    Circuit empty(2);
+    empty.add_op<unsigned>(OpType::TK1, {0, 0, 0}, {0});
+    empty.add_op<unsigned>(OpType::TK1, {0, 0, 0}, {1});
+    empty.add_op<unsigned>(OpType::TK2, {0, 0, 0}, {0, 1});
+    empty.add_op<unsigned>(OpType::TK1, {0, 0, 0}, {0});
+    empty.add_op<unsigned>(OpType::TK1, {0, 0, 0}, {1});
+    REQUIRE(*(ExpBox().to_circuit()) == empty);
```
cqc-alec, line 211: Same here.

cqc-alec (changes requested):
Thanks, looks good. A few small comments.

pytket/docs/conf.py
```diff
@@ -40,7 +40,7 @@
 # The short X.Y version
 version = "1.7"
 # The full version, including alpha/beta/rc tags
-release = "1.7.2"
+release = "1.7.3"
```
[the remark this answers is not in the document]
sjdilkes: Yeah weird, I'm also not sure. Maybe it's because I haven't pulled in from develop for a while - have done so now incase it nudges github to not show the diff.

pytket/binders/placement.cpp
```diff
@@ -103,7 +65,7 @@ PYBIND11_MODULE(placement, m) {
             .def("__repr__",
                  [](const Placement &) { return "<tket::Placement>"; })
             .def("place",
-              [](const Placement &placement, Circuit &circ) {
+              [](Placement &placement, Circuit &circ) {
```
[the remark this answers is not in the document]
sjdilkes: The `Architecture` held in `Placement` is not guaranteed `const` because they can cache the distance between their `Node` as they are found for the first time. I'm not entirely sure how we got away with it being const before, but I suspect it involved refinding such distances. I'm happy to have a look at restoring the `const` nature though - probably by holding distance between `Node` in the `Placement` base class at point of construction that is then queried.

pytket/binders/placement.cpp
```diff
@@ -133,14 +95,14 @@ PYBIND11_MODULE(placement, m) {
                  "\n\n:param circuit: The circuit the maps are designed for."
                  "\n:return: list of dictionaries mapping " CLSOBJS(Qubit) " "
                  "to " CLSOBJS(Node),
-                 py::arg("circuit"))
+                 py::arg("circuit"), py::arg("matches"))
```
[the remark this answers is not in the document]
sjdilkes: Done. I've added "100" as the default though this is not particularly well reasoned, though I can't think of a good reason for any value: if you're calling this you probably want more than 1, so we could put an extremely large value in its place.

pytket/binders/placement.cpp
```diff
@@ -150,11 +112,16 @@ PYBIND11_MODULE(placement, m) {
...
+          py::init<Architecture &, unsigned, unsigned>(),
           "The constructor for a LinePlacement object. The Architecture "
           "object describes the connectivity "
-          "between qubits.\n\n:param arc: An Architecture object.",
-          py::arg("arc"))
+          "between qubits.\n\n:param arc: An Architecture object."
+          "\n:param maximum_line_gates: maximum number of gates used to "
+          "construct lines."
+          "\n:param maximum_line_depth: maximum depth gates are used to "
+          "construct lines.",
```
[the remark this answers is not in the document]
sjdilkes: Agreed - I've attempted to explain this in the constructor explanation:
```"The constructor for a LinePlacement object. The Architecture "
          "object describes the connectivity "
          "between qubits. In this class, a reduced qubit interaction"
          "subgraph is constructed where each node has maximum outdegree 2"
          "and does not construct a circle (i.e. lines)".
          "To place the Circuit, a Hamiltonian Path is found in the Architecture"
          "and this subgraph of lines is assigned along it."```
          
In addition to updating the parameter explanation as "lines" isn't clear. Please let me know if I should expand this more.

File: pytket/binders/placement.cpp (hunk 6 above)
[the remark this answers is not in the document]
Comment (cqc-alec) [reply pytket/binders/placement.cpp:68]: You could make the distance cache a `mutable` class member perhaps?

File: pytket/binders/placement.cpp [hunk 9]
```diff
@@ -195,40 +165,40 @@ PYBIND11_MODULE(placement, m) {
       "to find the best placement map.")
       .def(
           py::init<
-              Architecture &, avg_node_errors_t, avg_link_errors_t,
-              avg_readout_errors_t>(),
-          "The constructor for a NoiseAwarePlacement object. The Architecture "
-          "object describes the connectivity between qubits. "
+              Architecture &, unsigned, unsigned, unsigned, unsigned,
+              avg_node_errors_t, avg_link_errors_t, avg_readout_errors_t>(),
```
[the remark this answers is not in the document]
sjdilkes: Yeah that's a good point, I can switch them around.

pytket/tests/predicates_test.py
```diff
@@ -224,29 +234,22 @@ def test_routing_and_placement_pass() -> None:
     pl = Placement(arc)
     routing = RoutingPass(arc)
     placement = PlacementPass(pl)
-    nplacement = NaivePlacementPass(arc)
+    # nplacement = NaivePlacementPass(arc)
     cu = CompilationUnit(circ.copy())
     assert placement.apply(cu)
     assert routing.apply(cu)
-    assert nplacement.apply(cu)
```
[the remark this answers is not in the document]
sjdilkes: For full clarity (I mentioned this in the description of the PR), `Placement` used to do nothing when `place` was used, and we later added `NaivePlacement` to assign `Circuit` `Qubit` to any free `Architecture` `Node`. It seemed to me when I added `NaivePlacement` that really it was adding functionality that a generic `Placement` should do, i.e. just getting the `Qubit` anywhere on the `Architecture`.
`NaivePlacementPass` still exists and retains identical functionality: but it now just calls `PlacementPass` with `Placement`. The difference here is that placement.apply(cu) with default Placement now assigns everything, so the application of `NaivePlacementPass `afterwards now returns `false`.

pytket/tests/predicates_test.py
```diff
@@ -224,29 +234,22 @@ def test_routing_and_placement_pass() -> None:
     pl = Placement(arc)
     routing = RoutingPass(arc)
     placement = PlacementPass(pl)
-    nplacement = NaivePlacementPass(arc)
+    # nplacement = NaivePlacementPass(arc)
```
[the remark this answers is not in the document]
sjdilkes: added back

pytket/tests/predicates_test.py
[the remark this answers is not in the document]
sjdilkes: but also, done (re added)

tket/src/Placement/GraphPlacement.cpp
```diff
@@ -0,0 +1,308 @@
...
+      q_graph.add_node(node1);
+    }
+    bool e_01_exists = q_graph.edge_exists(node0, node1);
+    bool e_10_exists = q_graph.edge_exists(node1, node0);
+    if (e_01_exists || e_10_exists) {
+      throw std::invalid_argument(
+          "Graph can only have a single edge between a pair of Node.");
+    }
+    // TODO: add termination condition for for loop if all node out degrees are
+    // reached
```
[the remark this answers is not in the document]
sjdilkes: Already done! removed

tket/src/Placement/LinePlacement.cpp
```diff
@@ -0,0 +1,133 @@
...
+      [](qubit_vector_t x, qubit_vector_t y) { return (x.size() > y.size()); });
+  // Remove single qubit lines
+  while (!line_pattern.empty() && line_pattern.back().size() < 2) {
+    n_unused_nodes++;
+    line_pattern.pop_back();
+  }
+  // We want to find a useable subgraph of the Architecture for assigning these
+  // Qubit Note that this currently requires removing Nodes from the
+  // Architecture so we produce a copy
+  // TODO: see TKET-2494
```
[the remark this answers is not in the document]
sjdilkes: yeah, done

tket/src/Placement/GraphPlacement.cpp
```diff
@@ -0,0 +1,308 @@
...
+  std::vector<WeightedEdge> weighted_pattern_edges =
+      this->default_pattern_weighting(circ_);
+  std::vector<boost::bimap<Qubit, Node>> all_bimaps =
+      this->get_all_weighted_subgraph_monomorphisms(
+          circ_, weighted_pattern_edges, false);
+  std::vector<std::map<Qubit, Node>> all_qmaps;
+  unsigned counter = 0;
+  for (auto it = all_bimaps.begin();
+       it != all_bimaps.end() && counter < matches; ++it) {
+    // TODO: clean up the solution by removing low cost nodes/unconnected nodes
```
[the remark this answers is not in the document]
sjdilkes: So can remove TODO I guess?
Done

tket/src/Placement/Placement.cpp
```diff
@@ -123,52 +42,58 @@ void fill_partial_mapping(
...
-  return {get_placement_map(circ_)};
-}
+std::map<Qubit, Node> Placement::get_placement_map(const Circuit& circ_) {
+  std::vector<std::map<Qubit, Node>> all_maps =
+      this->get_all_placement_maps(circ_, 1);
 
-qubit_mapping_t NaivePlacement::get_placement_map(const Circuit &circ_) const {
-  return get_all_placement_maps(circ_).at(0);
+  // basic handling to avoid segmentation faults, as placement method may not
+  // return any valid map
```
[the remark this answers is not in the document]
sjdilkes: I'm not sure of a better solution, as anyone could write a method later that doesn't guarantee a map is returned by `get_all_placement_maps` (I don't expect it to occur with the current methods), in which case we don't want to return nothing.

tket/src/Placement/RelabelledGraphWSM.hpp
```diff
@@ -29,7 +29,7 @@ namespace WeightedSubgraphMonomorphism {
 template <class VertexType, class GraphType>
 class RelabelledGraphWSM {
  public:
-  explicit RelabelledGraphWSM(const GraphType& graph) {
+  explicit RelabelledGraphWSM(GraphType& graph) {
```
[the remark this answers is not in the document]
sjdilkes: Added back, was to do with distance caching but I guess I made and reverted a change as I went.

tket/src/Predicates/PassGenerators.cpp
```diff
@@ -352,7 +353,7 @@ PassPtr gen_placement_pass_phase_poly(const Architecture& arc) {
   // record pass config
   nlohmann::json j;
   j["name"] = "PlacementPass";
-  PlacementPtr pp = std::make_shared<GraphPlacement>(arc);
+  Placement::Ptr pp = std::make_shared<GraphPlacement>(arc, 100, 100);
```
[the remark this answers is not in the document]
sjdilkes: exposed these here, and in methods that call

tket/tests/tkettestsfiles.cmake
sjdilkes: Oops, thought I had. Done

tket/tests/Placement/test_Placement.cpp
sjdilkes: They have not all been moved over identically, but I've made sure that all the tests covering basic functionality are matched (in a neater fashion), and that any tests that were clearly there to catch some known bug at some point (i.e. a peculiar architecture or circuit construction) were moved over.

tket/tests/Circuit/test_Boxes.cpp
sjdilkes: This was my attempt to please the CI

tket/tests/Circuit/test_Boxes.cpp
sjdilkes: Also CI pleasing

tket/src/Graphs/include/Graphs/QubitGraph.hpp
```diff
@@ -0,0 +1,34 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include "Utils/GraphHeaders.hpp"
+
+namespace tket {
```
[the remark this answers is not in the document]
sjdilkes: Very reasonable, I've moved it over!

pytket/binders/placement.cpp
```diff
@@ -195,40 +165,40 @@ PYBIND11_MODULE(placement, m) {
...
+          py::arg("timeout") = 1000, py::arg("maximum_pattern_gates") = 100,
+          py::arg("maximum_pattern_depth") = 100,
+          py::arg("node_errors") = py::dict(),
           py::arg("link_errors") = py::dict(),
           py::arg("readout_errors") = py::dict())
-      .def(
-          "__repr__",
-          [](const Placement &) { return "<tket::NoiseAwarePlacement>"; })
-      .def(
-          "modify_config",
```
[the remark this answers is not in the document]
sjdilkes: Is there a standard way to make it give the warning? I've made it return `None` and cout, but it feels like these days we'd have a slightly more formal way of doing it (though I couldn't find an example in the other binders)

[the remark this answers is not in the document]
cqc-alec: How about the answer given here: https://stackoverflow.com/questions/62465062/best-way-to-mark-a-pybind11-binding-as-deprecated ?

pytket/binders/placement.cpp
```diff
@@ -195,40 +165,40 @@ PYBIND11_MODULE(placement, m) {
...
+          ":param maximum_matches: The total number of weighted subgraph "
+          "monomorphisms that can be found before matches are returned.\n"
+          ":param timeout: Total time in seconds before stopping search for "
+          "monomorphisms.\n"
+          ":param maximum_pattern_gates: The upper bound on the number of "
+          "circuit gates used to construct the pattern graph for finding "
+          "subgraph monomorphisms.\n"
+          ":param maximum_pattern_depth: The upper bound on the circuit depth "
+          "gates are added to the pattern graph to for finding subgraph "
+          "monomorphisms.\n"
```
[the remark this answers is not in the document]
sjdilkes: I'm hoping that with the extended constructor description this makes more sense now. From a user perspective, I feel like making a wsm_params struct and passing it in is less intuitive than having them as arguments to the constructor (on the condition the descriptions make sense).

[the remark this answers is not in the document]
sjdilkes: However, if you think it makes more sense I'm happy to go make changes!

pytket/tests/placement_test.py
```diff
@@ -99,34 +105,6 @@ def test_placements_serialization() -> None:
     assert NoiseAwarePlacement.from_dict(noise_pl_serial).to_dict() == noise_pl_serial
 
 
-def test_placement_config() -> None:
```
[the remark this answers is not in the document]
sjdilkes: Added back

tket/src/Placement/GraphPlacement.cpp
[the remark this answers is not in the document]
cqc-alec: Actually I think I'd leave the TODO there; it is useful for people reading the code.

tket/tests/Circuit/test_Boxes.cpp
cqc-alec: OK I see, but even so I don't think it makes sense to `REQUIRE` these exact decompositions, since we don't promise any specific circuit here. (Test coverage would not be affected by removing the circuit-equality checks.)

tket/tests/Circuit/test_Boxes.cpp
cqc-alec: Same comment.

pytket/binders/placement.cpp
```diff
@@ -136,7 +144,22 @@ PYBIND11_MODULE(placement, m) {
...
           "The constructor for a GraphPlacement object. The Architecture "
           "object describes the connectivity "
-          "between qubits.\n\n:param arc: An Architecture object.\n"
+          "between qubits. To find a qubit to node assignment, this method "
+          "constructs a pattern graph where vertices are Circuit qubits and "
+          "edges mean a pair of qubits have an interaction in the circuit, "
+          "and then tries to find a weighted subgraph monomorphsim to the "
+          "architecture connectivity, or target, graph. Edges in the pattern "
+          "graph are weighted by the circuit depth the interaction between a "
+          "pair of qubit occurs at. "
```
cqc-alec, line 153:
```suggestion
          "graph are weighted by the circuit depth at which the interaction between a "
          "pair of qubit occurs. "
```
Slightly easier to parse I think.

pytket/binders/placement.cpp
```diff
@@ -136,7 +144,22 @@ PYBIND11_MODULE(placement, m) {
...
+          "The number of edges added to the pattern graph is effected by the "
```
cqc-alec, line 154:
```suggestion
          "The number of edges added to the pattern graph is affected by the "
```

pytket/binders/placement.cpp
```diff
@@ -136,7 +144,22 @@ PYBIND11_MODULE(placement, m) {
...
+          "maximum_pattern_gates and maximum_pattern_depth arguments. "
+          "If no subgraph monomorphism can be found, "
+          "Lower edge weights are removed from the pattern graph, are more "
```
cqc-alec, line 157:
```suggestion
          "lower edge weights are removed from the pattern graph, are more "
```

pytket/binders/placement.cpp
```diff
@@ -183,22 +225,29 @@ PYBIND11_MODULE(placement, m) {
...
+          [](NoiseAwarePlacement & /*pobj*/, py::kwargs /*kwargs*/) {
+            PyErr_WarnEx(
+                PyExc_DeprecationWarning,
+                "NoiseAwarePlacement.modify_config no longer changes the "
+                "parameters for finding solutions. Please create a new "
+                "NoiseAwarePlacement object with the changed parameters.",
+                1);
+            return;
+          },
+          "Deprecated and no longer modifies paramters for finding solutions. "
```
cqc-alec, line 249:
```suggestion
          "Deprecated and no longer modifies parameters for finding solutions. "
```

pytket/tests/placement_test.py
```diff
@@ -100,31 +109,10 @@ def test_placements_serialization() -> None:
...
-    )
-    map2 = test_pl.get_placement_map(test_circuit)
-    test_pl.place(circ2)
-    assert map1 != map2
-
-    mm = MappingManager(test_architecture)
-    mm.route_circuit(circ1, [LexiLabellingMethod(), LexiRouteRoutingMethod()])
-    mm.route_circuit(circ2, [LexiLabellingMethod(), LexiRouteRoutingMethod()])
-    assert circ1.n_gates < circ2.n_gates
+    assert test_pl.modify_config() == None
```
cqc-alec, line 115:
```suggestion
    assert test_pl.modify_config() is None
```
(This is a better test since `==` can always be overridden.)
But I think we should include a bit more in this test -- just to make sure `test_pl.place()` still works after this. There may be user (or extensions) code that is doing this and we don't want to break it.

tket/src/Predicates/include/Predicates/PassGenerators.hpp
```diff
@@ -102,19 +102,50 @@ PassPtr aas_routing_pass(
  * @param lookahead parameter for the recursion depth in the algorithm, the
  * value should be > 0
  * @param cnotsynthtype parameter for the type of cnot synth
+ * @param graph_placement_maximum_matches parameter effecting the number of
```
cqc-alec, line 105:
```suggestion
 * @param graph_placement_maximum_matches parameter affecting the number of
```

tket/src/Predicates/include/Predicates/PassGenerators.hpp
```diff
@@ -102,19 +102,50 @@ PassPtr aas_routing_pass(
...
+ * matches found during the GraphPlacement substep
+ * @param graph_placement_timeout timeout for finding subgraph monomorphisms
```
cqc-alec, line 107: Please specify units.

tket/src/Predicates/include/Predicates/PassGenerators.hpp
```diff
@@ -102,19 +102,50 @@ PassPtr aas_routing_pass(
...
+ * during the GraphPlacement substep
+ * @param graph_placement_maximum_pattern_gates parameter effecting the size
```
cqc-alec, line 109:
```suggestion
 * @param graph_placement_maximum_pattern_gates parameter affecting the size
```

tket/src/Predicates/include/Predicates/PassGenerators.hpp
```diff
@@ -102,19 +102,50 @@ PassPtr aas_routing_pass(
...
+ * of the target graph, constructed from a phase polynomial, during
+ * the GraphPlacement substep, by restricting the number of gates in
+ * the phase polynomial used
+ * @param graph_placement_maximum_pattern_depth parameter effecting
```
cqc-alec, line 113:
```suggestion
 * @param graph_placement_maximum_pattern_depth parameter affecting
```

tket/tests/Circuit/test_Boxes.cpp
sjdilkes: Removed exact decomposition check as above

pytket/binders/placement.cpp
```diff
@@ -165,26 +140,57 @@ PYBIND11_MODULE(placement, m) {
...
+            return;
           },
-          "Overides default Placement parameters to given values. Timeout is "
-          "in milliseconds"
-          "\n:param \\**kwargs: Parameters for placement: "
-          "(int)depth_limit=5, (int)max_interaction_edges=edges in "
-          "the "
-          "device graph, (int)max_matches=10000, "
-          "(int)contraction_ratio=10, (int)timeout=60000.");
+          "Deprecated and no longer modifies paramters for finding solutions. "
```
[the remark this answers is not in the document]
sjdilkes: done

pytket/binders/placement.cpp
sjdilkes: done

pytket/binders/placement.cpp
sjdilkes: done

pytket/binders/placement.cpp
sjdilkes: done

pytket/tests/placement_test.py
sjdilkes: updated test!

tket/src/Predicates/include/Predicates/PassGenerators.hpp
sjdilkes: done

tket/src/Predicates/include/Predicates/PassGenerators.hpp
sjdilkes: done

cqc-alec (approved):

PR #556, opened 2022-09-29
add wasm file checks

cqc-melf:
add checks of the wasm files, the functions and the functions signatures


cqc-melf:
see https://github.com/CQCL/tket/pull/561

Issue #514, opened 2022-09-10
Segfault from `get_commands()` and error from `CXMappingPass`


cqc-alec:
Not sure if this is one bug or two. Both seem to be regressions in 1.6.0 (the code below works in 1.5.2).

```python
from pytket.architecture import Architecture
from pytket.circuit import Circuit
from pytket.passes import CXMappingPass, SynthesiseTket
from pytket.placement import Placement

c = Circuit(3, 3)
c.H(1).CX(1, 2).CX(0, 1)
c.Measure(0, 0)
c.Measure(1, 1)
c.X(2, condition_bits=[0, 1], condition_value=1)

arch = Architecture([(0, 1), (0, 2), (1, 2)])
placement = Placement(arch)
p = CXMappingPass(arch, placement, delay_measures=False)

SynthesiseTket().apply(c)
# print(c.get_commands())  # Segfault!
p.apply(c)  # Error!
```

The segfault may be related to #475 but I do not know.

The error is:
```
RuntimeError: No RoutingMethod suitable to map given subcircuit.
```

PR #557, opened 2022-09-30
[bugfix] Make commute_through_multis() work with conditional gates

lmondada:
Classical control of single-qubit operations make it harder to commute them through the circuit, as operation conditioned on classical controls can introduce dependencies between qubits that aren't obvious at first glance and can introduce cycles in the DAG. The bug #514 was due to this.

Solving this would require checking every time that two gates should be commuted whether this will break causal order. In the worst case this can take up to O(n) time in the size of the circuit. Alternatively, this could be pre-computed once before running the pass. Saving this causal ordering in a space-efficient way would require some work.

I do not think solving this case is a priority at the moment. For now I have just disabled gate commutations for single-qubit gates that are classically controlled. This reduces the power of this pass, but I think that these cases remain rare for now and fixing it would either require a heavy runtime cost or a lot of software engineering.

I'm adding a task about it and a `TODO` in the code.


lmondada:
I haven't touched any Python file, so I think the failure of the python coverage test can safely be ignored.

cqc-alec:
> I haven't touched any Python file, so I think the failure of the python coverage test can safely be ignored.

Yeah, I had the same issue in #558 -- where I added a few more tests. If we merge that first this problem should go away.

tket/src/Transformations/BasicOptimisation.cpp
```diff
@@ -206,33 +224,33 @@ static bool commute_singles_to_front(Circuit &circ) {
...
+          EdgeVec rewire_edges;
+          op_signature_t edge_types;
+          for (const Edge &e : circ.get_in_edges(prev_v)) {
+            EdgeType type = circ.get_edgetype(e);
+            Edge boundary_edge;
+            if (type == EdgeType::Quantum) {
+              boundary_edge = circ.get_last_edge(current_v, current_e);
+            } else {
+              // There should not be any Classical edges for conditional ops
+              TKET_ASSERT(type == EdgeType::Boolean);
```
cqc-alec, line 242: I'm still a bit uneasy about this -- even though you added a test with conditional measurements. (Are we OK there because they only have one quantum wire?) I don't think we can really guarantee this, so would prefer an exception be thrown rather than an assert.

cqc-alec (changes requested):

tket/src/Transformations/BasicOptimisation.cpp
lmondada: Of course I can replace this with an exception. But this is really something that should be documented and we should have a definite answer to. More generally a docs page about how classical wires and conditions work exactly would be very useful.

I have to say that personally I keep getting surprised with conditional wires -- hence also this bug for example.

lmondada: Done :)

tket/src/Transformations/BasicOptimisation.cpp
```diff
@@ -238,8 +238,12 @@ static bool commute_singles_to_front(Circuit &circ) {
             if (type == EdgeType::Quantum) {
               boundary_edge = circ.get_last_edge(current_v, current_e);
             } else {
-              // There should not be any Classical edges for conditional ops
-              TKET_ASSERT(type == EdgeType::Boolean);
+              if (type != EdgeType::Boolean) {
+                // There should not be any Classical edges for conditional ops
+                TKET_ASSERT(type == EdgeType::Classical);
+                throw CircuitInvalidity(
+                    "Malformed conditional operation: found Classical wire");
```
cqc-alec, line 245: Sorry to quibble again, but is it actually malformed? I may be missing something but I still don't see why this couldn't arise in some legitimate way?

cqc-alec (approved):

tket/src/Transformations/BasicOptimisation.cpp
lmondada: Turns out this was very easy.

cqc-alec (approved):

tket/src/Transformations/BasicOptimisation.cpp
cqc-alec: Thanks! I had missed the fact that `ends_commute()` guarantees this condition.

PR #559, opened 2022-10-03
Improve documentation of default_compilation_pass().

cqc-alec:


pytket/pytket/backends/backend.py
```diff
@@ -150,11 +150,18 @@ def default_compilation_pass(self, optimisation_level: int = 1) -> BasePass:
         will be suitable to run on this backend with as few preconditions as
```
CalMacCQ, line 150: I'm not quite sure what "as few preconditions as possible" means here. Perhaps the wording could be improved?

Surely all the preconditions/predicates should be met if the circuit the circuit is to be run on a `Backend`.

PR #560, opened 2022-10-03
add notebook link

CalMacCQ:
Add notebook link under "more dcoumentation"


pytket/docs/index.rst
```diff
@@ -128,6 +128,7 @@ Licensed under the `Apache 2 License <http://www.apache.org/licenses/LICENSE-2.0
     
     Manual <https://cqcl.github.io/pytket/manual/index.html>
     Extensions <https://cqcl.github.io/pytket-extensions/api/index.html>
+    Example notebooks <https://github.com/CQCL/pytket/tree/main/examples>
```
cqc-melf, line 131: is there an index.html file somewhere in the directory with more information?

cqc-melf, line 131:
```suggestion
    Example notebooks <https://github.com/CQCL/pytket/tree/main/examples#pytket-examples>
```

cqc-melf: How about this?

CalMacCQ: So I built the docs locally and then checked the built index.html file - the link seemed to work. Is this what you mean?

cqc-melf: I was thinking about changing the added link in a way that is point to an index.html file in the directory of the examples

cqc-melf: (see the comment below)

CalMacCQ: I tried out the suggested change and this seems to redirect to the README of the examples repo. Is this what you intended? If so this is fine with me.

Is there another file I should change as well?

cqc-melf:
This is something we want in the release? Probably needs a rebase from develop and it should work then.

cqc-melf (approved):

PR #563, opened 2022-10-03
Set hypothesis seed when generating coverage report.

cqc-alec:


cqc-melf (approved):
Do we want this in the release, happy to approve it.

PR #566, opened 2022-10-04
Specify bit when possible.

cqc-alec:
This makes the output more explicit, in cases where the target register width exceeds the source width,


ss2165 (approved):
Thanks Alec

PR #569, opened 2022-10-04
Revert "add wasm file checks with wasmer"

cqc-melf:
Have to revert the PR because of M1 problems with python 3.8

Reverts CQCL/tket#561


cqc-alec (approved):

cqc-alec:
I think the coverage failure can be ignored.

PR #572, opened 2022-10-04
fix m1 python 3.8 wasm issue

cqc-melf:


cqc-melf:
Some tests are still failing because of: https://github.com/CQCL/tket/pull/571

cqc-melf:
(And I have seen that python 3.10 has the same problem, so I am skipping that as well. :( )

cqc-alec:
Following the discussion today I am inclined to abandon wasmer.

.github/workflows/build_and_test.yml
```diff
@@ -254,7 +254,7 @@ jobs:
     - name: Install pybind11
       run: conan create --profile=tket recipes/pybind11
     - name: Build pytket (3.8)
-      if: github.event_name == 'push'
+      if: github.event_name == 'pull_request'
```
cqc-alec, line 257: Presumably this change (and below) should be reverted.

pytket/pytket/wasm/wasm.py
```diff
@@ -16,24 +16,182 @@
...
+
+        # stores the names of the functions mapped
+        #  to the number of parameters and the number of return values
+        self._functions = dict()
+
+        # contains the list of functions that are not allowed
+        # to use in pytket (because of types that are not i32)
+        self._unsupported_function = []
+
+        # only run the wasmer import and so on if the check is requested to run
```
cqc-alec, line 80: Is this comment out of date? There is no wasmer now?

pytket/pytket/wasm/wasm.py
```diff
@@ -16,24 +16,182 @@
...
+    }
+
+    def __init__(self, filepath: str, check_file: bool = True):
+        """construct a wasm file handler
+        :param filepath: path to the wasm file
+        :type filepath: str
+        :param check_file: If this parameter is set to True the wasm
+        file will be checked if the funtion signatures are supported.
+        This check is currently NOT supported on M1 mac in
+        python 3.8 and 3.10.
```
cqc-alec, line 52: Is this still true?

pytket/pytket/wasm/wasm.py
```diff
@@ -16,24 +16,182 @@
...
+        """checks a given function name and signature if it is included
+        :param function_name: name of the function that is checked
+        :type function_name: str
+        :param number_of_parameters: number of i32 parameters of the function
+        :type number_of_parameters: int
+        :param number_of_returns: number of i32 return values of the function
+        :type number_of_returns: int
+        :return: true if the signature and the name of the function is correct"""
+        if not self._check_file:
+            return True
```
cqc-alec, line 192: Perhaps this should raise an exception rather than return True?

pytket/setup.py
```diff
@@ -254,6 +254,7 @@ def finalize_options(self):
         "jinja2 ~= 3.0",
         "types-pkg_resources",
         "typing-extensions ~= 4.2",
+        "qwasm ~= 1.0",
```
cqc-alec, line 257: Is qwasm open-source?

cqc-alec (changes requested):

pytket/pytket/wasm/wasm.py
cqc-melf: No, I will remove this.

pytket/pytket/wasm/wasm.py
cqc-melf: The user can't choose if the check is executed. I was thinking about if this hole option to opt out of the check should be removed...

pytket/setup.py
cqc-melf: Yes:
https://github.com/CQCL/qwasm
https://pypi.org/project/qwasm/

.github/workflows/build_and_test.yml
cqc-melf: Yes

pytket/pytket/wasm/wasm.py
cqc-melf: I think we don't need it, it is only making thinks more complicated and in-transparent for the user.

pytket/setup.py
cqc-alec: Cool, would be good to add a link to the repo from the pypi page.

cqc-alec (commented):
Happy to approve for merge once https://github.com/CQCL/tket/actions/runs/3188604820 has completed.

cqc-melf:
I have added the link on https://pypi.org/project/qwasm/

cqc-alec (approved):

PR #593, opened 2022-10-09
Set tket-libs as the only remote.

cqc-alec:
Currently, we have stored some packages on our artifactory for configurations that aren't built by default on conan-center (e.g. apple-clang 14, VSCode 17). This means that if a new _revision_ of a package recipe is uploaded to conan-center, until we upload the corresponding revision, our builds will fail (because they look for the most recent revision and then look for packages matching the configuration for that revision).

We solve that problem by only using our artifactory (not conan-center) during builds.

I've added a workflow script that can be used to uplaod latest revisions of any required packages. We can choose to run this if we want a newer revsion (or package version).

Another advantage is that we have control over the exact versions of all dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,11 @@ recommended in the warning message:
 conan profile update settings.compiler.libcxx=libstdc++11 tket
 ```
 
-Set the `tket-libs` repository as your remote:
+Set the `tket-libs` repository as your remote. (Note that the following commands
+affect your conan configuration across all projects, so if you are working on
+other projects with conan you will want to revert them afterwards. A simple way
+is to back up the file `~/.conan/remotes.json`. You can view your current
+remotes list with `conan remote list`.)
 
 ```shell
 conan remote clean
--- a/README.md
+++ b/README.md
@@ -82,9 +82,10 @@ recommended in the warning message:
 conan profile update settings.compiler.libcxx=libstdc++11 tket
 ```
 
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -82,9 +82,10 @@ recommended in the warning message:
 conan profile update settings.compiler.libcxx=libstdc++11 tket
 ```
 
-Add the `tket.libs` repository to your remotes:
+Set the `tket-libs` repository as your remote:
 
 ```shell
+conan remote clean
```
cqc-melf, line 88: Do we want to add more explanations on this here in the readme why we are doing this? Is this something we are recommending for all users? Does this have negative side effects when working on other project as well?

cqc-alec: This is recommended when building tket, yes. It might affect other projects -- but then, so would the existing recommendation to add the tket-libs remote. I'll add a warning about this.

cqc-alec: Done.

cqc-melf: That is what I was assuming, thank you for adding the note.

cqc-melf (approved):

cqc-melf:
Have you seen the issue with the coverage check? That is not running? @cqc-alec

cqc-alec:
> Have you seen the issue with the coverage check? That is not running? @cqc-alec

It has to wait until "Analyse tket C++ test coverage" has finished.

PR #597, opened 2022-10-10
Feature/contextual cn x decomp

yao-cqc:
The `CnX` decomposition method we are using for moderate number of qubits (<= 50) have the form 
`CnX = Q_dagger P_dagger Q CU P`

Since CnX = CnX.dagger, we have
`CnX = P_dagger CU_dagger Q_dagger P Q`

If two CX gates are connected, then it's possible to have gates cancelled by composing `P` and `P_dagger` and reordering the control qubits. This method doesn't work when the `CnX` gates are sparse, but can optimise the circuit produced by a `ToffoliBox` a little further.

I think we should have a more general method for this contextual multi-controlled gate decomposition problem (e.g. find the optimal decomposition for each gate by some tree search),  but let's wait until we also have those decomposition methods using ancillas.


pytket/binders/circuit/library.cpp
```diff
@@ -113,7 +113,7 @@ void init_library(py::module &m) {
       "position.");
   library_m.def(
       "_CCX_normal_decomp", &CircPool::CCX_normal_decomp,
-      "Equivalent to CCX, using five CX");
+      "Equivalent to CCX, using 6 CX");
```
sjdilkes, line 116: Was it always 6 and the previous "five" was a typo?

tket/src/Circuit/include/Circuit/CircPool.hpp
```diff
@@ -124,7 +124,7 @@ const Circuit &Z0_CX();
  */
 const Circuit &CCX_modulo_phase_shift();
 
-/** Equivalent to CCX, using five CX */
+/** Equivalent to CCX, using 6 CX */
```
sjdilkes, line 127: Similarly to other comment, was this always 6?

tket/src/Transformations/Decomposition.cpp
```diff
@@ -1815,6 +1815,178 @@ Transform decomp_arbitrary_controlled_gates() {
...
+    bool success = commute_through_multis().apply(circ);
+    success |= remove_redundancies().apply(circ);
+
+    // Cache CnX decompositions
+    std::map<unsigned, Circuit> cnx_cache;
+    // Replaced vertices to delete at the end
+    VertexList bin;
+
+    // Find all CnX gates, including conditional ones
+    std::vector<Command> commands = circ.get_commands();
```
sjdilkes, line 1870: I wonder if in the future we can make runtime improvements by avoiding flattening the DAG and just iterating through graph vertices until a match is found. We can shelve it for now but it might be worth adding a very low priority TKET task for this.

sjdilkes (changes requested):
Wonderful stuff. Just a few minor comments and clarifications!

tket/src/Transformations/Decomposition.cpp
```diff
@@ -1815,6 +1815,178 @@ Transform decomp_arbitrary_controlled_gates() {
...
+  }
+
+  // Reorder the replacement circuit by renaming the qubits
+  if (new_args != std::nullopt) {
+    unit_map_t qmap;
+    for (unsigned i = 0; i + 1 < new_args.value().size(); i++) {
+      auto it =
+          find(original_args.begin(), original_args.end(), new_args.value()[i]);
+      int index = it - original_args.begin();
+      qmap.insert({Qubit(i), Qubit(index)});
```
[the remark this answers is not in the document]
yao-cqc: The circuit `cnx` we are trying to rename is generated by those CnX decomposition methods in tket, which only use default registers; but I think it is a good idea add a check here. We might have some new decomposition methods in the future that uses multiple registers (e.g. Ancilla).

tket/src/Transformations/Decomposition.cpp
```diff
@@ -1815,6 +1815,178 @@ Transform decomp_arbitrary_controlled_gates() {
...
+      bool cmd1_as_dagger = false;
+      bool cmd2_as_dagger = false;
+
+      if (args1.size() == 3 && args2.size() == 3) {
+        // CCX gates are handled differently, there
+        // is a maximum of 1 CX reduction (needs to be realised with
+        // CliffordSimp), when they have args: [a,b,c] and [c,a,b]
+        std::set<Qubit> args1_set(args1.begin(), args1.end());
+        std::set<Qubit> args2_set(args2.begin(), args2.end());
+        if (args1_set == args2_set) {
```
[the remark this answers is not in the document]
yao-cqc: We can only do optimisation if the two CCXs are acting on the same _set_ of qubits, and that is what this `if` statement is testing. The inner `if_else` checks if they have the same target qubit.

tket/src/Transformations/Decomposition.cpp
```diff
@@ -1815,6 +1815,178 @@ Transform decomp_arbitrary_controlled_gates() {
...
+          if (new_args1.back() == new_args2.back()) {
+            cmd1_as_dagger = true;
+            new_args2 = new_args1;
+          } else {
+            Qubit a = (args1[0] == args2.back()) ? args1[1] : args1[0];
+            new_args1 = {a, args2.back(), args1.back()};
+            new_args2 = {args1.back(), a, args2.back()};
+          }
+        }
+      } else {
```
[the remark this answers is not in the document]
yao-cqc: Sorry, I think this is also causing the previous confusion. I'll add comments

tket/src/Transformations/Decomposition.cpp
yao-cqc: Yeah, I agree. I think we can also consider how to pick an _optimal_ decomposition if multiple matches are found (e.g. multiple CnX, CX, etc. neighbouring a CnX).

yao-cqc:
#CX reduction vs  n_qubits
Here is a comparison between `[RemoveRedundancies, DecomposeMultiQubitsCX, RemoveRedundancies]` and `[CnXPairwiseDecomposition, DecomposeMultiQubitsCX, RemoveRedundancies]`. y-axis is the percentage of CX saving.

tket/src/Transformations/Decomposition.cpp
sjdilkes: Makes sense. You talked before about potentially producing a new IR for doing better decompositions of sequences of multi-controlled gates (plausibly some kind of "`MultiplexorGraph`"), so maybe we try to assign some time in the future for considering that.

sjdilkes (approved):
Nice!

PR #600, opened 2022-10-11
Install conan v1

cqc-alec:
Conan v2 will be arriving soon (in beta now). Until we're ready for it, let's make sure we always install v1 in the CI workflows.


cqc-melf (commented):
Does this require an update of the tket versions and new uploads?

cqc-alec:
> Does this require an update of the tket versions and new uploads?

No, it's not changing anything in the build -- just making sure that when conan 2,0 is released it won't be installed automatically and we will continue to use conan 1.

cqc-melf (approved):

PR #604, opened 2022-10-13
Fix to avoid Static Initialization Order Fiasco in proptest

cqc-alec:
I tripped over this while doing some build-system refactoring work.


yao-cqc (approved):
Thanks, didn't know this could be an issue.

PR #605, opened 2022-10-17
add links to default pass docs

CalMacCQ:
I've added some links to the documentation for ``default_compilation_pass``. Hopefully this should make it easier for users to find out which passes are applied with different levels of optimisation.

This should be up to date once these two PR's are merged
https://github.com/CQCL/pytket-qiskit/pull/8
https://github.com/CQCL/pytket-quantinuum/pull/45


cqc-alec (commented):
I don't think this is the right place for this. This is describing the abstract pytket method; surely this documentation belongs with the respective backends?

PR #606, opened 2022-10-17
Bugfix/remove unused scratch registers from qasm and fix logic exp args update

### LINKED ISSUE #567: Conversion of QASM to and from tket introduces unused `tk_SCRATCH_BIT` registers.
See for example the test files `test17_output.qasm` and `test18_output.qasm`.

Not quite a bug, but not desired behaviour.

yao-cqc:
resolves #567 and fixes the issue that ``Circuit.rename_units`` doesn't update args in ``LogicExp``.


tket/src/Circuit/include/Circuit/ClassicalExpBox.hpp
```diff
@@ -23,13 +23,34 @@
 
 namespace tket {
 
+// Used as an intermediate class in between ClassicalExpBox and Box
+class ClassicalExpBoxPyBase : public Box {
```
cqc-alec, line 27: Since there is no requirement for a `ClassicalExpBox` to be templated on a Python object, I would suggest calling this simply `ClassicalExpBoxBase`. We should however document (for `ClassicalExpBox`) the requirements of the class `T`.

tket/src/Circuit/include/Circuit/ClassicalExpBox.hpp
```diff
@@ -23,13 +23,34 @@
...
+   */
+  virtual bool rename_units(const std::map<Bit, Bit> &bm) const {
+    (void)bm;
+    return false;
+  }
+};
+
 /**
  * Holding box for abstract expressions on Bits
  * Templated by type which holds expression
```
cqc-alec, line 49: As far as I can see the requirements for `T` (which we should document here) are the copy-constructor, the `equal()` and the `attr()` methods, but maybe it is simplest to just say that it must implement https://pybind11.readthedocs.io/en/stable/reference.html#_CPPv4I0E10object_api .

cqc-alec (changes requested):
Thanks! I just made a few comments.

cqc-alec (approved):

PR #608, opened 2022-10-18
Add dependabot file to check periodically for actions updates

cqc-alec:
This should generate a new PR when any update to an external github action is detected.


cqc-melf (approved):
Good idea. I will think about adding this for the extensions, too.

PR #616, opened 2022-10-20
Fix `LinePlacement` Serialization

sjdilkes:
`COMPPASSJSONTEST` in test_json.cpp was failing for a Placement Pass (as returned by `gen_placement_pass`) constructed with some `LinePlacement` objects.

There were two points of failure for this:
1. `Architecture::get_distances(Node node)` has had its meaning (though not its type) changed at some point. Originally it returned a "lexicographical" `std::vector<size_t>` with diameter number of entries, where the value at the [diameter-distance] reflected the number of other Nodes at "distance" away from the node passed to the method. Instead, it went through every other Node in the architecture and constructed a vector where the entries were the distance between "node" and the others (see point 2 for why this ordering was different pre and post serialisation). This effected the `Node` objects removed by `Architecture::remove_worst_nodes`, as it uses a lexicographical comparison to work out which `Node` to remove. So as to not change the API since 1.0, I've updated `Architecture::find_worst_node` to convert the returned "distances" from `Architecture::get_distances` to the required lexicographical vector for comparison.
2. The BFS and DFS methods used in `Architecture` used the ordering of `Node` (converted to "vertices) given by `boost::vertices(graph)`. This ordering was different between the pre and post serialised Architectures, meaning any methods that iterated through `Node` objects (lots!) would iterate in a different order, which can cause discrepancies. I've updated `to_json` (which amusingly noted that the ordering of Nodes should be preserved) to now use the ordering of `Node` implied by `boost::vertices(graph)`.


tket/tests/test_json.cpp
```diff
@@ -632,6 +632,8 @@ SCENARIO("Test compiler pass serializations") {
   COMPPASSJSONTEST(PlacementPass, gen_placement_pass(place))
   // TKET-1419
   COMPPASSJSONTEST(NoiseAwarePlacement, gen_placement_pass(na_place))
+  COMPPASSJSONTEST(NaivePlacementPass, gen_naive_placement_pass(arc))
+  COMPPASSJSONTEST(LinePlacement, gen_placement_pass(la_place))
```
cqc-melf, line 636: Is the line placement compatible with a check of the NaivePlacementPass, too? Could you add an additional check if that is possible?

tket/src/Architecture/Architecture.cpp
```diff
@@ -120,23 +121,30 @@ std::optional<Node> Architecture::find_worst_node(
   if (bad_nodes.empty()) {
     return std::nullopt;
   }
-
+  auto make_lexicographic = [](std::vector<size_t> distances) {
+    unsigned max = distances.size();
+    std::vector<size_t> out(max + 1, size_t(0));
+    for (const auto& distance : distances) {
+      TKET_ASSERT(distance < max);
```
cqc-melf, line 128: Not sure if this should better be an exception? Have you thought about that?

cqc-melf (commented):
Two short questions and could you maybe add something to the changelog please?

tket/tests/test_json.cpp
sjdilkes: The addition of the check for `gen_naive_placement_pass`is more papering over an oversight of the last PR. `gen_naive_placement_pass` takes an architecture not a placement object and so they are not compatible

cqc-melf (approved):

PR #620, opened 2022-10-21
Feature/support qubit permutation in q control box decomp

### LINKED ISSUE #615: `QControlBox` decomposition should handle inner circuits with implicit wire swaps
Currently throws an exception with the message "Circuit has implicit wireswaps".

yao-cqc:
Resolves #615 
Fix the issue that the `QControlBox` decomposition doesn't ignore all identities.

I've also added two circuit methods `replace_SWAPs` and `replace_implicit_wire_swaps` to the python binder because I think they can be useful.

### DIFF
--- a/tket/src/Circuit/CircUtils.cpp
+++ b/tket/src/Circuit/CircUtils.cpp
@@ -616,23 +616,23 @@ static Circuit with_controls_symbolic(const Circuit &c, unsigned n_controls) {
   if (c.n_bits() != 0 || !c.is_simple()) {
     throw CircuitInvalidity("Only default qubit register allowed");
   }
-  if (c.has_implicit_wireswaps()) {
-    throw CircuitInvalidity("Circuit has implicit wireswaps");
-  }
+
+  Circuit c1(c);
+  // Replace wire swaps with SWAP gates
+  c1.replace_all_implicit_wire_swaps();
 
   // Dispose of the trivial case
   if (n_controls == 0) {
-    return c;
+    return c1;
   }
 
   static const OpTypeSet multiq_gate_set = {
       OpType::CX, OpType::CCX, OpType::CnX, OpType::CRy, OpType::CnRy,
       OpType::CZ, OpType::CnZ, OpType::CY,  OpType::CnY};
 
-  unsigned c_n_qubits = c.n_qubits();
+  unsigned c_n_qubits = c1.n_qubits();
 
   // 1. Rebase to {CX, CCX, CnX, CnRy} and single-qubit gates
-  Circuit c1(c);
   VertexList bin;
   BGL_FORALL_VERTICES(v, c1.dag, DAG) {
     Op_ptr op = c1.get_Op_ptr_from_Vertex(v);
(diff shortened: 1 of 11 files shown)


tket/src/Circuit/include/Circuit/Circuit.hpp
```diff
@@ -1342,14 +1342,23 @@ class Circuit {
...
    *
    * @param first qubits to add the wireswap on
    * @param second qubits to add the wireswap on
+   * @param using_cx use three CX gates instead of a SWAP gate
    *
    * O(c)
    */
-  void replace_implicit_wire_swap(const Qubit first, const Qubit second);
+  void replace_implicit_wire_swap(
+      const Qubit first, const Qubit second, const bool using_cx = true);
```
cqc-alec, line 1354: No need for the `const` on the `bool` since it is an integral type (so it makes no difference, a local copy is always made in the function).

tket/src/Circuit/macro_manipulation.cpp
```diff
@@ -382,17 +382,55 @@ void Circuit::replace_SWAPs() {
...
+  qubit_map_t perm = implicit_qubit_permutation();
+  std::set<Qubit> fixed_qubits;
+  // iterate permutation cycles and add swap for every adjacent elements
+  for (const std::pair<const Qubit, Qubit>& pair : perm) {
+    if (fixed_qubits.find(pair.first) != fixed_qubits.end()) {
+      // skip if visited
+      continue;
+    }
+    // start traverse a cycle
+    Qubit head = pair.first;
```
cqc-alec, line 411:
```suggestion
    const Qubit head = pair.first;
```
This helps me read the code...

tket/src/Circuit/macro_manipulation.cpp
```diff
@@ -382,17 +382,55 @@ void Circuit::replace_SWAPs() {
...
+        break;
+      }
+      replace_implicit_wire_swap(current, next, false);
+      fixed_qubits.insert(current);
+      auto it = perm.find(next);
+      if (it == perm.end()) {
+        throw CircuitInvalidity(
+            "Unknown error in replace_all_implicit_wire_swaps: invalid qubit "
+            "permutation.");
+      }
```
cqc-alec, line 427: Can we `TKET_ASSERT` here since it means a bug in our code?

tket/tests/Circuit/test_Boxes.cpp
```diff
@@ -889,6 +889,59 @@ SCENARIO("QControlBox", "[boxes]") {
...
+    c0.add_op<unsigned>(OpType::Rx, 0., {0});
+    c0.add_op<unsigned>(OpType::CRx, 4., {0, 1});
+    c0.add_op<unsigned>(OpType::noop, {0});
+    CircBox cbox(c0);
+    Op_ptr op = std::make_shared<CircBox>(cbox);
+    QControlBox qcbox(op, 1);
+    std::shared_ptr<Circuit> c = qcbox.to_circuit();
+    REQUIRE(c->n_gates() == 0);
+    REQUIRE(equiv_0(c->get_phase()));
+  }
```
cqc-alec, line 944: Could we also add a test for a `QControlBox` with a `Phase` op with a nonzero phase? (This should produce a `U1` or equivalent.)

cqc-alec (changes requested):

tket/tests/Circuit/test_Boxes.cpp
yao-cqc: Isn't there already a test for that? https://github.com/CQCL/tket/blob/6fd11d1912634dd25d0245d2ab63a9b4eaf8ae14/tket/tests/Circuit/test_Boxes.cpp#L880

tket/src/Circuit/CircUtils.cpp
```diff
@@ -923,9 +923,11 @@ static Circuit with_controls_numerical(const Circuit &c, unsigned n_controls) {
   std::vector<Command> commands = c1.get_commands();
   std::vector<CnGateBlock> blocks;
 
-  for (const Command &c : commands) {
-    if (c.get_op_ptr()->get_type() != OpType::noop) {
-      blocks.push_back(CnGateBlock(c));
+  for (const Command &cmd : commands) {
+    // drop any gates that apply identity to the target qubit
+    if (cmd.get_op_ptr()->commuting_basis(cmd.get_args().size() - 1) !=
+        Pauli::I) {
```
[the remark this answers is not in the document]
yao-cqc: Good idea!

tket/tests/Circuit/test_Boxes.cpp
cqc-alec: Ah, so there is!

cqc-alec (approved):

PR #621, opened 2022-10-24
Replace `PauliGraph::TopSortIterator` with `boost::topological_sort`

cqc-alec:
While working on build system refactoring I encountered a mysterious segfault on Windows in the PauliGraph tests. This turned out to be due to corrupt graph iterators. The `TopSortIterator` has a raw pointer to a graph, which is only initialized when `begin()` is called and is assumed to remain valid.

I added a `PauliGraph::sanity_check()` method which provoked the segfault on Linux. This method doesn't do a great deal of sanity checking; it could be extended in the future if we wish. It's only called from the tests.

I replaced the `TopSortIterator` with `boost::topological_sort`, copying the method used in `Circuit`. The price we pay for this (in order to retain constness of other methods) is having to make `PauliGraph::graph_` an explicitly mutable member, because sorting requires indexing of the vertices.

### DIFF
--- a/tket/src/PauliGraph/PauliGraph.cpp
+++ b/tket/src/PauliGraph/PauliGraph.cpp
@@ -317,86 +317,6 @@ void PauliGraph::apply_pauli_gadget_at_end(
   if (get_predecessors(new_vert).empty()) start_line_.insert(new_vert);
 }
 
-PauliGraph::TopSortIterator::TopSortIterator()
-    : pg_(nullptr),
-      current_vert_(boost::graph_traits<PauliDAG>::null_vertex()) {}
-
-PauliGraph::TopSortIterator::TopSortIterator(const PauliGraph &pg) {
-  pg.sanity_check();
-  if (pg.start_line_.empty()) {
-    current_vert_ = boost::graph_traits<PauliDAG>::null_vertex();
-    return;
-  }
-  pg_ = &pg;
-  for (const PauliVert &vert : pg.start_line_) {
-    search_set_.insert({pg.graph_[vert].tensor_, vert});
-  }
-  current_vert_ = search_set_.begin()->second;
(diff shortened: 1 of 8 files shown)


tket/src/PauliGraph/PauliGraph.cpp
```diff
@@ -420,4 +423,26 @@ void PauliGraph::to_graphviz(std::ostream &out) const {
   out << "}";
 }
 
+void PauliGraph::sanity_check() const {
+  for (TopSortIterator it = begin(); it != end(); ++it) {
```
yao-cqc: What exactly caused the segfault? Since the `begin` is called at the beginning, why is the `TopSortIterator` corrupt?

yao-cqc: This commit has a `pg.sanity_check();` in the constructor `TopSortIterator(const PauliGraph &pg)`. Since `sanity_check()` calls `begin()`, which then calls the `TopSortIterator` constructor, maybe the segfault was caused by this infinite loop?

cqc-alec: That's a good point and it could be. I saw this segfault on Windows with no changes to the `PauliGraph` code (see https://github.com/CQCL/tket/actions/runs/3313870277/jobs/5472416269 ), and tried to reproduce locally, but maybe I actually failed.

However, the change in this PR _did_ fix the segfault I was observing on Windows. (See https://github.com/CQCL/tket/actions/runs/3314406374/jobs/5476363405 -- the failure on that run is unrelated and happened after the test run successfully completed).

So I'd still like to merge this even without fully understanding the original problem.

yao-cqc: I agree. The new code is definitely better.

yao-cqc (approved):

PR #626, opened 2022-10-31
Bugfix zeropad oversized register

Roland-djee:
This PR addresses the bug that when `add_c_setreg` is called on a register with a width > `_TKET_REG_WIDTH` (32), the bits above aren't properly zero-padded. When reading that register, it leads to undefined behaviour therefore random bits retrieval (especially on Windows). The big fix is to zero-pad to the full length of the register.


pytket/binders/circuit/Circuit/add_classical_op.cpp
```diff
@@ -242,15 +242,19 @@ void init_circuit_add_classical_op(
             std::vector<bool> vals(reg.size());
             for (unsigned i = 0; i < reg.size(); i++) {
               args[i] = reg[i];
-              vals[i] = bs[i];
+	      vals[i] = false;
+	      if (i < _TKET_REG_WIDTH) {
+		  vals[i] = bs[i];
+	      }
```
cqc-alec, line 248:
```suggestion
	      vals[i] = (i < _TKET_REG_WIDTH) && bs[i];
```
This is simpler I think.

pytket/binders/circuit/Circuit/add_classical_op.cpp
```diff
@@ -242,15 +242,19 @@ void init_circuit_add_classical_op(
             std::vector<bool> vals(reg.size());
             for (unsigned i = 0; i < reg.size(); i++) {
```
cqc-alec, line 243: As discussed I think we should raise an exception if `value` does not fit in the register, i.e. if
```
value >> reg.size() != 0
```

cqc-alec (changes requested):

pytket/tests/classical_test.py
```diff
@@ -130,6 +130,82 @@ def test_c_ops() -> None:
     assert str(cmds[3]) == "SetBits(110) c0[0], c0[1], c0[2];"
 
 
+def test_add_c_setreg_with_size_gt_32bits() -> None:
+    c = Circuit()
+    b = c.add_c_register("b", 64)
+    c.add_c_setreg(100, b)
+
+    expected_reg = [
```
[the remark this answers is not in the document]
Roland-djee: Done.

pytket/binders/circuit/Circuit/add_classical_op.cpp
Roland-djee: Thanks.

pytket/binders/circuit/Circuit/add_classical_op.cpp
Roland-djee: Done.

pytket/binders/circuit/Circuit/add_classical_op.cpp
```diff
@@ -237,20 +237,26 @@ void init_circuit_add_classical_op(
           "add_c_setreg",
           [](Circuit &circ, const _tket_uint_t value, const BitRegister &reg,
              const py::kwargs &kwargs) {
+	    unsigned int value_int = value;
+	    unsigned int tket_reg_width = _TKET_REG_WIDTH;
```
cqc-alec, line 241: Why are these conversions necessary? Also, this is the wrong check: we want to test against `reg.size()`, not `_TKET_REG_WIDTH`.

cqc-alec (changes requested):

pytket/binders/circuit/Circuit/add_classical_op.cpp
Roland-djee: > Why are these conversions necessary? 

If I don't do it, I get the following error:

```sh
 /Users/roland/Desktop/Projects/Tket/tket/pytket/binders/circuit/Circuit/add_classical_op.cpp:242:20: error: shift count >= width of type [-Werror,-Wshift-count-overflow]
                  if (value_int >> _TKET_REG_WIDTH != 0) {
                                ^  ~~~~~~~~~~~~~~~
      1 error generated.
```

which I interpret as being unable to define a right shift operator for an untyped variable defined in a macro.

> Also, this is the wrong check: we want to test against `reg.size()`, not `_TKET_REG_WIDTH`.

Is it ? My understanding was that we wanted to check if the value could be held on the tket register. But I understood that if I try to set a value that overflows the 32-bit register, I get a type error of some sort:

```sh
>>> a = c.add_c_register("a", 32)
>>> c.add_c_setreg(10 ** 24, a)
  File "<stdin>", line 1, in <module>
TypeError: add_c_setreg(): incompatible function arguments. The following argument types are supported:
    1. (self: pytket._tket.circuit.Circuit, value: int, arg: pytket._tket.circuit.BitRegister, **kwargs) -> pytket._tket.circuit.Circuit

Invoked with: [], 1000000000000000000000000, BitRegister("a", 32)
```

Unless you meant checking that a value could be held on a register of smaller size ?

cqc-alec: The value will always fit in 32 bits because it's a `uint32_t`. Don't we want to check that it fits in this particular register?

Roland-djee: Yes, a register of a smaller size. At first, I thought for a register of a larger size which didn't make much sense.

Roland-djee: Done.

pytket/binders/circuit/Circuit/add_classical_op.cpp
```diff
@@ -237,20 +237,24 @@ void init_circuit_add_classical_op(
           "add_c_setreg",
           [](Circuit &circ, const _tket_uint_t value, const BitRegister &reg,
              const py::kwargs &kwargs) {
+	    if (value >> reg.size() != 0) {
```
cqc-alec: It appears that when `reg.size() >=32` the shift amount is reduced mod 32, so we actually want
```suggestion
	    if (reg.size() < 32 && (value >> reg.size()) != 0) {
```

pytket/binders/circuit/Circuit/add_classical_op.cpp
```diff
@@ -237,20 +237,24 @@ void init_circuit_add_classical_op(
...
+	      throw std::runtime_error("Value " STR(value) " cannot be held on a " STR(reg.size()) "-bit register.");
```
cqc-alec: This `STR` doesn't work here; I think if you want to include the values in the string you have to explicitly construct the string from a stringstream.

cqc-alec (changes requested):

pytket/binders/circuit/Circuit/add_classical_op.cpp
Roland-djee: Yes, this is in the last commit.

pytket/binders/circuit/Circuit/add_classical_op.cpp
Roland-djee: Ditto, in the last commit.

cqc-alec (approved):

PR #633, opened 2022-11-02
Update GraphPlacement to allow Barrier gates

sjdilkes:
There is a condition in `default_pattern_weighting` that stops `GraphPlacement` from being run on `Circuits` with multi-qubit quantum gates with more than two qubits. However, Barrier gates are multi qubit (non-quantum) gates with more than two qubits and this condition does not make a special case for them. This PR updates the condition to make sure an error isn't thrown for Barrier gates. It also makes sure that two-qubit Barrier gates aren't placed.

### DIFF
--- a/tket/src/Placement/GraphPlacement.cpp
+++ b/tket/src/Placement/GraphPlacement.cpp
@@ -47,11 +47,11 @@ GraphPlacement::default_pattern_weighting(const Circuit& circuit) const {
        gate_counter < this->maximum_pattern_gates_ && !frontier.slice->empty();
        i++) {
     for (const Vertex& vert : *frontier.slice) {
-      OpType optype = circuit.get_OpType_from_Vertex(vert);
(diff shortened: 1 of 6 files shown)


tket/src/Placement/GraphPlacement.cpp
```diff
@@ -47,10 +47,11 @@ GraphPlacement::default_pattern_weighting(const Circuit& circuit) const {
        gate_counter < this->maximum_pattern_gates_ && !frontier.slice->empty();
        i++) {
     for (const Vertex& vert : *frontier.slice) {
+      OpType optype = circuit.get_OpType_from_Vertex(vert);
```
cqc-alec, line 50:
```suggestion
      if (circuit.get_OpType_from_Vertex(vert) == OpType::Barrier) continue;
```
And revert the changes below? Maybe this is simpler?

cqc-alec (approved):

PR #636, opened 2022-11-02
Coherent Tableau

willsimmons1465:
Took a bit longer than expected to get the synthesis algorithm working, but we now have a working draft implementation for Coherent Tableaux. This PR covers the basic data structure, manipulations, and conversions to and from circuits. The conversion to circuits will need amendment once we support post-selections in tket. The main current use case for this is in the upcoming PauliGraph refactor (useful for generalising PauliSimp and ZX/MBQC extraction), but could also be provided with a Python interface and boxes like we have for UnitaryTableau.


cqc-alec (commented):
It's now a requirement when a PR changes the C++ that the tket version be bumped in:
* recipes/tket/conanfile.py
as well as the required tket version in
* recipes/tket-tests/conanfile.py
* recipes/tket-proptests/conanfile.py
* pytket/conanfile.txt
to match it. That's why the "check tket version" test failed. Could you make this change please? Thanks!

tket/src/Clifford/CoherentTableau.cpp
```diff
@@ -0,0 +1,666 @@
...
+    }
+  }
+  if (x_row) {
+    // Remove the X stabilizer
+    remove_row(*x_row);
+  }
+}
+
+void CoherentTableau::remove_row(unsigned row) {
+  unsigned n_rows = get_n_rows();
```
cqc-alec, line 468: Can we please add a check that `row < n_rows`?

tket/src/Clifford/CoherentTableau.cpp
```diff
@@ -0,0 +1,666 @@
...
+  }
+  tab_.xmat_.conservativeResize(n_rows - 1, n_cols);
+  tab_.zmat_.conservativeResize(n_rows - 1, n_cols);
+  tab_.phase_.conservativeResize(n_rows - 1);
+  --tab_.n_rows_;
+}
+
+void CoherentTableau::remove_col(unsigned col) {
+  unsigned n_rows = get_n_rows();
+  unsigned n_cols = get_n_boundaries();
```
cqc-alec, line 483: Can we please add a check that `col < n_cols`?

tket/src/Converters/CoherentTableauConverters.cpp
```diff
@@ -0,0 +1,672 @@
...
+   * Gaussian elimination. The leading 1s indicate which qubits we are isolating
+   * them onto and the rest gives the CXs required to reduce them to just the
+   * leading qubits. After all post-selections have been identified, we enter
+   * the main loop of handling all other inputs and reducing them one at a time
+   * to either an identity wire or Z-decoherence (OpType::Collapse) connected to
+   * an output, or to a discard. Let in_qb be this qubit to be solved. Pick a
+   * row containing X_in_qb (if one exists). Since it must contain some
+   * non-identity component in the output segment, we can pick one of these to
+   * be out_qb and apply unitary gates at the input and output to reduce the row
+   * to X_in_qn X_out_qb. We do the same with Z (if a row exists) to necessarily
```
cqc-alec, line 60:
```suggestion
   * to X_in_qb X_out_qb. We do the same with Z (if a row exists) to necessarily
```

tket/src/Converters/CoherentTableauConverters.cpp
```diff
@@ -0,0 +1,672 @@
...
+  for (const std::pair<unsigned, unsigned>& op : col_ops) {
+    tab.tab_.apply_CX(op.first, op.second);
+    CoherentTableau::col_key_t ctrl = tab.col_index_.right.at(op.first);
+    CoherentTableau::col_key_t trgt = tab.col_index_.right.at(op.second);
+    in_circ.add_op<Qubit>(OpType::CX, {ctrl.first, trgt.first});
+  }
+
+  // Post-select rows
+  // Post-selection op is not yet available in tket - replace this once
+  // implemented
```
cqc-alec, line 153:
```suggestion
  // TODO Post-selection op is not yet available in tket - replace this once
  // implemented
```
For searchability.

tket/src/Converters/CoherentTableauConverters.cpp
```diff
@@ -0,0 +1,672 @@
...
+  //   || row.first.string.map.begin()->second != Pauli::Z) throw
+  //   std::logic_error("Unexpected error during post-selection identification
+  //   in CoherentTableau synthesis"); Qubit post_selected_qb =
+  //   row.first.string.map.begin()->first; if (row.second.coeff == -1.)
+  //   in_circ.add_op<Qubit>(OpType::X, {post_selected_qb});
+  //   tab.remove_row(tab.get_n_rows() - 1);
+  //   post_selected.insert(post_selected_qb);
+  //   throw std::logic_error("Not yet implemented: post-selection required
+  //   during CoherentTableau synthesis");
+  // }
```
cqc-alec, line 170: Assume this is left intentionally as an implementation hint?

tket/src/Converters/CoherentTableauConverters.cpp
```diff
@@ -0,0 +1,672 @@
...
+
+  // Input-first gaussian elimination to solve input-sides of remaining rows
+  tab.canonical_column_order(CoherentTableau::TableauSegment::Input);
+  tab.gaussian_form();
+
+  // Iterates through remaining inputs and reduce output portion to a single
```
cqc-alec, line 176:
```suggestion
  // Iterate through remaining inputs and reduce output portion to a single
```

tket/src/Converters/CoherentTableauConverters.cpp
```diff
@@ -0,0 +1,672 @@
...
+  // elimination subroutine to identify a combination of gates that won't add
+  // Z_out_qb onto other qubits. Since we have already removed them from all
+  // other rows containing input components, we only need to consider the
+  // output-only rows, which all exist at the bottom of the tableau.
+  unsigned out_stabs = 0;
+  while (out_stabs < tab.get_n_rows()) {
+    CoherentTableau::row_tensor_t rten =
+        tab.get_row(tab.get_n_rows() - 1 - out_stabs);
+    if (rten.first.string.map.size() != 0) {
+      // Reached the rows with non-empty inut segment
```
cqc-alec, line 430:
```suggestion
      // Reached the rows with non-empty input segment
```

cqc-alec (changes requested):
Very nice! Just a few comments.

I assume we cannot yet start using this for contextual optimization?

tket/src/Clifford/SymplecticTableau.cpp
```diff
@@ -120,6 +120,8 @@ std::ostream &operator<<(std::ostream &os, const SymplecticTableau &tab) {
 bool SymplecticTableau::operator==(const SymplecticTableau &other) const {
   bool same = this->n_rows_ == other.n_rows_;
   same &= this->n_qubits_ == other.n_qubits_;
+  // &= does not short-circuit but we need to guard the following checks
```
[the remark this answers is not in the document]
willsimmons1465: Yeah, not sure why it was originally split into multiple lines to begin with. Will leave a comment to clarify that the format is intentional as short-circuiting is required

tket/src/Clifford/include/Clifford/CoherentTableau.hpp
```diff
@@ -0,0 +1,252 @@
...
+// limitations under the License.
+
+#pragma once
+
+#include "SymplecticTableau.hpp"
+
+namespace tket {
+
+// Forward declare Circuit for friend converter
+class Circuit;
```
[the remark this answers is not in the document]
willsimmons1465: It would require making all of the internal data of the `CoherentTableau` public (i.e. the underlying `SymplecticTableau` and the map between column indices and qubits). Do we have a standard policy/guidelines wrt encapsulation and hiding?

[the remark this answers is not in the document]
cqc-alec: I don't actually see a problem with making these data public. I think we'd only need getters (const methods), so there wouldn't be any scope for client code messing with internal constraints of the class, but I may be missing something.

tket/src/Clifford/CoherentTableau.cpp
```diff
@@ -0,0 +1,666 @@
...
+        ++tab_.n_qubits_;
+      }
+      break;
+    }
+    case OpType::Collapse: {
+      collapse_qubit(qbs.at(0), seg);
+      break;
+    }
+    default: {
+      throw BadOpType(
```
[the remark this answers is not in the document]
willsimmons1465: `Phase` isn't generally Clifford. `UnitaryTableauBox` might be better waiting until we write converters between different tableau variants and then using the `compose` method (plus, the method is `apply_gate`, so I think we can get away with only supporting gate types and requiring decompositions and rebases for others). I'll add the rest though

tket/src/Converters/CoherentTableauConverters.cpp
willsimmons1465: Yep, will make this more clear in the preceding comment

willsimmons1465:
I've renamed the class in light of finding the `MixedStabilizer` class in the `QuantumClifford` Julia package that resembles what I have made here, minus the use of CJ isomorphism for handling inputs.

> I assume we cannot yet start using this for contextual optimization?

My reason for implementing this was to be able to use it for contextual optimisation within the upcoming `PauliGraph` refactor. As it currently stands, no contextual optimisations have been built using this data structure and I'm not sure how useful it would be for contextual optimisation outside of `PauliGraph` (ok, maybe for start- and end-of-circuit Clifford resynthesis, but the synthesis algorithm I wrote here is not expected to be even near optimal, so I doubt it will be much more useful than unitary resynthesis + existing peepholey contextual optimisation).

tket/src/Clifford/ChoiMixTableau.cpp
```diff
@@ -0,0 +1,728 @@
...
+      break;
+    }
+    case OpType::Y: {
+      apply_S(qbs.at(0), seg);
+      apply_S(qbs.at(0), seg);
+      apply_V(qbs.at(0), seg);
+      apply_V(qbs.at(0), seg);
+      break;
+    }
+    case OpType::Phase:
```
cqc-alec, line 239: The `Phase` operator is just a global phase; it is not the same as the `S` gate (yeah I just realized the name is confusing, sorry).

cqc-alec (changes requested):
* Any thoughts on [making the matrix getters public](https://github.com/CQCL/tket/pull/636#discussion_r1042121322)?
* Seems to be a minor formatting issue.

cqc-alec:
Checks seem to be stuck in limbo. Will close and reopen. Sometimes that helps.

cqc-alec (approved):

PR #638, opened 2022-11-04
Add Circuit method for counting the number of n-qubit gates in a Circuit

sjdilkes:
Also expose method to pytket, with additional `Circuit.n_1qb_gates()` and `Circuit.n_2qb_gates()` methods.

### DIFF
--- a/tket/src/Circuit/macro_circ_info.cpp
+++ b/tket/src/Circuit/macro_circ_info.cpp
@@ -92,6 +92,30 @@ std::list<Command> Circuit::get_commands_of_type(OpType op_type) const {
   }
   return coms;
 }
+unsigned Circuit::count_n_qubit_gates(unsigned size) const {
+  unsigned counter = 0;
+  if (size == 0) return counter;
+  BGL_FORALL_VERTICES(v, dag, DAG) {
+    if (n_in_edges_of_type(v, EdgeType::Quantum) == size) {
+      const Op_ptr op_ptr = get_Op_ptr_from_Vertex(v);
+      switch (op_ptr->get_type()) {
+        case OpType::Input:
+        case OpType::Create:
+        case OpType::Output:
+        case OpType::Discard:
+        case OpType::Reset:
+        case OpType::Measure:
+        case OpType::Barrier: {
+          break;
+        }
+        default: {
+          counter++;
+        }
+      }
+    }
+  }
+  return counter;
+}
 
 Circuit Circuit::subcircuit(const Subcircuit& sc) const {
   Circuit sub;
--- a/tket/tests/Circuit/test_Circ.cpp
+++ b/tket/tests/Circuit/test_Circ.cpp
@@ -2775,6 +2775,89 @@ SCENARIO("Named operation groups") {
   }
 }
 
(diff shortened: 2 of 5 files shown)


tket/src/Circuit/macro_circ_info.cpp
```diff
@@ -92,6 +92,30 @@ std::list<Command> Circuit::get_commands_of_type(OpType op_type) const {
...
   return coms;
 }
+unsigned Circuit::count_n_qubit_gates(unsigned size) const {
+  unsigned counter = 0;
+  if (size == 0) return counter;
+  BGL_FORALL_VERTICES(v, dag, DAG) {
+    if (n_in_edges_of_type(v, EdgeType::Quantum) == size) {
+      const Op_ptr op_ptr = get_Op_ptr_from_Vertex(v);
+      switch (op_ptr->get_type()) {
+        case OpType::Input:
```
cqc-melf, line 102: Do you need ClInput and Cloutput here as well?

tket/tests/Circuit/test_Circ.cpp
```diff
@@ -2775,6 +2775,89 @@ SCENARIO("Named operation groups") {
...
+    c.add_op<unsigned>(OpType::CnX, {0, 1, 2, 3, 4, 5, 6, 7});
+    c.add_op<unsigned>(OpType::CnX, {1, 2, 3, 4, 5, 6, 7, 8});
+    c.add_op<unsigned>(OpType::CnX, {2, 3, 4, 5, 6, 7, 8, 9});
+
+    c.add_op<unsigned>(OpType::CnX, {0, 1, 2, 3, 4, 5, 6, 7, 8});
+    c.add_op<unsigned>(OpType::CnX, {1, 2, 3, 4, 5, 6, 7, 8, 9});
+
+    c.add_op<unsigned>(OpType::CnX, {0, 1, 2, 3, 4, 5, 6, 7, 8, 9});
+    c.add_barrier({0, 1, 2, 3, 4, 5, 6, 7, 8, 9});
+    c.add_measure(Qubit(0), Bit(0));
```
cqc-melf, line 2846: Is measure a one or two qubit gat within this definition?

cqc-melf (commented):
Two short questions.

tket/src/Circuit/macro_circ_info.cpp
sjdilkes: I haven't added `CLInput` or `CLOutput` here as they should never have an in or out edge of `EdgeType::Quantum`.

tket/tests/Circuit/test_Circ.cpp
sjdilkes: It would be considered a single qubit quantum operation, as it would only have on quantum in edge. However `Measure` gates are also not counted so it should not contribute to the final count.

tket/src/Circuit/macro_circ_info.cpp
cqc-melf: Ah, yes! Great!

cqc-melf (approved):

PR #644, opened 2022-11-08
add citation file to tket repo

CalMacCQ:
Added a citation file with the tket white paper as the preferred refernce.

About CITATION.cff files -> https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,6 +1,5 @@
 cff-version: 1.2.0
-message: "If you wish to cite tket in any academic publications, we generally recommend citing our software overview paper for most cases."
-version: 1.8.1
+message: "If you wish to cite tket in any academic publications, we generally recommend citing our software overview paper for most cases. For any papers in which tket is used please also cite the version."
 url: "https://github.com/CQCL/tket"
 preferred-citation:
   type: article
@@ -11,8 +10,6 @@ preferred-citation:
     given-names: "Silas"
   - family-names: "Cowtan"
     given-names: "Alexander"
-  - family-names: "Cowtan"
-    given-names: "Alexander"
   - family-names: "Simmons"
     given-names: "Will"
   - family-names: "Edgington"
@@ -20,8 +17,8 @@ preferred-citation:
   - family-names: "Duncan"
     given-names: "Ross"
   doi: "10.1088/2058-9565/ab8e92"
-  journal: "Quantum Science & Technology"
-  month: 1
+  journal: "Quantum Science and Technology"
+  month: 11
   title: "TKET: A Retargetable Compiler for NISQ devices"
   volume: 6
-  year: 2021
+  year: 2020
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,27 @@
+cff-version: 1.2.0
+message: "If you use this software, please cite it as below."
+version: 1.8.1
+url: "https://github.com/CQCL/tket"
+preferred-citation:
+  type: article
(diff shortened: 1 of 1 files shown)


CITATION.cff
```diff
@@ -0,0 +1,27 @@
+cff-version: 1.2.0
+message: "If you use this software, please cite it as below."
+version: 1.8.1
```
cqc-alec: I don't think we shou;d hard-code the version number here, it would have to be updated with every release. Just add something to the `message` saying that the version used should be included in the citation?

CITATION.cff
```diff
@@ -0,0 +1,27 @@
...
+  type: article
+  authors:
+  - family-names: "Sivirajah"
+    given-names: "Seyon"
+  - family-names: "Dilkes"
+    given-names: "Silas"
+  - family-names: "Cowtan"
+    given-names: "Alexander"
+  - family-names: "Cowtan"
+    given-names: "Alexander"
```
cqc-alec, line 15: Why twice?

CITATION.cff
```diff
@@ -0,0 +1,27 @@
...
+  - family-names: "Simmons"
+    given-names: "Will"
+  - family-names: "Edgington"
+    given-names: "Alec"
+  - family-names: "Duncan"
+    given-names: "Ross"
+  doi: "10.1088/2058-9565/ab8e92"
+  journal: "Quantum Science & Technology"
```
cqc-alec, line 23:
```suggestion
  journal: "Quantum Science and Technology"
```

CITATION.cff
```diff
@@ -0,0 +1,27 @@
...
+  month: 1
```
cqc-alec, line 24:
```suggestion
  month: 1
```
I think this should be "number" (if that is the correct tag to use)? The actual month was November.

CITATION.cff
```diff
@@ -0,0 +1,27 @@
...
+  title: "TKET: A Retargetable Compiler for NISQ devices"
+  volume: 6
+  year: 2021
```
cqc-alec, line 27: According to https://iopscience.iop.org/article/10.1088/2058-9565/ab8e92 it was November 2020, where did you get this from?

cqc-alec (changes requested):

CITATION.cff
CalMacCQ: I think I got it from here -> https://ui.adsabs.harvard.edu/abs/2021QS%26T....6a4003S/abstract

You're right, I should've checked with the IOP directly.

CITATION.cff
CalMacCQ: Good idea. Thanks.

CITATION.cff
CalMacCQ: Oops, fixed now.

CITATION.cff
```diff
@@ -0,0 +1,24 @@
+cff-version: 1.2.0
+message: "If you wish to cite tket in any academic publications, we generally recommend citing our software overview paper for most cases. For any papers in which tket is used please also cite the version."
```
cqc-alec, line 2:
```suggestion
message: "If you wish to cite TKET in any academic publications, we generally recommend citing our software overview paper. For any papers in which TKET is used please also cite the version."
```

cqc-alec (changes requested):

cqc-alec (approved):

PR #645, opened 2022-11-08
Correct handling of conditional phases

### LINKED ISSUE #576: `squash_1qb_to_tk1` does not preserve conditional phase
The following code

```
Circuit circ(1, 1);
circ.add_conditional_gate<unsigned>(OpType::Rz, {2.}, {0}, {0}, 1);
std::cout << circ;
Transforms::squash_1qb_to_pqp(OpType::Rz, OpType::Ry).apply(circ);
std::cout << circ;
```

prints an empty circuit with phase 1.0, ie:
```
Phase (in half-turns): 1.0
```

expected:
```
IF ([c[0]] == 1) THEN Rz(2) q[0];
Phase (in half-turns): 0.0
```

or equivalent.

cqc-alec:
Fixes #576 .
After merging I plan to make a dev release and then update all the extensions to make sure they work with this change,

### DIFF
--- a/tket/src/Transformations/Rebase.cpp
+++ b/tket/src/Transformations/Rebase.cpp
@@ -59,7 +59,7 @@ static bool standard_rebase(
         tk1_replacement) {
   bool success = false;
   VertexSet bin;
-  for (const Vertex& v : circ.vertices_in_order()) {
+  for (const Vertex& v : circ.all_vertices()) {
     Op_ptr op = circ.get_Op_ptr_from_Vertex(v);
     unsigned n_qubits = circ.n_in_edges_of_type(v, EdgeType::Quantum);
     if (n_qubits <= 1) continue;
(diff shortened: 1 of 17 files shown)


tket/src/Transformations/Rebase.cpp
```diff
@@ -59,7 +59,7 @@ static bool standard_rebase(
         tk1_replacement) {
   bool success = false;
   VertexSet bin;
-  BGL_FORALL_VERTICES(v, circ.dag, DAG) {
+  for (const Vertex& v : circ.vertices_in_order()) {
```
yao-cqc, line 62: Why do we need these changes?

cqc-alec: Without this change, a rebase pass operating on a circuit with a conditional operation enters an infinite loop, as the vertices in the DAG are modified inside the iteration. I don't quite get why we didn't see this before, maybe just luck...?

yao-cqc: That's interesting... anyway, the code looks safer now. However, do we really need a sorted list?

cqc-alec: Done.

cqc-alec:
> Should we also change the QASM converter to handle `OpType.Phase`s?

Yes if it doesn't already then we should. Can we do that in another PR?

yao-cqc (approved):

lmondada (commented):
Nice work, thank you!

PR #649, opened 2022-11-09
nb viewer link

CalMacCQ:
I created an examples.html documentation page with the links to the various examples which can be viewed through the nbviewer.

This will allow the notebooks to be downloaded but the notebooks themselves will not be executed (this requires binder which takes ages to build for the pytket examples).

This is a small change but seems minimally invasive for now and will not require conversion to .rst files or refactoring of our examples repository.

There is a download option available but it seems that you have to right click on the download button and click "save as" for it to work properly. I haven't figured out why this is the case but it seems to be addressed in this issue.

https://github.com/jupyter/nbviewer/issues/756

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,8 +16,7 @@ selection of example notebooks, please follow the links from the
 [pytket](https://github.com/CQCL/pytket) main page.
 
 Note that the various pytket extensions (which allow pytket to interface with
-other software packages and with quantum devices) live in the separate
-[pytket-extensions](https://github.com/CQCL/pytket-extensions) repository.
+other software packages and with quantum devices) live in the separate github repositories. A list of extensions can be found in [this page](https://cqcl.github.io/pytket-extensions/api/index.html).
 
 If you would like to build tket yourself and help to improve it, read on!
 
--- a/pytket/docs/examples.rst
+++ b/pytket/docs/examples.rst
@@ -1,0 +1,51 @@
+Example notebooks
+=================
+
+Here you can find notebooks showing features and example usage of pytket.
+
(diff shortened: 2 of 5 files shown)


cqc-melf:
Have you checked if this is free for us to use and the are not limits in the traffic on that page?

cqc-melf:
If you want to recommend this service we should update the link in the manual and the documentation, too.

This currently removes the information from the readme which is having some more information in there which is not shown on this link. Do you have an idea how to included that?

CalMacCQ:
The nbviewer service is offered for free. I am not sure about limits on traffic.

https://nbviewer.org/faq#where-is-nbviewerjupyterorg-hosted

CalMacCQ:
Good point about the README will have a look and see if that can be included.

CalMacCQ:
I have looked around and have not found a way to make the README to display on the nbviewer. Seems there have been a few issues for this.

https://github.com/jupyter/nbviewer/issues/498
https://github.com/github/markup/issues/929

I guess the main things that would be missing would be the "notes for developers section" and the descriptions of the examples. The binder link is available in nbviewer.

cqc-melf:
Maybe we can solve this by adding this new link in the readme? Or do you thing that we are not loosing something important? I think pointing to the repo and just adding the new link to the readme would be better to get the hole information to the users and all an easy download for them as well?

What do you think?

CalMacCQ:
After thinking about this for a while I am not sure this nbviewer change offers much benefit over what we have already. Should we link directly to the example notebook files in github from the links instead of the nbviewer?

cqc-melf:
> After thinking about this for a while I am not sure this nbviewer change offers much benefit over what we have already. Should we link directly to the example notebook files in github from the links instead of the nbviewer?

You you mean to link to the directory or to the individual files?

PR #652, opened 2022-11-10
Run doctests

cqc-alec:
Although we already pass the `--doctest-modules` to pytest on the CI, no doctests are run because it is executed from the `tests` directory. We have to execute from there in order to find certain files used by the tests, so (rather than modify a lot of test code), I just added a separate step to run the doctests from the `pytket` directory.

In fact, there were no existing doctests. I added one (to show the idea). I think we should add a lot more of these. Not only do they test the code, but they improve the API documentation by giving short examples showing how the methods and classes are used:
Screenshot from 2022-11-10 15-18-39

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -200,18 +200,23 @@ jobs:
       run: |
         cd pytket
         pip install -e . -v
+    - name: Run doctests
+      run: |
+        cd pytket
+        pip install pytest
+        pytest --doctest-modules pytket
     - name: Test pytket with coverage
       if: matrix.os == 'ubuntu-22.04' && matrix.pyver == '3.8'
       run: |
         cd pytket/tests
         pip install -r requirements.txt
-        pytest --ignore=simulator/ --doctest-modules --hypothesis-seed=1 --cov=../pytket --cov-branch --cov-report=html --cov-report=xml:htmlcov/cov.xml
+        pytest --ignore=simulator/ --hypothesis-seed=1 --cov=../pytket --cov-branch --cov-report=html --cov-report=xml:htmlcov/cov.xml
     - name: Test pytket
       if: matrix.os != 'ubuntu-22.04' || matrix.pyver != '3.8'
       run: |
         cd pytket/tests
         pip install -r requirements.txt
-        pytest --ignore=simulator/ --doctest-modules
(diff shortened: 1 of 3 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -200,18 +200,23 @@ jobs:
       run: |
         cd pytket
         pip install -e . -v
+    - name: Run doctests
+      run: |
+        cd pytket
+        pip install pytest
+        pytest --doctest-modules pytket
```
CalMacCQ, line 207: Can these doctests be run locally using this command?

`pytest --doctest-modules pytket`

.github/workflows/build_and_test.yml
```diff
@@ -200,18 +200,23 @@ jobs:
...
     - name: Test pytket with coverage
       if: matrix.os == 'ubuntu-22.04' && matrix.pyver == '3.8'
       run: |
         cd pytket/tests
         pip install -r requirements.txt
-        pytest --ignore=simulator/ --doctest-modules --hypothesis-seed=1 --cov=../pytket --cov-branch --cov-report=html --cov-report=xml:htmlcov/cov.xml
```
CalMacCQ, line 208: Were there any of these doctests in the pytket/tests directory before? Or was this --doctest-modules redundant before this PR?

CalMacCQ:
Good idea. I think we should definitely do this more in the documentation.

Changes look good, happy to approve. Just had a couple questions.

One additional question... Could make these tests for the pytket.passes module?  From my understanding the docstrings are in the C++ source.

.github/workflows/build_and_test.yml
cqc-alec: Yes, from the `pytket` directory that works.

.github/workflows/build_and_test.yml
cqc-alec: No, it was doing nothing. I'm not sure if it ever worked; I suspect not; in any case there were no doctests for it to test.

CalMacCQ (approved):

cqc-alec:
> One additional question... Could make these tests for the pytket.passes module? Where are the docstrings for pytket.passes?

This command should run it for all modules defined under `pytket`, including `pytket.passes`. However, I am not sure it works for docstrings in the binders, I will look into this.

PR #654, opened 2022-11-17
[debug] add wasm wire

cqc-melf:


cqc-melf:
This PR is only for discussions and questions about the additional wasm wire.

tket/src/Circuit/basic_circ_manip.cpp
```diff
@@ -459,6 +466,18 @@ register_t Circuit::add_c_register(std::string reg_name, unsigned size) {
...
 }
 
+void Circuit::add_wasm_register() {
+  if (!wasm_added) {
+    Vertex in = add_vertex(OpType::WASMInput);
+    Vertex out = add_vertex(OpType::WASMOutput);
+    add_edge({in, 0}, {out, 0}, EdgeType::WASM);
+    WASMUID wuid = WASMUID();
+    wasmwire = wuid;
+    boundary.insert({wasmwire, in, out});
```
cqc-melf, line 476: This this the right way to add new elements to the boundary? Is there something I am missing?

tket/src/Circuit/include/Circuit/Circuit.hpp
```diff
@@ -1742,6 +1758,19 @@ Vertex Circuit::add_op(
...
+    if (sig[args.size()] == EdgeType::WASM) {
+      Vertex out_vert = get_out(wasmwire);
+      Edge pred_out_e = get_nth_in_edge(out_vert, 0);
+      preds.push_back(pred_out_e);
+    } else {
+      throw CircuitInvalidity(
+          "The last element of the signature should be wasm for OpType::WASM");
+    }
+  }
   rewire(new_v, preds, sig);
```
cqc-melf, line 1774: Is there any additional information required for the rewire?

tket/src/Circuit/include/Circuit/Circuit.hpp
```diff
@@ -1742,6 +1758,19 @@ Vertex Circuit::add_op(
     Edge pred_out_e = get_nth_in_edge(out_vert, 0);
     preds.push_back(pred_out_e);
   }
+  if (op->get_type() == OpType::WASM) {
+    if (!wasm_added) {
+      add_wasm_register();
+    }
+    if (sig[args.size()] == EdgeType::WASM) {
+      Vertex out_vert = get_out(wasmwire);
```
cqc-melf, line 1766: This should always return the wasm out vertex from the boundary?

tket/src/Circuit/macro_circ_info.cpp
```diff
@@ -647,15 +647,27 @@ SliceVec Circuit::get_reverse_slices() const {
...
   std::function<bool(Op_ptr)> skip_func = [&](Op_ptr op) {
     return (op->get_type() == OpType::Barrier);
   };
+  std::cout << "Try to get depths - 2\n";
   Circuit::SliceIterator slice_iter(*this, skip_func);
+  std::cout << "Try to get depths - 3\n";
   if (!(*slice_iter).empty()) count++;
+  std::cout << "Try to get depths - 4\n";
+  unsigned i = 0;
   while (!slice_iter.finished()) {
```
cqc-melf, line 660: Is there is at least one wasm op with the new wasm wire in the circuit this loop will never stop.

tket/tests/test_wasm.cpp
```diff
@@ -107,6 +107,30 @@ SCENARIO("generating circ with wasm") {
...
+  GIVEN("wasmop add circuit III") {
+    Circuit u(6, 6);
+
+    const std::shared_ptr<WASMOp> wop_ptr =
+        std::make_shared<WASMOp>(1, uv_2, uv_3, wasm_func, wasm_file);
+
+    u.add_op<unsigned>(wop_ptr, {0});
+
+    std::cout << "try to get depth\n\n";
+    REQUIRE(u.depth() == 0);
```
cqc-melf, line 131: This is currently the only testcase which hits the loop issue.

tket/src/Circuit/macro_circ_info.cpp
```diff
@@ -647,15 +647,27 @@ SliceVec Circuit::get_reverse_slices() const {
...
+    i++;
+    std::cout << "while loop - ";
+    std::cout << i << std::endl;
     slice_iter.cut_ = this->next_cut(
```
cqc-melf, line 664: My current assumption is that this next_cut generation somehow has a special handling of the edge types somewhere, but I was not able to find it.

cqc-alec:
There are some places in the code, e.g. `basic_circ_manip.cpp` l. 100, where it is implicitly assumed that `EdgeType::Quantum` and `EdgeType::Classical` are the only two edge types. I don't know if that particular one is a problem but It would be worth checking for such cases if you have not already.

cqc-melf:
> There are some places in the code, e.g. `basic_circ_manip.cpp` l. 100, where it is implicitly assumed that `EdgeType::Quantum` and `EdgeType::Classical` are the only two edge types. I don't know if that particular one is a problem but It would be worth checking for such cases if you have not already.

Thank you for taking a look!
I have taken a look on a few of the this cases like in the check of the "out" types and updated them, but it might be the right assumption that I have missed some of them.
Probably the right place for further investigations. I think I will try to take a look on all of them again systematically to make sure there is nothing I have missed!

I have taken again a look on different dag constellations with wasm and tried several checks, but I have not found anything suspicious which might have caused any issue. All vertices are connected as intended and the signature type match the type of the edges. Which supports that there is somethings going wrong outside of the construction of the circuit and the new wasm register.

cqc-alec:
Probably unrelated to this issue, but have you considered giving _all_ circuits a single WASM wire at construction (with a unique ID)? This might simplify things a bit, e.g. you wouldn't need the `add_wasm_register()` method.

cqc-alec:
Another comment: I wonder if we should make the WASM UnitID an explicit argument passed to `add_op()` in the C++. We could still "hide" it in the Python layer.

Thinking some more about this and how we would want to generalize it in the future to allow more refined dependencies between WASM ops, I think we may want to make it an explicit argument (so that there can be several WASM wires each holding some "state"). We could keep the existing API for backwards compatibility: if you don't specify a WASM wire you get the default one.

cqc-melf:
> Probably unrelated to this issue, but have you considered giving _all_ circuits a single WASM wire at construction (with a unique ID)? This might simplify things a bit, e.g. you wouldn't need the `add_wasm_register()` method.

The idea with the `add_wasm_register()` method was that we don't have the wasm wire shown in circuits that are not having wasm in there, because it might confuse people. Ideally I wanted the wasm wire to be plotted by the DAG to make people aware of this if they use wasm. The add wasm function is only executed once, if one wasm wire exists there is nothing changed when running the method.

cqc-melf:
> Probably unrelated to this issue, but have you considered giving _all_ circuits a single WASM wire at construction (with a unique ID)? This might simplify things a bit, e.g. you wouldn't need the `add_wasm_register()` method.

I have thought about this quiet long. My idea was to add this in the feature if we maybe want to have more then the one wasmuid to allow the user more flexibility in the order of the wasm operation, but only to add this when we have all other things working. This might also a little bit more complicated because we would need to stop using add_op<unsigned> and add_op<Bit> for the wasm functions and always use add_op<UnitID> for all wasm. Or that we would need to add a new additional parameter only taking the wasm UID which would be a bigger change as well and probably less flexible for expansions.

cqc-alec:
The method `Circuit::assert_valid()` may be useful for debugging this. But first you should update the method `is_valid()` defined in `DAGProperties.cpp` since currently it will fail for circuits containing WASM ops.

cqc-melf:
> The method `Circuit::assert_valid()` may be useful for debugging this. But first you should update the method `is_valid()` defined in `DAGProperties.cpp` since currently it will fail for circuits containing WASM ops.

Thank you! I will take a look!

PR #660, opened 2022-11-22
Make `optimisation_level` default to 2

cqc-alec:
This will affect the default behaviour of `get_compiled_circuit()` in all backends. It will _not_ affect the default behaviour of `default_optimisation_pass()` since that is an abstract method which is reimplemented in the backends: changing that will require changes to the individual backends.


sjdilkes (approved):
Nice!

PR #661, opened 2022-11-23
Update `MappingFrontier::add_swap` to only edit struct attributes after any break conditions are passed

sjdilkes:
Basically as the name says. `MappingFrontier::add_swap` returns `false` if the vertex immediately before the proposed SWAP is an identical SWAP, thus avoiding getting stuck.
If the `false` condition is met, then the proposed SWAP is not implemented, but instead a path of SWAPs is found between one of the SWAPs in the `Node` and its target `Node` for being brought adjacent to.

During `MappingFrontier::add_swap`, if one of the `Node` "A" in the proposed SWAP between "A" and "B" is a known "ancilla node", then this `Node` "A" is removed from the "ancilla nodes container" and the `Node` "B" is added. Previously, this was incorrectly being updated before the condition for returning `false` was met. This left the "ancilla nodes container" incorrect.

A container of ancilla nodes is stored incase we choose to assign an unplaced `Circuit` `Qubit` to an `Architecture` `Node` that is being used as an ancilla. In this case, we merge the paths of the `UnitID` in the DAG, such that the (unitary preserving) operations of the ancilla node are run, before the path of previously unplaced `Qubit`. This hits no causal issues because we iterate through the `Circuit` slice by slice.

However, in this case a `Node` that was assumed "fair game" in the ancilla nodes container was actually that of an important program `Qubit`, due to the returning `false` scenario described previously. When merging paths this led to a cycle in the `Circuit`.

This PR guarantees that the check for breaking from the `MappingFrontier::add_swap` is completed and any break completed before any struct attributes are amended.

### DIFF
--- a/pytket/conanfile.txt
+++ b/pytket/conanfile.txt
@@ -1,5 +1,5 @@
 [requires]
-tket/1.0.30@tket/stable
+tket/1.0.31@tket/stable
 tklog/0.1.2@tket/stable
 pybind11/2.10.1
 nlohmann_json/3.11.2
--- a/pytket/docs/changelog.rst
+++ b/pytket/docs/changelog.rst
@@ -11,6 +11,8 @@ Fixes:
   relies on the circuit having gates only in the specified gate set should be
   updated to handle ``OpType.Phase`` as well when conditional operations are
   present.
+* A bug where the sequence of ``RoutingMethod`` used in ``DefaultMappingPass`` could 
+  add a cycle to the ``Circuit`` DAG has been fixed.
 
 API changes:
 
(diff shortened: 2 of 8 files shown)


yao-cqc:
Should we mention this in the changelog?

sjdilkes:
> Should we mention this in the changelog?

I will attempt 

> Should we mention this in the changelog?

I'm actually not sure, as from a user perspective the situation has gone from "Routing has a bug which could introduce cycles which led to a RuntimeError" to "Routing now has no known bugs which could introduce cycles which led to a RuntimeError" but in terms of interfacing it's the same for them. 
Would:
"A bug where the sequence of ``RoutingMethod`` used in ``DefaultMappingPass`` could add a cycle to the ``Circuit`` has been fixed."

yao-cqc (approved):
Thanks! I agree with the note you've added.

PR #663, opened 2022-11-25
Release/1.9.0

cqc-alec:


pytket/pytket/qasm/qasm.py
```diff
@@ -160,6 +160,7 @@ class QASMUnsupportedError(Exception):
     "bridge": OpType.BRIDGE,
     "iswapmax": OpType.ISWAPMax,
     "zzmax": OpType.ZZMax,
+    "ecr": OpType.ECR,
```
cqc-melf, line 163: The last commit from Silas is not updating the changelog? Should that be added?

cqc-alec: Done.

cqc-melf (approved):

PR #670, opened 2022-11-30
add function for opening a browser with circuit render

ss2165:


pytket/pytket/circuit/display/__init__.py
```diff
@@ -73,3 +74,20 @@ def render_circuit_jupyter(
     :param circuit: the circuit to render.
     """
     render_circuit_as_html(circuit, True)
+
+
+def view_browser(circuit: Union[Dict[str, Union[str, float, dict]], Circuit]) -> None:
```
cqc-alec, line 79: Could we make the `new` parameter to `webbroser.open()`, and the time passed to `sleep()`, into optional arguments (with the defaults you have)?

pytket/pytket/circuit/display/__init__.py
```diff
@@ -73,3 +74,20 @@ def render_circuit_jupyter(
...
+    fp = tempfile.NamedTemporaryFile(mode="w", suffix=".html", delete=False)
+    fp.write(cast(str, render_circuit_as_html(circuit)))
+    fp.flush()
+    fp.seek(0)
+    fp.close()
+    webbrowser.open(fp.name, new=2)
+
+    # give browser enough time to open before deleting file
+    time.sleep(5)
+    os.remove(fp.name)
```
cqc-alec, line 93: Can we use a `try` block to make this execute always even if an error is thrown above?

pytket/pytket/circuit/display/__init__.py
cqc-alec, line 79: I see that `Union[str, float, dict]]` is used in a few places; maybe worth making a descriptive alias for it.

cqc-alec (changes requested):

pytket/pytket/circuit/display/__init__.py
```diff
@@ -66,10 +69,34 @@ def render_circuit_as_html(
...
     """
     render_circuit_as_html(circuit, True)
+
+
+def view_browser(circuit: RenderCircuit, browser_new: int = 2, sleep: int = 5) -> None:
+    """Write circuit render html to a tempfile and open in browser.
+
+    Waits for some time for browser to load then deletes tempfile.
+
+    :param circuit: the circuit to render.
```
cqc-alec, line 86: Here we should describe the type (either a `Circuit` or (I assume) a serialized `Circuit`).

cqc-alec (dismissed):

cqc-alec (approved):

PR #673, opened 2022-12-05
Warn about rounding errors instead of asserting

### LINKED ISSUE #672: Critical Assertion in replace_TK2_2CX fails
When applying `FullPeepholeOptimise` to a quantum circuit with a quantum phase estimation part using a large number of measurement qubits (and therefore very small phase rotations), a critical assertion in `replace_TK2_2CX` ([CircUtils.cpp](https://github.com/CQCL/tket/blob/develop/tket/src/Circuit/CircUtils.cpp) in line 238) fails:

`[2022-12-01 14:34:12] [tket] [critical] Assertion 'equiv_0(params[2], 4, 1e-6)' (/Users/runner/.conan/data/tket/1.0.31/tket/stable/build/<...>/Circuit/CircUtils.cpp : replace_TK2_2CX : 238) failed.  Aborting.
`

I am using the FullPeepholeOptimise in a script to generate many quantum circuits and in that case the resulting `std::abort()` from the failed assertion is quite unfortunate, because I cannot treat/catch it.

The following code can be used to reproduce the result (python v3.9, pytket v1.9.0):
``` python

from qiskit.circuit.library import PhaseEstimation, PhaseGate
import numpy as np
from qiskit import QuantumCircuit, transpile
from pytket.extensions.qiskit import qiskit_to_tk
from pytket.passes import FullPeepholeOptimise, auto_rebase_pass
from pytket import OpType

u = PhaseGate(np.pi)
qc = PhaseEstimation(num_evaluation_qubits=30, unitary=u)
openqasm_gates = ["u3", "u2", "u1", "cx", "id", "u0", "u", "p", "x", "y", "z", "h", "s", "sdg", "t", "tdg", "rx", "ry", "rz", "sx", "sxdg", "cz", "cy", "swap", "ch", "ccx", "cswap", "crx", "cry", "crz", "cu1", "cp", "cu3", "csx", "cu",  "rxx", "rzz", "rccx", "rc3x", "c3x", "c3sqrtx", "c4x" ]
qc = transpile(qc, basis_gates=openqasm_gates)
qc_tket = qiskit_to_tk(qc)

ionq_rebase = auto_rebase_pass({OpType.Rz, OpType.Ry, OpType.Rx, OpType.XXPhase})
ionq_rebase.apply(qc_tket)

FullPeepholeOptimise().apply(qc_tket)
```

I converted the qiskit quantum circuit to contain only openQASM 2.0 gates such that any custom gates are avoided when `qiskit_to_tk` is applied and transpile it to the IonQ native gate set using pytket. If I skip the `rebase` step, the assert does not fail.

Maybe a higher tolerance within the `replace_TK2_2CX` assertion statement could solve the problem?

cqc-alec:
Fixes #672 .

Since this reduces branch coverage (I am not sure if I can easily reproduce it in a test), I have added a couple of unrelated edge-case tests to bring up the coverage numbers.

### DIFF
--- a/pytket/conanfile.txt
+++ b/pytket/conanfile.txt
@@ -1,5 +1,5 @@
 [requires]
-tket/1.0.31@tket/stable
+tket/1.0.32@tket/stable
 tklog/0.1.2@tket/stable
 pybind11/2.10.1
 nlohmann_json/3.11.2
--- a/pytket/docs/changelog.rst
+++ b/pytket/docs/changelog.rst
@@ -8,7 +8,10 @@ Minor new features:
 
 * New ``view_browser`` function for opening a browser with circuit render.
 
+Fixes:
 
+* Warn rather than abort when significant rounding errors are detected in
+  TK2-to-CX rebase.
 
 1.9.0 (November 2022)
 ---------------------
--- a/recipes/tket-proptests/conanfile.py
+++ b/recipes/tket-proptests/conanfile.py
@@ -28,7 +28,7 @@ class TketProptestsConan(ConanFile):
     generators = "cmake"
     exports_sources = "../../tket/proptests/*"
     requires = (
-        "tket/1.0.31@tket/stable",
+        "tket/1.0.32@tket/stable",
         "rapidcheck/cci.20220514",
     )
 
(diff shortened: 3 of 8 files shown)


sjdilkes (approved):

sjdilkes (changes requested):
Do you think it's worth mentioning this in the changelog as it's both a commonly used pass and a user hit the assertion? Happy for it to not be added if you don't think it's worthwhile

sjdilkes (approved):

PR #679, opened 2022-12-08
Update rzz qasm string in qasm converter

sjdilkes:
hqs qasm header defines OpType.ZZPhase gate as "RZZ", but in converter code it's been mistakenly set to "rzz" (which is how pytket defines it as a string). This gives examples where "RZZ" is not recognised as an OpType.ZZPhase gate and so is instead read in as a custom gate.

### DIFF
--- a/pytket/docs/changelog.rst
+++ b/pytket/docs/changelog.rst
@@ -12,8 +12,8 @@ Fixes:
 
 * Warn rather than abort when significant rounding errors are detected in
   TK2-to-CX rebase.
-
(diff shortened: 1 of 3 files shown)


cqc-alec:
Could we also add a new test case (as in the bug report) please?

cqc-alec:
And a changelog entry.

sjdilkes:
> Could we also add a new test case (as in the bug report) please?

On it!

cqc-alec (dismissed):

cqc-alec (approved):

PR #690, opened 2022-12-15
Stricter rules on conditional expressions being converted to QASM

cqc-alec:
This fixes two bugs:

* converting conditions defined on a strict subset of bits of a register (of a size larger than 1) to conditions on the whole register;
* ignoring the order of bits in conditions on a whole register.

In both cases, it should be possible to do the translation by being more clever. However, disallowing the conversions is better than converting them erroneously -- and matches what we were doing already in other similar situations.

### DIFF
--- a/pytket/pytket/qasm/qasm.py
+++ b/pytket/pytket/qasm/qasm.py
@@ -1144,8 +1144,8 @@ def circuit_to_qasm_io(
                     variable = control_bit.reg_name
                     if hqs_header(header) and bits != list(cregs[variable]):
                         raise QASMUnsupportedError(
-                            "HQS QASM conditions must be an entire classical register "
-                            "or a single bit"
+                            "hqslib1 QASM conditions must be an entire classical "
+                            "register or a single bit"
                         )
             if not hqs_header(header):
                 if op.width != cregs[variable].size:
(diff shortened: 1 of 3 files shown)


pytket/pytket/qasm/qasm.py
```diff
@@ -1142,12 +1142,17 @@ def circuit_to_qasm_io(
                     variable = control_bit
                 else:
                     variable = control_bit.reg_name
+                    if hqs_header(header) and bits != list(cregs[variable]):
+                        raise QASMUnsupportedError(
+                            "HQS QASM conditions must be an entire classical register "
```
sjdilkes, line 1147: Most pedantic comment over, but maybe `hqs_header QASM conditions...` is more precise these days

sjdilkes: Is this constraint specified in the hqs_header though? Or is it a hardware restriction that is bounced back via the Quantinuum API or something?

sjdilkes (changes requested):
Only a pedantic comment regarding the error message thrown. Apart from that looks good!

pytket/pytket/qasm/qasm.py
cqc-alec: I'm actually not sure (I've asked). If it is allowed then we need to do something more clever here -- but meanwhile, better to throw an error than do the _wrong_ thing.

cqc-alec: > Most pedantic comment over, but maybe `hqs_header QASM conditions...` is more precise these days

Fixed.

sjdilkes (approved):

PR #696, opened 2022-12-21
Update `MappingManager` to handle Vertices and Boundaries with Boolean Bundles properly

sjdilkes:
A Circuit can have a `ClInput` vertex with multiple out wires: a bundle of EdgeType::Boolean and a EdgeType::Classical edge. This PR updates how a `MappingFrontier` object is constructed to handle this case.

Additionally it refactors `MappingFrontier::advance_frontier_boundary`.


tket/src/Mapping/MappingFrontier.cpp
```diff
@@ -257,160 +260,196 @@ void MappingFrontier::advance_frontier_boundary(
...
+                this->linear_boundary, {this->circuit_.source(edge),
+                                        this->circuit_.get_source_port(edge)});
+            break;
+          }
+          case EdgeType::Boolean: {
+            // N.B. function will find uid even if the bundle has more than the
+            // single "edge" passed
+            uid = get_bit_from_bool_frontier(this->boolean_boundary, {edge});
+            break;
+          }
```
cqc-alec, line 315: Worth having a `dafault` case with a `TKET_ASSERT` that aborts. It's likely that a new edge type (WASM) will be added in the future and we want to make sure we catch this.

tket/src/Mapping/MappingFrontier.cpp
```diff
@@ -257,160 +260,196 @@ void MappingFrontier::advance_frontier_boundary(
...
+              Edge replacement_edge =
+                  next_cut.u_frontier->get<TagKey>().find(uid.first)->second;
+              this->linear_boundary->replace(
+                  this->linear_boundary->get<TagKey>().find(uid.first),
+                  {uid.first,
+                   {this->circuit_.source(replacement_edge),
+                    this->circuit_.get_source_port(replacement_edge)}});
+              break;
+            }
           }
```
cqc-alec, line 352: Again, worth having a `default` case with a `TKET_ASSERT`.

tket/src/Mapping/MappingFrontier.cpp
```diff
@@ -257,160 +260,196 @@ void MappingFrontier::advance_frontier_boundary(
...
+                // => can erase Bit from boolean boundary as its not longer used
+                this->boolean_boundary->erase(boolean_it);
+              } else {
+                // => either Vertex has edges in output boolean bundle
+                // => or there are other edges in boolean bundle held in boolean
+                // boundary that are not attached to this vertex
+                // => update boolean boundary
+                this->boolean_boundary->replace(boolean_it, {bit, out_bundle});
+              }
+              break;
```
cqc-alec, line 441: Add `default` case with `TKET_ASSERT`.

cqc-alec (changes requested):

tket/src/Mapping/MappingFrontier.cpp
sjdilkes: Oops, I thought I did these locally but it turns out my local file changes weren't being tracked by github for some reason. All done now.

tket/src/Mapping/MappingFrontier.cpp
sjdilkes: Done

tket/src/Mapping/MappingFrontier.cpp
sjdilkes: Done

PR #701, opened 2022-12-30
Feature/easier builds and integration with ides

trvto:
- Added top level CMakeLists.txt in `dev-tools` directory (just includes others) for better integration with IDEs (Tested with Clion)
- Added Makefile with commands for streamlined development, see `dev-tools/README.md`
- Changed name of variable used for conanbuild.info file location in subdirectory CMakeLists files to avoid clashes in top level CMakeLists (regular `conan create` builds still work)
- Passed test circuit directory to tket-tests as Macro. This way tests don't have to be started from binary directory


cqc-alec:
Nice! The M1 failure looks like a known ephemeral issue; not sure about the docs build failure -- just trying a re-run.

cqc-alec:
The docs building issue should be fixed with a merge from `develop`.

trvto:
Okay, thanks for having a look Alec! I've merged in the develop branch. I'm on holiday this week so if anything more comes up I won't be able to get to it until Monday

Makefile
```diff
@@ -0,0 +1,95 @@
...
+
+##@ Install Dependencies
+
+.PHONY: pybind
+pybind: ## install custom pybind package
+	conan profile update settings.build_type=$(build_type) $(conan_profile_name)
+	conan remove -f "pybind11/*"
+	conan create --profile=$(conan_profile_name) recipes/pybind11
+
+ # Currently the b2 package (a dependency of boost) in the remote tket-libs does not include compiler info
```
cqc-alec, line 39: This may be fixed now? I rebuilt it a couple of days ago with the latest recipe from the conan-center.

Makefile
```diff
@@ -0,0 +1,95 @@
...
+	conan install --profile=$(conan_profile_name) recipes/tket-tests --install-folder=$(conan_build_dir)/tket/tests
+	conan install --profile=$(conan_profile_name) recipes/tket-proptests --install-folder=$(conan_build_dir)/tket/proptests
+	cmake -DCMAKE_BUILD_TYPE=$(build_type) -DCMAKE_MAKE_PROGRAM=ninja -G Ninja -S . -B $(conan_build_dir)
+
+##@ Build
+
+.PHONY: build
+build: ## build with cmake
+	cmake --build $(conan_build_dir) --target test_tket -j 6
+	cmake --build $(conan_build_dir) --target proptest -j 6
```
cqc-alec, line 78: Instead of "6" we could make this configable, defaulting to `nproc`?

pytket/CMakeLists.txt
```diff
@@ -49,6 +49,11 @@ if (UNIX)
     set(CMAKE_INSTALL_RPATH "\${ORIGIN}")
 endif()
 
+if(CMAKE_CXX_COMPILER_ID MATCHES "(Apple)?Clang")
```
cqc-alec, line 52: This change is now on `develop`.

tket/proptests/CMakeLists.txt
```diff
@@ -26,15 +26,19 @@ endif()
...
+find_file(CONANBUILDINFO_FILE_TKET_PROPTESTS conanbuildinfo.cmake HINTS ${CMAKE_CURRENT_BINARY_DIR})
+include(${CONANBUILDINFO_FILE_TKET_PROPTESTS})
 conan_basic_setup()
 
 IF (WIN32)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /WX /EHsc")
 ELSE()
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Werror -Wunreachable-code -Wunused")
 ENDIF()
+if(CMAKE_CXX_COMPILER_ID MATCHES "(Apple)?Clang")
```
cqc-alec, line 38: This change is now on `develop`.

pytket/CMakeLists.txt
trvto: Removed from here

tket/proptests/CMakeLists.txt
trvto: removed from here

Makefile
trvto: I configured `n_cpus` using `python -c 'import multiprocessing as m; print(m.cpu_count())'`, which I think should work as long as python is installed, also on mac and windows

Makefile
trvto: Problem persists so I'll leave it for now

cqc-alec (commented):
I am in two minds. I see the convenience, but having a top-level Makefile which calls cmake which generates more makefiles, seems quite a complex layering.

(TBH, I am also a make fan and have my own Makefile sitting in this directory, which does something similar but with more fine-grained control over the conan build steps and including some other things like generating the coverage statistics. Having to rename this would be a minor issue, but I don't know if other people do something similar.)

Another issue of course is that this is liable not to work on a standard Windows dev environment.

My suggestion is to move the new `Makefile` and `CMakeLists.txt` into a subdirectory, called something like `dev-utils` (open to other suggestions), with its own README, which can be mentioned in the top-level README.

What do you think?

trvto:
> I am in two minds. I see the convenience, but having a top-level Makefile which calls cmake which generates more makefiles, seems quite a complex layering.
> 
> (TBH, I am also a make fan and have my own Makefile sitting in this directory, which does something similar but with more fine-grained control over the conan build steps and including some other things like generating the coverage statistics. Having to rename this would be a minor issue, but I don't know if other people do something similar.)
> 
> Another issue of course is that this is liable not to work on a standard Windows dev environment.
> 
> My suggestion is to move the new `Makefile` and `CMakeLists.txt` into a subdirectory, called something like `dev-utils` (open to other suggestions), with its own README, which can be mentioned in the top-level README.
> 
> What do you think?

I'm up for any suggestions.

Moving to a subdirectory sounds like a good idea.

Also, It would be great if it worked on Windows as well and was hoping it would actually work but I don't know a whole lot about developing on Windows. Does anyone in the team develop on Windows? The makefile could work as is on WSL I would suspect but am also not 100% sure.

Regarding your Makefile and the possibility of others having their own Makefiles. In my opinion It would be great to have that stuff in the Repository so that not every developer is "reinventing the wheel" in order to have an efficient development process. And it's also just nice to have a set of working workflows to reference when trying to understand Conan, Cmake, etc.

Makefile
```diff
@@ -0,0 +1,103 @@
...
+	-conan remote add tket-libs $(tket_remote_repository)
+	conan config set general.revisions_enabled=1
+	conan profile update options.tket:shared=True $(conan_profile_name)
+	conan profile update options.tklog:shared=True $(conan_profile_name)
+	conan profile update settings.build_type=$(build_type) $(conan_profile_name)
+
+##@ Install Dependencies
+
+.PHONY: pybind
+pybind: ## install custom pybind package
```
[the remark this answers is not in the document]
trvto: Yes, I was hoping to integrate building pytket into the structure I am using here but couldn't get it to work in the short time I tried and wanted to get this PR out to see if it was something we generally wanted. So I would probably remove this for now and try to get it working at a later time

cqc-alec:
> Moving to a subdirectory sounds like a good idea.
> 
> Also, It would be great if it worked on Windows as well and was hoping it would actually work but I don't know a whole lot about developing on Windows. Does anyone in the team develop on Windows? The makefile could work as is on WSL I would suspect but am also not 100% sure.

We used to have a developer who worked on Windows but he left.

Sure, it would work on WSL (I assume), but official packages (built on the CI) are built using the VS toolchain, so I would expect Windows developers to use that.

> Regarding your Makefile and the possibility of others having their own Makefiles. In my opinion It would be great to have that stuff in the Repository so that not every developer is "reinventing the wheel" in order to have an efficient development process. And it's also just nice to have a set of working workflows to reference when trying to understand Conan, Cmake, etc.

I agree, and have been doing some work in this direction (in particular looking at migration to conan2 which has better cmake integration). The goal is to be buildable either with conan or cmake.

So my makefile is very much a temporary solution; it could as well have been a python script to run the build commands, but I hope eventually even that won't be necessary.

If you move these into a dev-tools folder I could add my own Makefile there as an alternative (though it is a bit of a hack).

cqc-alec (dismissed):
Thanks! Happy for this to be merged. Please use the "squash and merge" option.

recipes/tket/conanfile.py
```diff
@@ -20,7 +20,7 @@
 
 class TketConan(ConanFile):
     name = "tket"
-    version = "1.0.44"
+    version = "1.0.45"
```
cqc-alec, line 23: Sorry, this may need another bump.

cqc-alec (approved):

PR #711, opened 2023-01-04
Bugfix/multi gate reorder routing method missing edge

yao-cqc:
Given a mapping frontier, the algorithm tires to commute physically permitted vertices to the frontier so advance_frontier_boundary can move pass them. The bug is caused by the algorithm trying to bring a vertex before the frontier to the frontier. This fix added a check to ignore any vertices that lie before the frontier. Additionally, the vertices will now be iterated in topological order.


sjdilkes (approved):
Looks good to me.

PR #712, opened 2023-01-05
Update copyright notices to 2023

cqc-alec:


.github/workflows/compare-coverage
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-# Copyright 2022 Cambridge Quantum Computing
+# Copyright 2019-2023 Cambridge Quantum Computing
```
cqc-melf, line 3: Is this intended or should this be 2022-2023?

cqc-alec: This is intended; there were two files which had this notice but I decided to change them all to the same -- makes search-and-replace easier in the future. We don't generally track the start date of each file individually.

cqc-alec: Apart from those two files in this directory the only other change not covered by the search-and-replace was to `pytket/docs/conf.py`.

cqc-melf: Sounds good!

cqc-melf (approved):

PR #715, opened 2023-01-06
Update `LexiRoute::reassign_node` to use Ancilla Node

sjdilkes:
See TKET-2761. In Routing the `Qubit` `UnitID` objects in a `Circuit` are relabelled to `Node` objects from the `Architecture` the `Circuit` is being routed to be compatible with. There are scenarios where a `Qubit` is labelled as one `Node`, but we can freely relabel it to another `Node` if the `Node` it is already assigned to is more helpful being used elsewhere. In this case we need to relabel the first `Qubit` to _another_ `Node` on the `Architecture` that is not already in use.
When we do this, the logic in `LexiRoute::reassign_node` only checks for `Node` in the `Architecture` that aren't in use and relabels to them. However, it is possible there are no "free" `Node` in the `Architecture` as they have been used as ancillas. This PR updates `LexiRoute::reassign_node` to relabel to an ancilla `Node` if there are no "free" `Node` left to use.


tket/tests/test_LexiRoute.cpp
```diff
@@ -1739,4 +1740,228 @@ SCENARIO(
...
+  PassPtr r_p = gen_routing_pass(
+      architecture, {std::make_shared<LexiLabellingMethod>(),
+                     std::make_shared<LexiRouteRoutingMethod>()});
+  REQUIRE(r_p->apply(cu));
+}
+
+SCENARIO(
+    "Test relabelling a Circuit UnitID that is an Architecture Node but "
+    "reassignable to an Ancilla Node.") {
+  GIVEN("Line Architecture, one reassignment.") {
```
yao-cqc, line 1771: Can you add some comments explaining why some UnitID have been reassigned to an ancilla? My next comment is related to this.

tket/tests/test_LexiRoute.cpp
```diff
@@ -1739,4 +1740,228 @@ SCENARIO(
...
+
+    std::map<Qubit, Node> p_map = {// mapping for qbs with 2qb gates
+                                   {Qubit(0), nodes[0]},
+                                   {Qubit(1), nodes[8]}};
+    Placement::place_with_map(circuit, p_map);
+    CompilationUnit cu(circuit);
+    PassPtr r_p = gen_routing_pass(
+        architecture, {std::make_shared<LexiLabellingMethod>(),
+                       std::make_shared<LexiRouteRoutingMethod>()});
+    REQUIRE(r_p->apply(cu));
```
yao-cqc, line 1797: I think `REQUIRE(r_p->apply(cu));` (also in the other new tests) doesn't say much. Are there anything else we can check to prevent future regression?

yao-cqc (commented):
Looks good to me. Only a few comments. Also a changelog entry would be helpful.

tket/tests/test_LexiRoute.cpp
sjdilkes: Unfortunately the condition in which this occurs is rather niche. I've added a small explanation to the test:
• If a non-labelled Qubit in a Circuit being mapped has no Quantum gates with physical constraints (i.e. mostly multi-qubit gates) then before mapping we assign it some "bad" Architecture Node (typically something on the edge of the coupling graph with low out-degree)
• Any non-labelled Qubit in a Circuit with Quantum gates with physical constraints are left unlabelled
• During mapping, if a multi-qubit gate with a non-labelled Qubit is encountered we need to allocate it to some _best_ Architecture Node
• In some cases, this _best_ Node may end up being a "bad" Node we've used to assign an "unimportant" (without connectivity graph related physical constraints) Qubit too. If this is the case we relabel the unlabelled Qubit to this _best_ Node and find a new Node to assign the "unimportant" Qubit to
• Ideally we just find a spare Architecture Node that hasn't previously been assigned to and reassign the "unimportant" Qubit to it. However in some cases there can be no spare Architecture Node as they have been used as ancilla Node for SWAP/BRIDGE gates
• This is the case that was not considered before - the assertion is basically saying "there should always be a spare Architecture Node" but this is not true **(this isn't in the comments)**
• In this case we take an Ancilla Node and wire its output to the input of the "unimportant" Qubit Path, essentially reassigning it.

yao-cqc: Thanks! this is much appreciated.

yao-cqc (approved):

PR #719, opened 2023-01-10
add note to qasm API documentation

CalMacCQ:
Updating the documentation to alert users to the fact that implicit qubit permutations are not handled by the qasm converters. Will also update the user manual in a separate PR.

### DIFF
--- a/pytket/docs/qasm.rst
+++ b/pytket/docs/qasm.rst
@@ -11,5 +11,7 @@ to save your :py:class:`Circuit` objects.
 
 However, we do support symbolic parameters of gates, both on import and export.
 
+.. note:: Unlke pytket backends the qasm converters do not handle `implicit qubit permutations <https://cqcl.github.io/pytket/manual/manual_circuit.html#implicit-qubit-permutations>`_ . In other words if a circuit containing an implicit qubit permutation is converted to a qasm file the implicit permutation will not be accounted for and the circuit may appear to be different when reimported.
+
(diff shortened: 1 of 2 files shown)


pytket/docs/qasm.rst
```diff
@@ -11,5 +11,7 @@ to save your :py:class:`Circuit` objects.
 
 However, we do support symbolic parameters of gates, both on import and export.
 
+.. note:: Unlke pytket backends the qasm converters do not handle `implicit qubit permutations <https://cqcl.github.io/pytket/manual/manual_circuit.html#implicit-qubit-permutations>`_ . In other words if a circuit containing an implicit qubit permutation is converted to a qasm file the implicit permutation will not be accounted for and the circuit may appear to be different when reimported.
```
cqc-alec, line 14:
```suggestion
.. note:: Unlke pytket backends the qasm converters do not handle `implicit qubit permutations <https://cqcl.github.io/pytket/manual/manual_circuit.html#implicit-qubit-permutations>`_ . In other words if a circuit containing an implicit qubit permutation is converted to a qasm file the implicit permutation will not be accounted for and the circuit will be missing this permutation when reimported.
```

cqc-alec (changes requested):

pytket/docs/qasm.rst
CalMacCQ: Thanks, done.

cqc-alec (approved):

PR #727, opened 2023-01-17
offline circuit renderer

DNA386:
Include the scripts required for the circuit renderer directly within pytket so that it works offline.

Note that this means we now require npm to install pytket, in order to build the circuit renderer app locally.


cqc-alec:
Is there any way this could be made a separate package from pytket that can be installed optionally?

I'm a bit reluctant to merge this because of the enormous number of dependencies and the fact that it dumps 388 MB of files into the `pytket/pytket/circuit/display/node_modules/` directory! (Do these get copied into the wheels, or are they just a build artefact?)

Note that it is not just the `build_and_test` workflow that needs updating but also the other ones that build pytket, such as `release.yml` and assocaiated scripts.

DNA386:
The npm modules directory shouldn't be included in the final package, but get compiled into a single js file (display/dist/main.js) that bundles all the required dependencies together, which I believe is only 288KB

cqc-alec:
> The npm modules directory shouldn't be included in the final package, but get compiled into a single js file (display/dist/main.js) that bundles all the required dependencies together, which I believe is only 288KB

I see, OK. Could it be directed to a temporary build directory then?

The `packages.lock` file is also pretty big! How was it generated?

Still feel quite uneasy about the mass of dependencies and would prefer a distinct package if that is possible.

DNA386:
`package-lock.json` is generated automatically from the dependencies listing in `package.json`. Of note is that most of these are dev-only, and used to build the final js files, so the dependencies for the production output will be more streamlined.

I can look into moving the temp files elsewhere - I initially put them inside the display module since that is where they are relevant. In this case the only non-python files needed are `circuit.html` and `dist/main.js` - everything else is just there to help build those files.

I can also look into splitting off the display module as its own package.
If doing that would we also want to keep the current 'online' mode as well? (In which case circuit renderer updates would need to be managed in both pytket and the new pytket-display)

DNA386:
New version: offline renderer now exported as an extension.
This PR now refactors the setup of the online renderer to make it easier to change to offline rendering.
See proposed offline renderer code here: https://github.com/CQCL/pytket-offline-renderer

PR #728, opened 2023-01-18
[debug] wasm op in json serialisation

cqc-melf:
This PR is only for the debug discussion of the json issue.


tket/src/OpType/OpTypeJson.cpp
```diff
@@ -36,14 +36,19 @@ const std::map<std::string, OpType>& name_to_optype() {
...
+    throw JsonError("No OpType with name " + name);
+  }
 }
 
 void from_json(const nlohmann::json& j, OpType& type) {
   const std::string name = j.get<std::string>();
   const auto result = name_to_optype().find(name);
   if (result == name_to_optype().end()) {
-    throw JsonError("No OpType with name " + name);
+    throw JsonError("No OpType with name " + name);  // TODO wasm wire
```
cqc-melf, line 51: This is the place where the deserialisation fails because of an unknown op name "???"

tket/src/OpType/OpTypeJson.cpp
```diff
@@ -36,14 +36,19 @@ const std::map<std::string, OpType>& name_to_optype() {
 }
 
 void to_json(nlohmann::json& j, const OpType& type) {
-  j = optypeinfo().at(type).name;
+  j = optypeinfo().at(type).name;  // TODO wasm wire
+  const std::string name = j.get<std::string>();
+  const auto result = name_to_optype().find(name);
+  if (result == name_to_optype().end()) {
```
cqc-melf, line 42: I thought this op name "???" is generated here, but the same check that fails later in the "from_json" is not showing something.

tket/src/OpType/OpTypeInfo.cpp
```diff
@@ -76,6 +76,10 @@ const std::map<OpType, OpTypeInfo>& optypeinfo() {
        {"ClInput", "C IN", {}, op_signature_t({EdgeType::Classical})}},
       {OpType::ClOutput,
        {"ClOutput", "C OUT", {}, op_signature_t({EdgeType::Classical})}},
+      {OpType::WASMInput,
```
cqc-melf, line 79: This definition is used in the `optypeinfo().at(type).name;`

pytket/tests/backendinfo_test.py
```diff
@@ -168,6 +168,7 @@ def test_to_json() -> None:
 @settings(deadline=None)
 def test_backendinfo_serialization(backinfo: BackendInfo) -> None:
     serializable = backinfo.to_dict()
+    # print(serializable)
```
cqc-melf, line 171: At this place the dict already contains the "???" ops.

pytket/tests/strategies.py
```diff
@@ -174,6 +174,7 @@ def backendinfo(
     supports_fast_feedforward = draw(st.booleans())
     supports_reset = draw(st.booleans())
     supports_midcircuit_measurement = draw(st.booleans())
+    # is it possible to exclude some optyps at this point
```
cqc-melf, line 177: Here is the `BackendInfo` created. I have not found a way to convert this into some helpful information without using the json functions.

cqc-melf:
I have a problem with the test of the backendinfo serialisation. The problem is that at least one of the optypes is converted to "???" in the serialisation and that this optype can't be converted back in the deserialisation.

Is there an obvious place where the new optypes I created needs to be added which I missed?

PR #737, opened 2023-01-23
Add an error wrapper for compiler passes

### LINKED ISSUE #708: Improve error handling in `CXMappingPass`
The error message `Cannot commute Measure through quantum gates to the end of the circuit` is not very helpful when running `CXMappingPass` (with the default `delay_measures=True`) on a circuit containing a measurement followed by a reset.

aborgna-q:
Adds a transparent `ErrorWrapPass` transformer to catch errors in Passes and rethrow them with more contextual information.

fixes #708 with a more appropriate error message:
```python
>>> pss.apply(c)
  File "<input>", line 1, in <module>
    pss.apply(c)
RuntimeError: `delay_measures` is set, but the circuit contains measurements that cannot be delayed.
```

### DIFF
--- a/tket/src/Predicates/PassGenerators.cpp
+++ b/tket/src/Predicates/PassGenerators.cpp
@@ -267,7 +267,10 @@ PassPtr gen_cx_mapping_pass(
   PassPtr return_pass =
       rebase_pass >> gen_full_mapping_pass(arc, placement_ptr, config);
   if (delay_measures) {
-    PassPtr delay_meas = WrapError<CircuitInvalidity>(DelayMeasures(), "The circuit contains measurements that cannot be delayed.");
+    PassPtr delay_meas = WrapError<CircuitInvalidity>(
+        DelayMeasures(),
+        "`delay_measures` is set, but the circuit contains measurements that "
+        "cannot be delayed.");
     return_pass = return_pass >> delay_meas;
   }
   return_pass = return_pass >> rebase_pass >>
--- a/tket/src/Predicates/PassGenerators.cpp
+++ b/tket/src/Predicates/PassGenerators.cpp
@@ -266,7 +266,10 @@ PassPtr gen_cx_mapping_pass(
       gen_rebase_pass(gate_set, CircPool::CX(), CircPool::tk1_to_tk1);
   PassPtr return_pass =
       rebase_pass >> gen_full_mapping_pass(arc, placement_ptr, config);
-  if (delay_measures) return_pass = return_pass >> DelayMeasures();
+  if (delay_measures) {
+    PassPtr delay_meas = WrapError<CircuitInvalidity>(DelayMeasures(), "The circuit contains measurements that cannot be delayed.");
+    return_pass = return_pass >> delay_meas;
+  }
   return_pass = return_pass >> rebase_pass >>
                 gen_decompose_routing_gates_to_cxs_pass(arc, directed_cx);
   return return_pass;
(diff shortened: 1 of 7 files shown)


cqc-alec:
The build failure is because you changed the tket code without bumping the tket version number. This needs to be edited in four places: `recipes/tket/conanfile.py`, and the "requires" field in `recipes/tket-tests/conanfile.py`, `recipes/tket-proptests/conanfile.py` and `pytket/conanfile.txt`.

tket/src/Predicates/include/Predicates/PassLibrary.hpp
```diff
@@ -123,4 +123,23 @@ const PassPtr &CnXPairwiseDecomposition();
...
+ * Nests the old error inside it.
+ * @param pass Compilation pass to wrap
+ * @param error_msg Message to use for the new error
+ *
+ * @tparam Error Specific error type to wrap
+ * @tparam NewError Type of the new error to throw
+ *
+ * @return compilation pass to perform this transformation
+ */
+template <typename Error, class NewError = Error>
```
cqc-alec, line 137: How would we wrap a pass that can throw one of several errors, each of which we want to catch?

tket/src/Predicates/PassGenerators.cpp
```diff
@@ -266,7 +266,13 @@ PassPtr gen_cx_mapping_pass(
...
   PassPtr return_pass =
       rebase_pass >> gen_full_mapping_pass(arc, placement_ptr, config);
-  if (delay_measures) return_pass = return_pass >> DelayMeasures();
+  if (delay_measures) {
+    PassPtr delay_meas = WrapError<CircuitInvalidity>(
+        DelayMeasures(),
+        "`delay_measures` is set, but the circuit contains measurements that "
+        "cannot be delayed.");
+    return_pass = return_pass >> delay_meas;
+  }
```
cqc-alec, line 275: Can we add a test demonstrating that this message is part of the exception in this case?

tket/src/Predicates/include/Predicates/PassLibrary.hpp
aborgna-q: I was thinking of changing the signature from a string `error_msg` to a function _Error -> NewError_ (and an overloaded version for when we just want to change the string). That would let us dynamically match on multiple kinds of errors.

The problem with that is that then the wrapper cannot be serialized...

tket/src/Predicates/PassGenerators.cpp
```diff
@@ -266,7 +266,13 @@ PassPtr gen_cx_mapping_pass(
       gen_rebase_pass(gate_set, CircPool::CX(), CircPool::tk1_to_tk1);
   PassPtr return_pass =
       rebase_pass >> gen_full_mapping_pass(arc, placement_ptr, config);
-  if (delay_measures) return_pass = return_pass >> DelayMeasures();
+  if (delay_measures) {
+    PassPtr delay_meas = WrapError<CircuitInvalidity>(
+        DelayMeasures(),
+        "`delay_measures` is set, but the circuit contains measurements that "
+        "cannot be delayed.");
```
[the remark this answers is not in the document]
aborgna-q: Something like a `class DelayMeasureError : public CircuitInvalidity` ?

Barring bugs, `DelayMeasures()` only throws when it can commute the gate, but it doesn't hurt to add more specific errors.

aborgna-q:
After toying with the idea, an error wrapper won't be serializable without some way to dynamically allocate errors from a serialized identifier.
I'm closing this PR, and going with another approach for #708.

PR #744, opened 2023-01-24
Add a CommutableMeasuresPredicate

### LINKED ISSUE #708: Improve error handling in `CXMappingPass`

aborgna-q:
Adds a Predicate to validate when all measurements can be commuted to the end of the circuit via `delay_measures()`. This is given as a precondition of the `DelayMeasures` pass, so now the pass doesn't throw errors while running.

The implementation of `delay_measures` didn't check inside boxes and conditionals. I unified the logic between the `DelayMeasures` pass, the new `CommutableMeasuresPredicate`, and the `NoMidMeasuresPredicate` so now they all look inside nested operations. If a measurement is found inside, the circuit is only found valid if the units of that measure are not interacted with afterwards (this works even when the unit is connected to other circBoxes which do not modify it).

Now if a circuit validates the `NoMidMeasuresPredicate` it will never be modified by a `DelayMeasures` pass.

This should close #708, which now throws an error with the specific predicate that fails to validate.


tket/src/Predicates/include/Predicates/Predicates.hpp
```diff
@@ -238,6 +238,18 @@ class NoBarriersPredicate : public Predicate {
...
+ public:
+  bool verify(const Circuit& circ) const override;
+  bool implies(const Predicate& other) const override;
+  PredicatePtr meet(const Predicate& other) const override;
+  std::string to_string() const override;
+};
+
 /**
  * Asserts that any measurements occur at the end of the circuit
  */
```
aborgna-q, line 255: Making `NoMidMeasurePredicate` inherit from `CommutableMeasuresPredicate` here breaks the serialization tests. I could not find the reason for that.

cqc-alec (commented):
Thanks, looks good! Please could you add a line to the changelog (`pytket/docs/changelog.rst`) since this is a new pytket feature?

cqc-alec (approved):

PR #748, opened 2023-01-26
Feature/state preparation

yao-cqc:
Implement the state prep method explained in the section 4 of [this paper](https://arxiv.org/abs/quant-ph/0406176)


tket/src/Gate/Rotation.cpp
```diff
@@ -462,6 +462,25 @@ std::vector<double> tk1_angles_from_unitary(const Eigen::Matrix2cd &U) {
...
 }
 
+std::tuple<double, double, double> get_bloch_coordinate_from_state(
+    const Complex &a, const Complex &b) {
+  // We require the state to be normalised
+  double norm = a.real() * a.real() + a.imag() * a.imag() +
+                b.real() * b.real() + b.imag() * b.imag();
+  if (std::abs(norm - 1) > EPS) {
+    throw std::invalid_argument(
+        "Attempt to find the block sphere coordinates for an unnormalised "
```
sjdilkes, line 472: Could you update this error message to also explain why it failed please

sjdilkes (changes requested):
Really great work - one request to make an error message provide more information but apart from that al good to merge!

tket/src/Gate/Rotation.cpp
yao-cqc: Is the updated message better? or do I still need to add something?

sjdilkes (dismissed):

sjdilkes (approved):

PR #775, opened 2023-03-08
Renderer update

DNA386:
Bump renderer version to 0.6

-- Edit

This PR also includes some more recent bugfixes:
- Allow resizing the circuit display container on firefox
- Display container when viewing circuit in the browser is by default smaller

### DIFF
--- a/pytket/pytket/circuit/display/__init__.py
+++ b/pytket/pytket/circuit/display/__init__.py
@@ -99,14 +99,24 @@ class CircuitRenderer:
         if jupyter:
             # If we are in a notebook, we can tell jupyter to display the html.
             # We don't import at the top in case we are not in a notebook environment.
+            import warnings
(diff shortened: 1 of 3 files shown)


cqc-alec (dismissed):
(mypy failure is unrelated)

DNA386:
The new bugfix relates to large circuits making the jupyter notebook crash silently. Was particularly bad for the offline renderer since the html needs to include the circuit renderer code and dependencies, but this should resolve both.

CalMacCQ (approved):

CalMacCQ:
Once the branch is updated could you double check that the `pytket.display` docs html pages looks okay? I can also do this if need be.

DNA386:
Docs look fine, reference the correct package name.

CalMacCQ:
Thanks, happy for this to be merged.

PR #779, opened 2023-03-10
fix typo in citation

ss2165:


CalMacCQ:
Ah my bad, thanks for correcting.

CalMacCQ (approved):

PR #782, opened 2023-03-13
Support `allow_swaps` parameter for `FullPeepholeOptimise` even when targeting `OpType.TK2`

cqc-alec:


yao-cqc:
Do you know why this wasn't supported?

cqc-alec:
> Do you know why this wasn't supported?

I couldn't remember, but it looks like it comes from https://github.com/CQCL/tket/pull/476 .

From there it sounds as if the present PR does not change the behaviour at all -- implicit swaps would never be added with TK2 even if allowed.

It may still be worth making this change, since it allows us to change that behaviour if we want to -- though I just realized the documentation of `two_qubit_squash()` would also need updating.

Alternatively we could just update the documentation to say that with `TK2` target gates swaps are never introduced -- but that binds us to it in the future, which may be undesirable.

yao-cqc (approved):
Thanks, all makes sense.

cqc-alec:
On second thoughts, I think this PR does change the behaviour, since it allows for instance `clifford_simp()` to introduce swaps. So I do think it's worth doing.

PR #789, opened 2023-03-16
Main

cqc-melf:


CalMacCQ:
Should the new ZX pass from Will appear in the changelog?  https://github.com/CQCL/tket/pull/769

cqc-alec (approved):

cqc-alec:
> Should the new ZX pass from Will appear in the changelog? #769

Yes that should have gone in the changelog. We could add it retrospectively.

cqc-melf:
Should we make a new PR + release to add that? I have found one small issues in the wasm documentation that could be updated together with this.

cqc-alec:
> @cqc-alec Should we make a new PR + release to add that? I have found one small issues in the wasm documentation that could be updated together with this.

Not worth doing a new release, but these things could be fixed in a new PR.

PR #796, opened 2023-03-23
Allow custom labelling of Architecture Generating Classes

sjdilkes:
When using `Architecture`, the class constructor takes pairs of Node as arguments which can be labelled by the user on creation.

When using `SquareGrid`, `RingArch` and `FullyConnected`, the class constructors just take dimension arguments and return an `Architecture` object that has the connection implied by these classes. 

This PR adds a new optional argument, `label` to the `TKET` and `pytket` `SquareGrid`, `RingArch` and `FullyConnected`. When passed, this method names all `Node` as the argument passed to `label`, instead of some other default.

### DIFF
--- a/pytket/docs/changelog.rst
+++ b/pytket/docs/changelog.rst
@@ -9,6 +9,10 @@ Minor new features:
 * Update to networkx 3.
 * Add "label" argument to ``SquareGrid``, ``RingArch`` and ``FullyConnected`` 
   ``Architecture`` classes to give custom name to constructed ``Node``.
+  
+Fixes:
+
+* Multiply symbolic parameters in auto-generated gate definitions by "/pi" in ``circuit_to_qasm_io``
 
 1.13.1 (March 2023)
 -------------------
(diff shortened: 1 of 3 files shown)


pytket/binders/architecture.cpp
```diff
@@ -103,20 +103,28 @@ PYBIND11_MODULE(architecture, m) {
...
+            return SquareGrid(x, y, 1, label);
+          }),
           "The constructor for a Square Grid architecture with some "
           "undirected connectivity between qubits.\n\n:param n_rows: "
           "The number of rows in the grid\n:param n_columns: The number "
-          "of columns in the grid",
-          py::arg("n_rows"), py::arg("n_columns"))
+          "of columns in the grid\n:param label: Name for Node in SquareGrid "
+          "Architecture",
+          py::arg("n_rows"), py::arg("n_columns"), py::arg("label"))
```
yao-cqc, line 115: Does this `label` also need a default?

sjdilkes: Yeah it does - I've added.

yao-cqc (approved):

PR #803, opened 2023-03-28
Feature/update docs version

CalMacCQ:
Upgrading docs theme to `sphinx_book_theme ~= 1.0.1` -> https://sphinx-book-theme.readthedocs.io/en/stable/changelog.html

Note this issue https://github.com/executablebooks/sphinx-book-theme/issues/711  (now fixed - using `sphinx_book_theme ~= 1.0.1` instead of 1.0.0).

The updated theme has the ability to toggle between dark and light mode. The css rules need to be changed to preserve our old design. The theme also allows for a navbar across the top of the page but I'll do that in a separate PR I think.

Here's how the docs page will look in light and dark mode. I've just used the default setting for dark mode at the moment.

Screenshot 2023-03-29 at 14 31 03

Screenshot 2023-03-29 at 14 32 51


pytket/docs/requirements.txt
```diff
@@ -1,6 +1,6 @@
 sphinx >= 4.5
 sphinx_autodoc_annotation >= 1.0
-sphinx_book_theme ~= 0.3.3
+sphinx_book_theme >= 1.0
```
cqc-alec, line 3:
```suggestion
sphinx_book_theme ~= 1.0
```
Probably better so that we don't pick up 2.0 accidentally.

cqc-alec (approved):

CalMacCQ:
Do you think its okay to merge this now. I suppose by next release we would ideally want to be using this everywhere. There were some issues with the DAG diagrams not looking so great in the manual for instance.

cqc-alec (approved):

PR #805, opened 2023-03-29
Release/1.13.2

cqc-alec:


cqc-melf (approved):

cqc-melf:
The circuit renderer update is not in the changelog, is that intended? @cqc-alec

cqc-alec:
> The circuit renderer update is not in the changelog, is that intended? @cqc-alec

Not intended, but I think we can live with it..

PR #809, opened 2023-03-30
Use conan2 for tket utility libraries

cqc-alec:
Note, most of the changes here are just moving files. There is no change in functionality.

Having fought for some time with the CI coverage checks I decided to remove them for now and made a new task to put them back. (EDIT: Added now.)

After this I will move on to the main tket library, tests, and pytket.

Another task is to add config generators to the cmake files so that the whole thing can be built without conan.


.github/workflows/test_libs.yml
```diff
@@ -64,36 +64,30 @@ jobs:
...
     - name: Set up Python 3.10
       uses: actions/setup-python@v4
       with:
         python-version: '3.10'
-    - name: install conan
-      run: pip install conan~=1.53
+    - name: Install conan
+      uses: turtlebrowser/get-conan@v1.2
+      with:
+        version: '2.0.2'
```
cqc-melf, line 76: Do we want to specify a fixed version here or something that is automatically updated?

.github/workflows/build_libs.yml
```diff
@@ -104,35 +92,37 @@ jobs:
...
-      run: conan upload ${{ matrix.lib }}/${LIB_VER}@tket/stable --all -r=tket-libs --skip-upload
+        lib_label=`./libver ${{ matrix.lib }}.json ${{ matrix.lib }}`
+        echo "LIB_LABEL=${lib_label}" >> $GITHUB_ENV
     - name: upload package
       if: github.event_name == 'push'
-      run: conan upload ${{ matrix.lib }}/${LIB_VER}@tket/stable --all -r=tket-libs
+      run: conan upload ${{ env.LIB_LABEL }} -r=tket-libs
     - name: unauthenticate
-      run: conan user -c
+      if: always()
```
cqc-melf, line 119: is always() a good condition?

.github/workflows/linuxbuildlib
```diff
@@ -19,28 +19,22 @@ set -evu
 # Choose a Python to install conan
 export PYBIN=/opt/python/cp39-cp39/bin
 
-export CONAN_REVISIONS_ENABLED=1
-
 ${PYBIN}/pip install --upgrade pip
-${PYBIN}/pip install conan~=1.53
+${PYBIN}/pip install conan~=2.0.2
```
cqc-melf, line 23:
```suggestion
${PYBIN}/pip install conan==2.0.2

Comment (cqc-melf) [reply .github/workflows/linuxbuildlib:23]: See question below, do we want this version or 2.0.x ?

**Review by cqc-melf — COMMENTED** [review]
Some few comments, I have not tried something yet. Maybe it would be good to do that?

File: .github/workflows/build_libs.yml (hunk 2 above)
Comment (cqc-alec) [reply .github/workflows/build_libs.yml:119]: Yes, this ensures it is always executed even when a step fails.

File: .github/workflows/linuxbuildlib (hunk 3 above)
Comment (cqc-alec) [reply .github/workflows/linuxbuildlib:23]: Happy to have compatibility (`~=`) here, but maybe actually it should be `~=2.0` ... will change.

Comment (cqc-alec) [reply .github/workflows/linuxbuildlib]: Changed.

**Comment by cqc-alec** [comment]
I put the coverage checks back. The line coverage is different from before, causing one check to fail, I think because it only checks the cpp files, not the hpp files. I think we should ignore this failure (and upload the new numbers).

**Comment by cqc-alec** [comment]
> I put the coverage checks back. The line coverage is different from before, causing one check to fail, I think because it only checks the cpp files, not the hpp files. I think we should ignore this failure (and upload the new numbers).

Now made it include the hpp files too as before. One is still failing, apparently detecting an extra line -- I don't know why (perhaps something changed between gcovr 5 and 6). I suggest ignoring this failure.

**Comment by cqc-alec** [comment]
> > I put the coverage checks back. The line coverage is different from before, causing one check to fail, I think because it only checks the cpp files, not the hpp files. I think we should ignore this failure (and upload the new numbers).
> 
> Now made it include the hpp files too as before. One is still failing, apparently detecting an extra line -- I don't know why (perhaps something changed between gcovr 5 and 6). I suggest ignoring this failure.

Old:
image
New:
image
Does look like a tooling issue.

**Comment by cqc-melf** [comment]
> > > I put the coverage checks back. The line coverage is different from before, causing one check to fail, I think because it only checks the cpp files, not the hpp files. I think we should ignore this failure (and upload the new numbers).
> > 
> > 
> > Now made it include the hpp files too as before. One is still failing, apparently detecting an extra line -- I don't know why (perhaps something changed between gcovr 5 and 6). I suggest ignoring this failure.
> 
> Old: image New: image Does look like a tooling issue.

I think this can be ignored

**Comment by cqc-melf** [comment]
I had some issues with the readme, but I now got it to work.

I think it would be good to add some lines to the conan 2 section, I would suggest the following:

Required when first using conan 2.0.2:

You need to use cmake version >= 3.26, you can check it with:

`cmake --version`

Install and setup conan:

`pip install conan==2.0.2`

Create a new profile for conan 2:

`conan profile detect --name tket`

We want to build tklog as shared rather than static libraries, so set this in the profile:

`conan profile update options.tklog:shared=True tket` (not sure if needed, with conan 2 not working)

Set the tket-libs repository as your remote. (Note that the following commands affect your conan configuration across all projects, so if you are working on other projects with conan you will want to revert them afterwards. A simple way is to back up the file ~/.conan/remotes.json. You can view your current remotes list with conan remote list.)

```
conan remote clean
conan remote add tket-libs https://quantinuumsw.jfrog.io/artifactory/api/conan/tket1-libs
```

To build the library tkxxx:

(Continue with the comments already in the readme)

File: libs/README.md [hunk 4]
```diff
@@ -11,3 +11,64 @@ downward line means "is required by"):
           |    X   |
           |   / \  |
     tktokenswap tkwsm
+
+# Building
```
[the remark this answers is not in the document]
cqc-melf: Done

cqc-melf (commented):
The changes are looking fine and I did the local checks (see comments about the readme)

The only thing left would be updating / discussion the readme.

cqc-alec:
> The changes are looking fine and I did the local checks (see comments about the readme)
> 
> The only thing left would be updating / discussion the readme.

Thanks, I'll update the README.

Building tklog as a shared library shouldn't be necessary here. It's necessary further down the line when we build pytket. But I will add instructions here about how to build the libraries as shared or static.

libs/README.md
```diff
@@ -17,8 +17,30 @@ downward line means "is required by"):
...
 ## conan 1
 
+```shell
+pip install conan~=1.59
+```
+
+Create a new profile called `tket`:
+
+```
+conan profile detect --name tket
```
cqc-melf, line 32:
```suggestion
conan profile new tket --detect

File: libs/README.md [hunk 6]
```diff
@@ -45,18 +67,37 @@ cd ${PKGPATH}/bin
...
 ## conan 2
 
+```shell
+pip install conan~=2.0
+```
+
+Use the default profile:
+
+```
+conan profile detect
```
cqc-melf, line 82: no name for the profile?

libs/README.md
cqc-alec: > There is a difference for conan 1 and 2

Yes, because there is no need to create a custom named profile as we're not updating it: in conan 2 the Linux `libcxx` configuration is correctly set with `detect`. It is easiest to use the `default` profile.

libs/README.md
cqc-alec: (See above.)

libs/README.md
cqc-alec: > I think this is in the conan 1 part and `conan profile detect` is not working for that? (Or I am doing something wrong locally in conan 1?)

Sorry yes I'll fix that.

cqc-melf:
Just to check: merging this will override the default values for the coverage checks?

libs/README.md
cqc-alec: > Are we sure that we don't want to use a specific profile to allow for some individual setting like debug mode or something like that in conan 2?

Individuals can do that if they want of course -- but these instructions aren't meant to cover all cases, just enough to do a build.

cqc-alec:
> Just to check: merging this will override the default values for the coverage checks?

Yes, that is what I expect to happen. (The "check coverage against latest" step is only run on pull requests.)

libs/README.md
cqc-alec: > > I think this is in the conan 1 part and `conan profile detect` is not working for that? (Or I am doing something wrong locally in conan 1?)
> 
> Sorry yes I'll fix that.

Done.

cqc-melf (approved):
I am happy now

PR #812, opened 2023-04-05
Feature/multiplexed tensor u2 box

yao-cqc:


yao-cqc:
Also replaced some `unsigned` with `unsigned long long` to prevent potential overflow.

pytket/docs/changelog.rst
```diff
@@ -8,6 +8,14 @@ Major new features:
 
 * Support for ARM Linux platforms.
 
+Minor new features:
+
+* Updated implementation of ``ToffoliBox`` utilising multiplexors
+  for improved decomposition.
+* Add new ``DiagonalBox`` that synthesises a diagonal unitary matrix
+  into a sequence of multiplexed-Rz gates.
+* Add new ``MultiplexedTensoredU2Box`` that synthesises multiplexed tensor product of U2 gates.
```
sjdilkes, line 17: Maybe these all count as major new features? @cqc-alec what do you think?

tket/src/Circuit/Multiplexor.cpp
```diff
@@ -429,23 +434,30 @@ MultiplexedRotationBox::MultiplexedRotationBox(const ctrl_op_map_t &op_map)
...
+        throw BadOpType(
+            "Ops passed to MultiplexedRotationBox must be either Rx, Ry, or "
+            "Rz.",
+            axis_);
       }
     } else {
       if (it->second->get_type() != axis_) {
-        throw std::invalid_argument("Ops must have the same rotation type.");
+        throw std::invalid_argument(
+            "Ops passed to MultiplexedRotationBox must have the same rotation "
```
sjdilkes, line 459: Nice updates on error messages.

tket/src/Circuit/Multiplexor.cpp
```diff
@@ -630,29 +643,222 @@ void MultiplexedU2Box::generate_circuit() const {
...
+        multip_rz.insert({bitstr, get_op_ptr(OpType::Rz, alpha)});
+      }
+      // update the diagonal on the control qubits
+      diag_vec[j] *= p;
+    }
+    if (!multip_rz.empty()) {
+      diag_circ.add_box(MultiplexedRotationBox(multip_rz), args);
+    }
+  }
+  if (impl_diag_) {
```
sjdilkes, line 854: So if `impl_diag_=False` then the Diagonal operator isn't added to the Circuit at all?

I'd maybe suggest that the functionality of `impl_diag_` is as to whether the diagonal operator is decomposed in this `generate_circuit` or left as a `DiagonalBox`. Ideally if a user chooses to leave it as a `DiagonalBox` then there are later compilation passes they can apply (that we would need to write!) that would suitably commute these through end of circuit measures, or other sets of controls etc to lead to gate cancellations.

sjdilkes (changes requested):
Code all looks great. Only comment I have is on the functionality of `impl_diag_`: I'd suggest `impl_diag_` determines whether the `DiagonalBox` is decomposed in `MultiplexedTensoredU2Box::generate_circuit` (if true) or left as a `DiagonalBox` (if false), instead of not adding it to the Circuit. 

The follow up to this would be to add a compilation pass that is able to commute and merge `DiagonalBox` appropriately to lead to gate count reduction, i.e. things like merging them with certain other `Box` types and commuting them past end of circuit measures or controls.

tket/src/Circuit/Multiplexor.cpp
yao-cqc: Makes sense. I would suggest that the behaviour of `impl_diag_` in `MultiplexedU2Box` should also be changed for the sake of consistency. It sounds like the commute DiagonalBox pass you suggested must be prioritised because removing them is not straightforward with the python API.

pytket/docs/changelog.rst
cqc-alec: Agreed , major new features.

tket/src/Circuit/Multiplexor.cpp
sjdilkes: Agreed on consistency suggestion.

Also agreed on prioritising such a commuting `DiagonalBox` pass. We could maybe do something sneaky by

yao-cqc: On second thought, since the decomposition already produces box type gates, I don't see any reason to decompose the `DiagonalBox` in `generate_circuit`. How about I simply remove the `impl_diag_` from `MultiplexedTensorU2Box`? We can keep `MultiplexedU2Box` as it is for now.

sjdilkes: Works for me!

pytket/docs/changelog.rst
```diff
@@ -16,6 +20,10 @@ Minor new features:
 Major new features:
 
 * Support for ARM Linux platforms.
+* Updated implementation of ``ToffoliBox`` utilising multiplexors
```
sjdilkes, line 23: I think ToffoliBox and DiagonalBox can both be moved to Major new features also

sjdilkes (changes requested):
Looking good, just the single request to make DiagonalBox and updated ToffoliBox also major new features.

sjdilkes (approved):

Issue #746, opened 2023-01-26
c++ Circuit::assert_valid() failing for some valid circuits


cqc-melf:
The function has some problems with a vertex containing more than one quantum and more then one classical port in the vertex signature. Some of the problems might be solved by https://github.com/CQCL/tket/pull/743 , but probably not all.

PR #813, opened 2023-04-05
fix circuit assertion

cqc-melf:
fix https://github.com/CQCL/tket/issues/746


cqc-melf:
Fix the missing handling of nodes with more than one classical and more than one quantum vertices to the is_valid() function of the circuit dag.

tket/src/Circuit/DAGProperties.cpp
```diff
@@ -231,10 +241,7 @@ bool is_valid(const DAG &G) {
       CHECK(w_out_empty);
       CHECK(c_in_ports == c_out_ports);  // bijection between in and out ports
       CHECK(q_in_ports == q_out_ports);  // bijection between in and out ports
-      // this case should only happen for measure ops
-      CHECK(
-          q_in.size() == 1 && q_out.size() == 1 && c_in.size() == 1 &&
-          c_out.size() == 1);
+      CHECK(q_in.size() == q_out.size() && c_in.size() == c_out.size());
```
yao-cqc, line 244: I think this condition is implied by the:

```
      CHECK(c_in_ports == c_out_ports);  // bijection between in and out ports
      CHECK(q_in_ports == q_out_ports);  // bijection between in and out ports
```

and the previous:

```
    CHECK(q_in_ports.size() == q_in_deg);
    CHECK(q_out_ports.size() == q_out_deg);
    
    CHECK(c_in_ports.size() == c_in_deg);
    CHECK(c_out_ports.size() == c_out_deg);
```

cqc-melf: Oh yes you are right, I forgot about the other one. I have removed that.

yao-cqc (approved):
Thanks!

PR #817, opened 2023-04-12
Feature/allow barriers in q control box

### LINKED ISSUE #807: Allow barriers in QControlBoxes
Barriers within the controlled circuit should be left in place (with no control).

yao-cqc:
Resolves #807


tket/src/Circuit/CircUtils.cpp
```diff
@@ -995,6 +1029,16 @@ static Circuit with_controls_numerical(const Circuit &c, unsigned n_controls) {
...
       continue;
     }
+    // Barriers are left in place
+    if (b.is_barrier) {
+      qubit_vector_t new_args;
+      for (const unsigned i : b.control_qubits) {
+        new_args.push_back(Qubit(i + n_controls));
+      }
+      new_args.push_back(Qubit(b.target_qubit + n_controls));
+      c2.add_op(b.ops[0], new_args);
```
cqc-alec, line 1039: Maybe assert that `b.ops.size() == 1`?

cqc-alec (changes requested):
Needs tket lib version update now.

cqc-alec (approved):
Great!

PR #819, opened 2023-04-13
Release/1.14

cqc-alec:


cqc-melf:
Should we do an RC release first?

cqc-alec:
> Should we do a dev release first?

That's a good idea, I'll do that now.

cqc-alec:
> > Should we do a dev release first?
> 
> That's a good idea, I'll do that now.

Done: pytket 1.14.0rc0

cqc-melf (approved):

Issue #778, opened 2023-03-10
Classical ops missing from API documentation


cqc-alec:
For example, `CopyBitsOp`,  `RangePredicateOp` etc are not documented.

cqc-melf:
This also effect MultiBitOp, SetBitsOp, ClassicalEvalOp, ClassicalOp

PR #825, opened 2023-04-18
add missing op to docs

cqc-melf:
solves https://github.com/CQCL/tket/issues/778


CalMacCQ:
I think maybe we should remove listing alll the classical ops classes here if possible

https://cqcl.github.io/tket/pytket/api/circuit.html

cqc-melf:
Do you think they should be removed or listed somewhere else? @CalMacCQ

cqc-alec (approved):
Thanks!

PR #826, opened 2023-04-19
Build everything with conan 2

cqc-alec:
After merging I will update the names of the "required" checks, which have changed (hence the apparently incomplete checks on his PR).

This completes the migration of the build system to conan 2, by including pytket.

An important next step is uploading pre-built binaries for the dependencies in the required configurations to the artifactory, so that CI builds take less long.

There are other small bits of clean-up to do, such as removing conan1 restoration from the M1 workflow scripts, once this is merged.

Then: make any necessary tweaks and document how to build using cmake only.


README.md
```diff
@@ -49,11 +49,11 @@ setting `CC` and `CXX` in your environment (e.g. `CC=gcc-11` and `CXX=g++-11`),
 or on Debian-based Linux systems using `update-alternatives`.
 
 You should also have Python (3.9, 3.10 or 3.11) and `pip` installed. We use
-`cmake` and the package manager `conan` to build tket. Both can be installed
-with `pip`:
+`cmake` and the package manager `conan` to build tket and pytket. The latter can
```
cqc-melf, line 52: Should we point out the cmake version requirement at this point?

.github/workflows/build_and_test.yml
cqc-melf, line 1: From my understanding the `check_changes` is not detecting changes to the .github directory, is that something we maybe should add?

.github/workflows/build_and_test.yml
```diff
@@ -222,97 +216,40 @@ jobs:
         key: ${{ github.job }}-${{ runner.os }}-ccache-${{ steps.current_time.outputs.formattedTime }}
         restore-keys: |
           ${{ github.job }}-${{ runner.os }}-ccache-
-    - name: apt update
-      if: matrix.os == 'ubuntu-22.04'
-      run: sudo apt update
-    - name: Check doxygen
```
cqc-melf, line 228: Where has the doxygen doc build moved over to?

.github/workflows/check-tket-reqs
```diff
@@ -18,35 +18,20 @@ version = sys.argv[1]
...
-        )
+assert conan_ver == 2
+for path in paths:
+    with open(jsondump, "w") as f:
+        subprocess.call(["conan", "graph", "info", path, "--format", "json"], stdout=f)
+    with open(jsondump) as f:
+        info = json.load(f)
+    os.remove(jsondump)
+    assert any(
+        node["ref"].startswith(f"tket/{version}@tket/stable") for node in info["nodes"]
```
cqc-melf, line 36: Do we want to make checks for the libs here as well?

dev-tools/README.md
```diff
@@ -1,5 +1,8 @@
 # Development Tools
 
+*NB* The notes and scripts in this directory are obsolete since the migration
```
cqc-melf, line 3: Do we want to handle this now? Or is there a task created for that?

.github/workflows/build_and_test.yml
cqc-melf: Questions answered

README.md
cqc-alec: Will do.

.github/workflows/build_and_test.yml
cqc-alec: Not sure, the purpose of this `check_changes` is to check for changes to the code (and recipes). It's true we'd want to test changes to the workflow when the code doesn't change, but I think that can be done by developer tests and review.

dev-tools/README.md
cqc-alec: I'd rather not handle this now. I will create a task. (As far as I know only one person uses these scripts and they're not tested on the CI.)

README.md
cqc-alec: Actually it already says: "You will need at least cmake version 3.26".

cqc-alec:
> We are not running a coverage check on the binders yet? Or did I miss that somewhere?

No, I don't think the coverage of the binders code is tested. (Not sure how we'd do that.)

README.md
cqc-melf: Oh yes, I missed that

.github/workflows/build_and_test.yml
cqc-melf: I think it would be good to add `.github/**` to all of the filter in he line https://github.com/CQCL/tket/blob/feature/conan2-pytket/.github/workflows/build_and_test.yml#L39 and following. With developer tests you mean something separate to the PR checks?

cqc-alec: As it stands, this could cause problems e.g. with the line https://github.com/CQCL/tket/blob/feature/conan2-pytket/.github/workflows/build_and_test.yml#L62 which would mean we'd need to bump the tket version whenever the workflow file changes, which is unnecessary: `tket_changed` currently just captures the data that feeds into the conan recipe version.

So I think this would need some reworking, and since we're not departing from the existing logic maybe it should be deferred to another PR?

cqc-melf: Yes we can do that

README.md
```diff
@@ -65,246 +67,31 @@ sudo apt install ninja-build ccache
...
 
-The code coverage of the `pytket` tests is reported
-[here](https://cqcl.github.io/tket/pytket/test-coverage/index.html). This report
-reflects the coverage of the `develop` branch, and is updated with every push.
-The same report can be found in XML format
-[here](https://cqcl.github.io/tket/pytket/test-coverage/cov.xml).
+### Building and testing pytket
 
-Lines and branch coverage results are also checked with every PR to `develop`.
+See the [README](pytket/README) in the `pytket` directory for instructions on
```
cqc-melf, line 93: This link give a 404 for me?

README.md
```diff
@@ -65,246 +67,31 @@ sudo apt install ninja-build ccache
...
+random-number generation and so on). This is not necessary if you just want to
+build tket or pytket since the recipes or binaries will be automatically
+downloaded from the above conan remote.
 
-In both cases, PRs to `develop` check that the coverage has not decreased, and
-merging is blocked until the coverage is at least as good as before.
+### Building and testing the tket library
 
-### pytket
+See the [README](tket/README) in the `tket` directory for instructions on
```
cqc-melf, line 88: This link give a 404 for me?

tket/README.md
```diff
@@ -60,88 +40,65 @@ cd ${PKGPATH}/bin
...
 
-To build the property tests:
-
-```shell
-conan create tket/proptest --build=missing  -o boost/*:header_only=True --format json > proptest-tket.json
-```
+## Generating a test coverage report
 
-To extract the root package path to environment variable:
+If using gcc and with gcovr installed (`pip install gcovr`), a test coverage
```
cqc-melf, line 87: I am only able to get this to work without gcovr from python and with the global installation. Should we change this here? If this is working for you I am happy to keep it as it is and just will keep this in mind

README.md
cqc-alec: Oops, missed the `.md`, thanks.

README.md
cqc-alec: Missed the `.md`, thanks.

tket/README.md
cqc-alec: Hmm, works for me with `gcovr` installed using `pip`.

cqc-alec: (Works on the CI too...)

cqc-melf: Do you have a global installation with the same version?

cqc-melf: If that is the way it works on the CI lets keep it as it is :)

cqc-alec: > Do you have a global installation with the same version?

No I don't have a version installed outside of my python environment.

README.md
cqc-alec: Fixed.

README.md
cqc-alec: Fixed.

cqc-melf (approved):

PR #835, opened 2023-04-24
Feature/circuit renderer default options

DNA386:
Allow user to set default options for the circuit renderer:
The usual functions with default options remain available, but now the an instance of the renderer class is also exposed, with methods to set user preferences.

Also make the initial display size in jupyter taller.

### DIFF
--- a/pytket/docs/display.rst
+++ b/pytket/docs/display.rst
@@ -10,21 +10,49 @@ Contains several functions for rendering interactive circuit diagrams.
         pip install pytket-offline-display
 
 .. automodule:: pytket.circuit.display
-    :members: render_circuit_jupyter, render_circuit_as_html, view_browser
+    :members: get_circuit_renderer, CircuitRenderer
+
+Example usage:
+--------------
 
 .. jupyter-execute::
 
     from pytket import Circuit
-    from pytket.circuit.display import render_circuit_jupyter
+    from pytket.circuit.display import get_circuit_renderer
 
-    circ = Circuit(2) # Define Circuit
-    circ.H(0).H(1).CX(0, 1).Rz(0.4, 1).CX(0, 1).H(0).H(1)
-    render_circuit_jupyter(circ) # Render interactive display
-    
+    circuit_renderer = get_circuit_renderer() # Instantiate a circuit renderer
+    circuit_renderer.set_render_options(zx_style=False) # Configure render options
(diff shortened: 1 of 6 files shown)


pytket/pytket/circuit/display/__init__.py
```diff
@@ -71,9 +71,68 @@ def _render(self, filename: str) -> markupsafe.Markup:
...
+    min_width: str = "500px"
+
     def __init__(self, env: Environment):
         self.env = env
 
+    def set_render_options(self, **kwargs: Union[bool, str]) -> None:
+        """
+        Set rendering defaults.
+
+        :param min_height: str, initial height of circuit display.
```
CalMacCQ, line 102: Looks good to me, thanks :) 

I just wanted to check how the defaults are configured.

 If I use `set_render_options` in a notebook environment does this override the defaults for all other diagrams that I would render afterwards? Would restarting the jupyter kernel revert back to the default settings?

pytket/docs/display.rst
```diff
@@ -10,21 +10,49 @@ Contains several functions for rendering interactive circuit diagrams.
         pip install pytket-offline-display
 
 .. automodule:: pytket.circuit.display
-    :members: render_circuit_jupyter, render_circuit_as_html, view_browser
```
cqc-alec, line 13: Are we removing these methods from the docs?

pytket/pytket/circuit/display/__init__.py
DNA386: Yeah. It overrides the defaults for all diagrams you render afterwards using that instance (ie calling `circuit_renderer.render_circuit_jupyter`),but doesn't affect eg the default `render_circuit_jupyter` you can import directly.

pytket/docs/display.rst
DNA386: I removed them from being listing explicitly since they are the same as those found in the class documentation that now gets listed, but I can add them back in if you think its helpful

cqc-alec: > I removed them from being listing explicitly since they are the same as those found in the class documentation that now gets listed, but I can add them back in if you think its helpful

Ah I see, no that's fine.

cqc-alec (approved):

CalMacCQ (approved):

Issue #777, opened 2023-03-09
Using add_blank_wires does not preserve the Connectivity


nquetschlich:
Hi,

this issue is closely related to #770 (and #774). As mentioned in #770, I tried to compile quantum circuit to certain target devices with the goal of having a resulting quantum circuit with as many qubits as the targeted device provides. @cqc-alec suggested to use `add_blank_wires` which I have used after the actual compilation has finished (in this case, for a coupling map which is directed): 

```python
from pytket import architecture, OpType
from pytket.qasm import circuit_from_qasm_str

from pytket.passes import (
    auto_rebase_pass,
    CXMappingPass
)
from pytket.placement import GraphPlacement
from pytket.qasm import circuit_from_qasm_str

qc_tket = circuit_from_qasm_str('OPENQASM 2.0;\ninclude "qelib1.inc";\nqreg q[3];\ncreg c[2];\nh q[0];\nh q[1];\nx q[1];\nx q[2];\nh q[2];\ncx q[0],q[2];\nh q[0];\ncx q[1],q[2];\nx q[1];\nh q[1];\nbarrier q[0],q[1],q[2];\nmeasure q[0] -> c[0];\nmeasure q[1] -> c[1];\n')

cmap = [[0, 1], [0, 7], [1, 2], [2, 3], [7, 6], [6, 5], [4, 3], [4, 5]]
arch = architecture.Architecture(cmap)
native_gate_set_rebase = auto_rebase_pass({OpType.Rz, OpType.SX, OpType.X, OpType.ECR, OpType.Measure})

native_gate_set_rebase.apply(qc_tket)
placer = GraphPlacement(arch) 
CXMappingPass(arc=arch, placer=placer, directed_cx=True, delay_measures=False).apply(qc_tket)
print(qc_tket.valid_connectivity(arch, directed=True))

diff = 8 - qc_tket.n_qubits 
qc_tket.add_blank_wires(diff)

print(qc_tket.valid_connectivity(arch, directed=True))
```

While the first `print` statement returns `True` and, by that, confirm the valid connectivity of the compiled circuit, the second `print` statement returns `False` after adding the blank wires. To me, this is was unexpected and I wanted to confirm whether this behavior is intended.

nquetschlich:
> Hi,
> 
> The `valid_connectivity` method also checks whether the names of the qubits in your circuit match the nodes in the architecture.
> 
[...]

PR #840, opened 2023-04-25
Bugfix/routing segfault 777

yao-cqc:
This should address the bug raised in this https://github.com/CQCL/tket/issues/777#issuecomment-1508934186

Initial and final maps can be considered as mapping logical qubits to physical qubits. When adding an ancilla node `node[x]` during routing, the current logic adds the entry {`node[x]`, `node[x]`} to the initial and final maps. So we are essentially giving both the ancilla logical qubit and its physical mapping the same name `node[x]`. However, this is error prone since there might already exits a logical qubit with the name `node[x]`. For example, route a circuit twice.

This PR fixes this issue by naming the ancilla logical qubit `tk_ROUTING_ANCILLA_REG[i]`, and making sure this qubit was not previously present.


tket/src/Mapping/MappingFrontier.cpp
```diff
@@ -763,22 +788,39 @@ void MappingFrontier::add_bridge(
...
+
+  this->bimaps_->initial.insert({placeholder_q, ancilla});
+  this->bimaps_->final.insert({placeholder_q, ancilla});
   this->ancilla_nodes_.insert(Node(ancilla));
-  UnitID uid_ancilla(ancilla);
-
-  unit_map_t update_map;
-  update_map.insert({uid_ancilla, uid_ancilla});
-
-  update_maps(this->bimaps_, update_map, update_map);
```
yao-cqc, line 781: I'm not sure if this actually does anything.

pytket/docs/changelog.rst
```diff
@@ -24,6 +24,8 @@ Fixes:
 
 * Fix bug in `get_operator_expectation_value()` computation when operator
   includes `Pauli.I` terms.
+* Fix bug in routing when the circuit contains empty qubit wires and some of its qubits
```
sjdilkes, line 27: *Fix bug in routing code occurring in Circuits with qubit wires with no operations and some (other or same) qubits pre-labelled as Node from the Architecture being routed to.

tket/src/Mapping/MappingFrontier.cpp
sjdilkes: I think if you're directly modifying this->bimaps_->initial/final yourself then we can probably avoid this? Also if tests are passing then seems fine to remove.

sjdilkes (changes requested):
Just the minor edit to the changelog I think, code changes look good. 

Side note - have you tried running the prop tests a bunch? I appreciate the CI should capture any bugs but I found when dealing with this sort of bug originally that the proptests were helpful for occasionally generating a rogue Circuit. Not a very scientific approach, but I'd recommend quickly rerunning the proptests a bunch of times locally just to confirm nothing comes up.

yao-cqc:
> Just the minor edit to the changelog I think, code changes look good.
> 
> Side note - have you tried running the prop tests a bunch? I appreciate the CI should capture any bugs but I found when dealing with this sort of bug originally that the proptests were helpful for occasionally generating a rogue Circuit. Not a very scientific approach, but I'd recommend quickly rerunning the proptests a bunch of times locally just to confirm nothing comes up.

Yes, I have run the prop test in a for loop, but nothing has come up.

sjdilkes (approved):

PR #852, opened 2023-05-09
POC: add unit tests as optional part of tket package

trvto:
While trying to get the `dev-tools` stuff working again, I found an alternate way to setup the testing for `tket`. It is the way suggested by Conan -> Have the unit tests as an optional part of the main package.

This draft implements Conans suggestions (mainly from [here](https://docs.conan.io/2/tutorial/creating_packages/build_packages.html)).

Now `conan create` with `-c tools.build:skip_test=False` will build tket, tests and proptests and run the tests and proptests.
Similarly, `conan install` + repeated `conan build` with `-c tools.build:skip_test=False` can be used to do this iteratively in a local build directory. Seems to also work with ccache. Only changes are rebuilt (regardless whether they come from tket, tests, or proptests). It is also possible to only run tests or proptests using `conan build . -c tools.build:skip_test=True` and then running the built test executables directly from the command line.

This would require some changes to the git workflows, so they won't be passing for now.


tket/CMakeLists.txt
```diff
@@ -68,15 +73,15 @@ IF (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
...
-target_link_libraries(tket PRIVATE boost::boost)
-target_link_libraries(tket PRIVATE symengine::symengine)
-target_link_libraries(tket PRIVATE Eigen3::Eigen)
-target_link_libraries(tket PRIVATE nlohmann_json::nlohmann_json)
-target_link_libraries(tket PRIVATE tklog::tklog)
-target_link_libraries(tket PRIVATE tkassert::tkassert)
-target_link_libraries(tket PRIVATE tkrng::tkrng)
-target_link_libraries(tket PRIVATE tktokenswap::tktokenswap)
-target_link_libraries(tket PRIVATE tkwsm::tkwsm)
+target_link_libraries(tket PUBLIC boost::boost)
```
trvto, line 76: Compiling the tests wasn't working with these as `PRIVATE`. If they need to be `PRIVATE` there might be a different way to get it working though...

PR #697, opened 2022-12-22
Add support for Python 3.11 and remove support for 3.8

cqc-alec:
Closes #646 .

Will modify names of required CI checks if approved.

Release workflow tested [here](https://github.com/CQCL/tket/actions/runs/3759526258).

PR #857, opened 2023-05-11
ZX Tensor Evaluation Improvements

willsimmons1465:
Motivated by the ZX manual section (in final checks before making a PR), the code examples benefitted from the ability to evaluate any diagram as a tensor without having to convert to spiders. This PR adds the other generators to tensor evaluation, moves it into the core of pytket, and fixes some bugs that appeared during writing the example code for the manual.


aborgna-q (approved):
Nice!

There're some CI issues, like `quimb` [not being included](https://github.com/CQCL/tket/pull/697#discussion_r1060396462) in some tests,
but the rest LGTM.

aborgna-q (approved):
👍

PR #859, opened 2023-05-12
Make everything buildable without conan

### LINKED ISSUE #464: How to build tket and install into $PREFIX?
I am trying to create the FreeBSD port for tket.
README has a lot of custom ```conan```-based commands. It looks like this is to integrate it with conan.

But how to just build binaries and install them into $PREFIX?

cqc-alec:
Closes #464 .

The "lines changed" count is misleading as most of them are just inserting `tket/` into an include path. I suggest looking at the commits one by one so as not to be distracted by these changes. It's really just a few tweaks to the cmake and conan files. I've added full instructions for how to build without conan in the file `build-without-conan.md`.

This won't replace the "official" builds using conan. Still it might be worth adding a scheduled workflow on the CI to make sure the method continues to work, as it could be useful for developers or packagers.


trvto:
Installing gmp from source insn't working for me on my arm64 mac. Not sure why. But if it's working for you then the rest of the cmake only build seems ok.

cqc-alec:
> Installing gmp from source insn't working for me on my arm64 mac. Not sure why. But if it's working for you then the rest of the cmake only build seems ok.

Yeah works for me -- you may not need it though, if gmp is already installed in a system directory that will be found by `find_package()`. (If the symengine build succeeds then it probably is.)

cqc-alec:
I fixed a few mistakes in the documentation, and added a workflow script to check (weekly) that everything works on stock Ubuntu at least. Tested [here](https://github.com/CQCL/tket/actions/runs/4971487929).

trvto (changes requested):

trvto (approved):
Looks good!

PR #864, opened 2023-05-15
require autoray>=0.6.3 in tests

yao-cqc:
https://github.com/CQCL/tket/actions/runs/4978898957/jobs/8909810253

seems to be caused by using an older version of autoray


willsimmons1465:
I just spotted this when fixing up #857. I think autoray v 0.6.1 is enough to get it to work. As #857 is already moving these requirements to extras_require in setup.py, I will add the specification there. It also looks like quimb now works with python 3.11 so I'll be removing that specification as well

yao-cqc:
@willsimmons1465 ahh cool, I'm closing this one. Thanks!

PR #869, opened 2023-05-19
Add `with_initial_reset` option to `StatePreparationBox`

cqc-alec:


pytket/binders/circuit/boxes.cpp
```diff
@@ -515,12 +515,15 @@ void init_boxes(py::module &m) {
...
+          py::init<const Eigen::VectorXcd &, bool, bool>(),
           "Construct from a statevector\n\n"
-          ":param statevector: normalised statevector\n",
+          ":param statevector: normalised statevector\n"
           ":param is_inverse: whether to implement the dagger of the state "
-          "preparation circuit, default to false",
-          py::arg("statevector"), py::arg("is_inverse") = false)
+          "preparation circuit, default to false\n"
+          ":param with_initial_reset: whether to explicitly set the state to "
+          "zero initially (otherwise the initial zero state is assumed)",
```
sjdilkes, line 524: Might be worth adding a further explanation to also cover:
```(by default it is assumed that the initial state is zero and no explicit  reset is applied)```

File: tket/src/Circuit/macro_manipulation.cpp [hunk 2]
```diff
@@ -248,16 +249,30 @@ void Circuit::substitute(
                  // trivial, essentially rewiring on a cut
   std::map<Edge, Vertex> c_out_map;
 
+  std::set<Qubit> reset_qbs;
+  for (auto qb : to_insert.all_qubits()) {
```
sjdilkes, line 253: `const Qubit& qb` - or is there a reason to not pass by constant reference/reference here?

sjdilkes (changes requested):

pytket/binders/circuit/boxes.cpp
cqc-alec: Done.

tket/src/Circuit/macro_manipulation.cpp
cqc-alec: Done.

sjdilkes (approved):
LGTM

PR #872, opened 2023-05-25
Refactor gen sythesise pauli graph

trvto:
Add new pass PauliExponentials that generates PauliExp type boxes
The gen synthesise pauli graph pass now uses pauli exponentials + decompose boxes


sjdilkes (changes requested):
Looks great! My only requested changes are minor doc changes, typos and the renaming of two methods in the new Box classes.

sjdilkes (approved):
Looks great!

PR #873, opened 2023-05-26
Add workflow dispatch to workflows

trvto:
- allows to trigger workflow from the github UI on any branch (without needing to open a PR and run other workflows)
- I left out issue.yml,  didn't seem necessary...
- Can't actually trigger manually until this is merged I think
- Drive-By: Pin Conan version to 2.0.5 or lower, Conan 2.0.6 breaks pytket installs (with pip) and the `rootpath` script

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -11,7 +11,7 @@ on:
   schedule:
     # 03:00 every Saturday morning
     - cron: '0 3 * * 6'
-  workflow_dispatch:
+  workflow_dispatch: {}
 
 concurrency:
   group: ${{ github.head_ref || github.run_id }}
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -11,6 +11,7 @@ on:
   schedule:
     # 03:00 every Saturday morning
     - cron: '0 3 * * 6'
+  workflow_dispatch:
 
 concurrency:
   group: ${{ github.head_ref || github.run_id }}
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -450,7 +450,7 @@ jobs:
         eval "$(pyenv init -)"
         pyenv shell tket-3.9
         PKG_CONFIG_PATH="$(brew --prefix openblas)"/lib/pkgconfig pip install -U scipy
-        pip install -U conan
+        pip install conan==${{ env.conan_version }}
         conan remove -c "pybind11/*"
         conan remove -c "pytket/*"
         conan create recipes/pybind11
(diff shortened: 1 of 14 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -11,6 +11,7 @@ on:
   schedule:
     # 03:00 every Saturday morning
     - cron: '0 3 * * 6'
+  workflow_dispatch:
```
cqc-alec, line 14: A lot of the steps below are triggered depending on `github.event_name`. You probably want to run the same checks as if it were a pull request, so I'd suggest looking for `github.event_name == 'pull_request'` and adding `workflow_dispatch` as an alternative trigger. (Similarly in some of the other workflows.)

cqc-alec (changes requested):

.github/workflows/build_and_test.yml
trvto: Do I need to do the "Check tket version bump" step for a manual trigger? As long as "Upload package" doesn't run, this isn't needed, right?

cqc-alec: A package is never uploaded on a pull-request workflow run, so I guess it's not necessary to do this check.

.github/workflows/build_and_test.yml
```diff
@@ -410,7 +411,7 @@ jobs:
     name: Build and test pytket (macos-arm64)
     runs-on: ['self-hosted', 'macOS', 'ARM64']
     needs: check_changes
-    if: github.event_name == 'push' || github.event_name == 'schedule' || github.event.pull_request.head.repo.full_name == github.repository
+    if: github.event_name == 'push' || github.event_name == 'schedule' || github.event.pull_request.head.repo.full_name == github.repository || github.event.workflow_dispatch.repository == github.repository
```
trvto, line 414: This is possible according to the github reference, but not sure if it's necessary. Maybe `github.event_name == 'workflow_dispatch'` would be enough the workflow dispatch case? Not 100% why it's necessary for the 'pull_request" case...

cqc-alec: It's because we don't want PRs from forks to run on the self-hosted Mac M1 (a bit of paranoia).

.github/workflows/packages.yml
```diff
@@ -7,6 +7,7 @@ on:
   push:
     branches:
       - 'packages-upload-new/**'
+  workflow_dispatch:
```
cqc-alec, line 10: I don't think we want this -- this is quite a special-purpose workflow that uploads packages to the artifactory.

.github/workflows/docs.yml
```diff
@@ -6,6 +6,7 @@ on:
       - develop
     paths:
       - 'tket/src/**'
+  workflow_dispatch:
```
cqc-alec, line 9: Probably don't want this either (publishes docs on push to develop).

cqc-alec (approved):

PR #879, opened 2023-05-31
Unpin from conan 2.0.5 and fix parsing scripts

### LINKED ISSUE #874: Update uses of `conan graph info xxxx --format json` for compatibility with conan 2.0.6

cqc-alec:
Closes #874 .


cqc-alec:
Just noticed I missed the pytket build...

cqc-alec:
> Just noticed I missed the pytket build...

Fixed this. By a coincidence the Python used to analyse the json in `pytket/setup.py` works for both the old and the new versions (which would explain why the build succeeded).

trvto:
> > Just noticed I missed the pytket build...
> 
> Fixed this. By a coincidence the Python used to analyse the json in `pytket/setup.py` works for both the old and the new versions (which would explain why the build succeeded).

Ah ok. For me on Friday pytket builds were failing without 2.0.6, but I didn't spend much time investigating because pinning to 2.0.5 fixed it

cqc-alec:
> Edit: Seems like they still are on the CI at least

Yes, oops -- will fix...

cqc-alec:
Should be fixed now!

trvto (approved):

PR #884, opened 2023-06-06
Solve: Assertion failure when adding gate to bit

cqc-melf:


cqc-melf:
I think this does not need to go into the changelog?

cqc-alec (approved):

PR #886, opened 2023-06-06
Correct Tableau semantics

willsimmons1465:
I noticed our existing tests for various kinds of stabilizer tableau tested for internal consistency wrt conversions but not actual objective tests against intended semantics. There were some differences in the way phases are handled that I wanted to clear up to make sure users who are expecting it to match the literature aren't disappointed. I also finally switched out the `CliffTableau` class with a new `UnitaryRevTableau` which builds on top of `SymplecticTableau` and so supports better interop with `UnitaryTableau` and `ChoiMixTableau`.


tket/include/tket/Clifford/UnitaryTableau.hpp
```diff
@@ -169,4 +172,130 @@ JSON_DECL(UnitaryTableau)
 
 std::ostream& operator<<(std::ostream& os, const UnitaryTableau& tab);
 
+class UnitaryRevTableau {
+  /**
+   * Whereas UnitaryTableau rows are over the output segment with fixed single
+   * paulis over the input segment, UnitaryRevTableau flips this around with
+   * each row representing a Pauli string over the input segment and a fixed
+   * signel pauli over the output segment. That is, the ith X row is the
```
cqc-alec, line 180:
```suggestion
   * single pauli over the output segment. That is, the ith X row is the
```

tket/include/tket/Clifford/UnitaryTableau.hpp
```diff
@@ -169,4 +172,130 @@ JSON_DECL(UnitaryTableau)
...
+   * (phaseful) Pauli string P such that X_i C P = C, and similarly for the Z
+   * rows.
+   *
+   * This is maintained by storing a UnitaryTableau for the inverse unitary
+   * C^dagger. Since each Pauli string is Hermitian, X_i C P = C is encoded as
+   * C^dagger = (X_i C P)^dagger = X_i C^dagger P, so no need to modify rows on
```
cqc-alec, line 186:
```suggestion
   * C^dagger = (X_i C P)^dagger = P C^dagger X_i, so no need to modify rows on
```
Is it the same?

cqc-alec (approved):

tket/src/PauliGraph/PauliGraph.cpp
```diff
@@ -113,13 +113,12 @@ void PauliGraph::apply_gate_at_end(
     case OpType::CY:
     case OpType::CZ:
     case OpType::SWAP:
-    case OpType::noop:
-    case OpType::Phase: {
```
cqc-alec, line 117: Can we still handle the global-phase op?

willsimmons1465: It isn't listed in the accepted gates in the preconditions for the PauliGraph passes, and they currently do not preserve global phase so I don't see much value in trying to handle them at this point. I could allow them to be treated as a noop, but the fact that a programmer is using a Phase gate explicitly tells me that they are interested in determining semantics with global phase, so silently removing it without error or warning could be frustrating for them.

willsimmons1465: Ok, I have added Phase gate handling back in. Did not know it was implicitly allowed by GateSetPredicate, but that makes enough sense for it to be handled

cqc-alec (approved):

PR #887, opened 2023-06-06
Move tket tests inside tket package

trvto:
- Unit tests and proptests are now an optional part of the tket package
  - build both using `-o with_all_tests=True` in the conan commands
  - build just unit tests using `-o with_test=True`
  - build just proptests using `-o with_proptest=True`
- Also:
  - the logic in the `build_and_test` workflow has changed a bit. @cqc-alec please verify that it makes sense. 
  - fixed bug in the cleaning of windows line endings (cmake/* wasn't being cleaned, now everything in tket/* is)
  - ccache working again for almost all tket builds on the CI (with warm cache builds now ca 1 min instead of 20-40 min)
    - only exception: Windows build as shared library (windows with ccache is easier (maybe only possible) with ninja but ninja build of shared library fails and I don't know why...
    - Caveat: If any dependency needs to be built from source, the way conan works will trip up ccache and will cause a complete rebuild, so need to keep remote conan repo (tket-libs) up-to-date!
    - Removed --build=missing from most builds so that a missing dependency will be apparent (just upload to tket-libs if necessary)
    - Added build of shared tklog in `build_libs` workflow
    - Added Debug build in `packages` workflow
    - Put priority of tket-libs before conan center in all workflows to circumvent build from source when conan package is found but does not contain binary


.github/workflows/build_and_test.yml
```diff
@@ -56,36 +53,51 @@ jobs:
...
+          brew install clang-format@16
+          git ls-files "*.cpp" "*.hpp" | xargs clang-format -style=file --dry-run --Werror
+
   build_test_tket:
     name: Build and test (tket)
     needs: check_changes
+    if: needs.check_changes.outputs.tket_changed == 'true'
     strategy:
       matrix:
-        os: ['ubuntu-22.04', 'macos-12', 'windows-2022']
```
cqc-alec, line 71: Is the windows build so different that it is worth separating, rather than making some steps conditional?

.github/workflows/build_and_test.yml
```diff
@@ -96,97 +108,148 @@ jobs:
...
+      matrix:
+        os: ['windows-2022']
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v3
+      - name: Select Python 3.10
+        # otherwise turtlebrowser/get-conan@v1.2 fails on macos-12
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
```
cqc-alec, line 152: This probably isn't necessary on Windows. (Not even sure it's still an issue on MacOS.)

.github/workflows/build_and_test.yml
trvto: Yes, unfortunately, it's completely different. The ccache cache setup has to be done by hand (the ccache action doesn't quite work for windows). Could theoretically do it by hand for the others, but then the commands will be different for bash vs. powershell anyway. Also, the ccache`base_dir` setting doesn't work on windows so have to use conan `install` + `build` for ccache to work.

.github/workflows/coverage.yml
```diff
@@ -37,40 +37,38 @@ jobs:
     name: Generate coverage report
     needs: changes
     if: needs.changes.outputs.tket == 'true' || github.event_name == 'schedule'
-    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        os: ["ubuntu-22.04"]
+    runs-on: ${{ matrix.os }}
```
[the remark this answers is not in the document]
cqc-alec: I wouldn't do it just for the sake of it but if we need to refer to the name somewhere in the steps then it's probably worth it.

trvto:
I've update based on the comments. Couldn't get ccache working for the dynamic build on windows, so that will have to wait.

.github/workflows/build_and_test.yml
```diff
@@ -96,97 +108,143 @@ jobs:
...
+            [IO.File]::WriteAllText($f, $normalized_file)
+          }
+      - name: ninja ccache setup
+        # This is what hendrikmuhs/ccache-action does, but for windows (theirs is buggy on windows)
+        id: ninja-ccache-setup
+        run: |
+          choco upgrade ccache ninja
+          $ccache_dir = ccache --get-config cache_dir
+          echo ccache_dir=$ccache_dir >> $env:GITHUB_OUTPUT
+          ccache --set-config=max_size='500M'
```
cqc-alec, line 172: Why are there still these ccache steps on Windows if we are not using it?

.github/workflows/coverage.yml
```diff
@@ -37,40 +37,38 @@ jobs:
...
+        conan build tket --user=tket --channel=stable -s build_type=Debug -o boost/*:header_only=True -o tket/*:profile_coverage=True  -o with_test=True -o test-tket/*:with_coverage=True -of build/tket
+        conan export-pkg tket --user=tket --channel=stable -s build_type=Debug -o boost/*:header_only=True -o tket/*:profile_coverage=True -o with_test=True -of build/tket -tf ""
     - name: Install runtime test requirements
       if: github.event_name == 'schedule'
       run: |
         sudo apt-get update
         sudo apt-get install texlive texlive-latex-extra latexmk
         mkdir -p ~/texmf/tex/latex
         wget http://mirrors.ctan.org/graphics/pgf/contrib/quantikz/tikzlibraryquantikz.code.tex -P ~/texmf/tex/latex
-    - name: Run tket tests
```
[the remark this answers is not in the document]
cqc-alec: ?

.github/workflows/build_and_test.yml
cqc-alec: Ah I see, OK.

.github/workflows/coverage.yml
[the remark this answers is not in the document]
trvto: The default tests are run automatically during the build step

cqc-alec:
LGTM; just going to run a few local tests before approving...

cqc-alec (approved):
Thanks!

PR #896, opened 2023-06-21
Generate documentation for enums

cqc-alec:
Make sphinx generate documentation for Python Enum values. It is not currently doing so.

Before:

image

After:

image


CalMacCQ (approved):
Looks good. I think we will need to do this separately on the extensions repos as well.

cqc-alec:
> I think we will need to do this separately on the extensions repos as well.

Yes, I was going to make another PR for pytket-quantinuum (I noticed that the values of the `Language` enum weren't documented), but we should do the same for all extensions, at least those that have `Enum` definitions.

PR #903, opened 2023-06-28
Feature/pauli graph commute synth

yao-cqc:


tket/src/Converters/PauliGraphConverters.cpp
```diff
@@ -176,4 +195,116 @@ Circuit pauli_graph_to_pauli_exp_box_circuit_sets(
...
+      QubitOperator new_gadgets;
+      QubitOperator &old_gadgets = commuting_gagdets[j];
+      for (auto it = old_gadgets.begin(); it != old_gadgets.end(); it++) {
+        QubitPauliTensor new_qpt = tab.get_row_product(it->first);
+        new_gadgets.insert({new_qpt, it->second});
+      }
+      commuting_gagdets[j] = new_gadgets;
+    }
+    // 5. update the final tableau
+    final_tab = UnitaryRevTableau::compose(tab, final_tab);
```
willsimmons1465, line 298: An alternative to this is to instead maintain just an accumulated tableau of the Clifford so far and only conjugate the remaining gadgets at the point of synthesising them. Then just compose the two tableaus at the end of synthesis. This might yield some slight complexity gains without changing the actual resulting circuit at all

yao-cqc: Thanks! That's a good point.

PR #905, opened 2023-07-03
Implement `get_unitary()` for boxes

### LINKED ISSUE #895: Implement `get_unitary()` for all box types
Default case would fall back to `get_circuit().get_unitary()`. Sometimes we can do better, e.g. for `QControlBox` (see #866 ).

cqc-alec:
Closes #895 .

Moved the `Simulation` directory inside the `Circuit` directory, since it is no longer an independent component.


cqc-alec:
The C++ formatting check is failing because of brew flakiness, will retry later but have checked formatting locally and happy to merge without this check.

sjdilkes (approved):

PR #918, opened 2023-07-07
Add nix support

jake-arkinstall-quantinuum:
Not yet ready for review. Aside from possible design changes, documentation and test automation need to be added too. Testing on MacOS is also required. Windows support is not generally available in Nix at this stage (to my knowledge).

Nix is an alternative, hermetic way of managing packages. With this PR, I've added support for building within Nix.

The `nix.flake` file specifies the core nix dependencies. It also specifies where the build instructions can be found - these are bundled into a new `nix-support` directory.

The structure of `nix-support` avoids having nix files scattered throughout the project. The package files are:
- `symengine.nix`: provides a derivation of SymEngine built to suit the needs of this project. Specifically, through `symengine.patch`, dependencies are linked as required. Additionally, symengine.nix provides both static and shared library versions of SymEngine.
- `third-party-python-packages.nix` - provides packages missing from nixpkgs. These are:
  - pybind11_json
  - qwasm
  - lark-parser
  - types-pkg_resources
- `libs.nix` - provides the following as static and shared derivations:
  - tklog
  - tkrng
  - tkassert
  - tktokenswap
  - tkswm
 - `tket.nix` - provides tket, as both static and shared derivations
 - `pytket.nix` - provides the binders and pytket, using only the shared derivations.
 
 An additional change is to `setup.py`, which now has a nix-specific build extension. This uses the environment variables provided by nix to find tklog, tket, and binders. CMakeLists.txt for tket is now also told *not* to error on -wmaybe-uninitialized, because of a false positive that crops up in GCC 12.2. If this isn't wanted for general builds, this flag can be added as a patch for nix-specific builds.
 
These derivations are all provided through fixed point combinators, so that flake.nix can use them as overloads to pkgs.
 
Additional changes that are required for compilation, though not specific enough for this PR, are:
- use of size_t in the codebase. The relevant <c....> headers only guarantee that `std::size_t` is provided, not C's `size_t`. Most implementations accept this, but GCC 12.2 does not.
- Missing break statements in `tket/include/tket/ArchAwareSynth/SteinerTree.hpp`. Although an assert is fired on default paths, indirectly blocking fallthrough, the compiler can't always detect this, and warns on possible fallthrough. This causes a failed build due to `-Werror`.

Example usage (replace the github organisation once merged, and this requires the above additional changes to run) is below. Note that the machine does not need any dependencies, including python or GCC, installed in advance. It takes some time to compile - and gives my threadripper a run for its money, but promises a reliable build.

```
$ nix develop --refresh github:jake-arkinstall-quantinuum/tket?ref=nix-restructure

$ python3
Python 3.10.11 (main, Apr  4 2023, 22:10:32) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pytket
>>> foo = pytket.Circuit(2)
>>> foo.X(0)
[X q[0]; ]
>>> foo.H(1)
[X q[0]; H q[1]; ]
>>>
```


tket/test/src/TokenSwapping/test_BestTsaFixedSwapSequences.cpp
```diff
@@ -15,6 +15,8 @@
 #include <algorithm>
 #include <catch2/catch_test_macros.hpp>
 
+#include <algorithm>
+
```
cqc-alec, line 19:
```suggestion
#include <algorithm>
#include <catch2/catch_test_macros.hpp>

```

pytket/setup.py
```diff
@@ -128,10 +128,40 @@ def run(self):
                 if not os.path.isdir(libpath):
                     shutil.copy(libpath, extdir)
 
+class NixBuild(build_ext):
```
cqc-alec, line 131:
```suggestion


class NixBuild(build_ext):
```

pytket/setup.py
```diff
@@ -128,10 +128,40 @@ def run(self):
...
+        assert location_tklog and location_tket and location_binders
+        for location in [location_tklog[0],
+                         location_tket[0],
+                         location_binders[0]]:
+            for lib in os.listdir(location):
+                libpath = os.path.join(location, lib)
+                if not os.path.isdir(libpath):
+                    shutil.copy(libpath, extdir)
 
 setup_dir = os.path.abspath(os.path.dirname(__file__))
```
cqc-alec, line 155:
```suggestion
                    shutil.copy(libpath, extdir)


setup_dir = os.path.abspath(os.path.dirname(__file__))
```

pytket/setup.py
```diff
@@ -128,10 +128,40 @@ def run(self):
...
 
+def get_build_ext():
+    if os.getenv("USE_NIX"):
+        return NixBuild
+    elif os.getenv("NO_CONAN"):
+        return CMakeBuild
+    else:
+        return ConanBuild
 
 class bdist_wheel(_bdist_wheel):
```
cqc-alec, line 166:
```suggestion
        return ConanBuild


class bdist_wheel(_bdist_wheel):
```

pytket/setup.py
```diff
@@ -129,9 +129,42 @@ def run(self):
...
+            for lib in os.listdir(location):
+                libpath = os.path.join(location, lib)
+                if not os.path.isdir(libpath):
+                    shutil.copy(libpath, extdir)
+
+
 setup_dir = os.path.abspath(os.path.dirname(__file__))
 plat_name = os.getenv("WHEEL_PLAT_NAME")
 
+def get_build_ext():
```
cqc-alec, line 160:
```suggestion
plat_name = os.getenv("WHEEL_PLAT_NAME")


def get_build_ext():
```

pytket/setup.py
```diff
@@ -129,10 +129,44 @@ def run(self):
...
+
+        nix_ldflags = os.environ['NIX_LDFLAGS'].split()
+        build_inputs = os.environ['propagatedBuildInputs'].split()
+        location_tklog = [l[2:] for l in nix_ldflags if '-tklog' in l]
+        location_tket = [l[2:] for l in nix_ldflags if '-tket' in l]
+        location_binders = [f"{l}/lib" for l in build_inputs if '-binders' in l]
+        assert location_tklog and location_tket and location_binders
+        for location in [location_tklog[0],
+                         location_tket[0],
+                         location_binders[0]]:
```
cqc-alec, line 150:
```suggestion
        nix_ldflags = os.environ["NIX_LDFLAGS"].split()
        build_inputs = os.environ["propagatedBuildInputs"].split()
        location_tklog = [l[2:] for l in nix_ldflags if "-tklog" in l]
        location_tket = [l[2:] for l in nix_ldflags if "-tket" in l]
        location_binders = [f"{l}/lib" for l in build_inputs if "-binders" in l]
        assert location_tklog and location_tket and location_binders
        for location in [location_tklog[0], location_tket[0], location_binders[0]]:
```

cqc-alec:
Looks good, I think this just needs:
* editing the workflow script to run on both `ubuntu-22.04` and `macos-12` (using the `matrix` method, see other workflow scripts for examples);
* adding a paragraph to the top-level `README.md`, probably just before the "API documentation" section.

jake-arkinstall-quantinuum:
I've now added python tests, so `nix flake check` is (I think) complete.

> Looks good, I think this just needs:
> 
> * editing the workflow script to run on both `ubuntu-22.04` and `macos-12` (using the `matrix` method, see other workflow scripts for examples);

I'll look into that now. 

> * adding a paragraph to the top-level `README.md`, probably just before the "API documentation" section.

I've now added this. It points to a README.md in the nix-support directory. I've also added an example usage flake within there, and documented that (lightly) to provide an example of what simple usage might look like.

jake-arkinstall-quantinuum:
Test matrix now added. This will be the first time this approach has been tried on MacOS. I'll run it on my macbook first and diagnose any issues that may arise, then once that's done I'm happy to promote this from a draft PR.

jake-arkinstall-quantinuum:
All tests passing on MacOS now. It was surprisingly awkward due to dynamic_casts failing, and it seems to be thanks to dylib's two-level namespaces (identified thanks to https://developer.apple.com/forums/thread/693914).

I want to verify my changes on Linux again, do some tidying around the edges, then I'll promote this to a full PR.

nix-support/tket.nix
```diff
@@ -0,0 +1,44 @@
...
+      cp -r ${../tket/CMakeLists.txt} $out/CMakeLists.txt;
+    '';
+  };
+in {
+  tket = super.stdenv.mkDerivation {
+    name = "tket";
+    src = tket-without-tests;
+    nativeBuildInputs = [ super.cmake ];
+    propagatedBuildInputs = super.tklibs
+      ++ [ super.boost super.symengine super.eigen super.nlohmann_json ];
```
cqc-alec, line 23: Do we specify the versions of these libraries anywhere or does nix just pick up the latest available?

nix-support/third-party-python-packages.nix
```diff
@@ -0,0 +1,44 @@
...
+    };
+    nativeBuildInputs = [ super.cmake ];
+    buildInputs = [ super.python3Packages.pybind11 super.nlohmann_json ];
+  };
+  qwasm = super.python3.pkgs.buildPythonPackage {
+    name = "qwasm";
+    src = super.fetchFromGitHub {
+      owner = "CQCL";
+      repo = "qwasm";
+      rev = "35ebe1e2551449d97b9948a600f8d2e4d7474df6";
```
cqc-alec, line 18: There's likely to be a certain amount of bitrot here as dependency versions get updated; maybe we could prevent this with some ugly CI checks but probably not worth it for now.

cqc-alec (approved):
LGTM!

nix-support/tket.nix
jake-arkinstall-quantinuum: The versions of the packages are defined by the nixpkgs commit that the flake.lock file is pinned to.

The typical way of bumping the packages when they're available in a newer version of nixpkgs is to run "nix flake update", which updates the lock file.

If a version is needed that isn't in nixpkgs yet, or if the version in nixpkgs is too new, or if something needs to be tweaked, the derivation of the dependency can be defined in a nix file within tket (this is what I did with jsonschema and SymEngine, for example).

PR #921, opened 2023-07-07
Implement `get_box_unitary()` for `ToffoliBox`

### LINKED ISSUE #914: Implement `get_box_unitary()` for `ToffoliBox`
Easy to construct from the permutation, and should be much faster than the fallback to circuit decomposition.

cqc-alec:
Closes #914 .

Also some small refactoring, to store the number of qubits (and its exponential) on construction.

Realized that the `Box.get_box_unitary()` methods added in #905 weren't actually being called. Oops. Fixed this.


yao-cqc (dismissed):
Looks good!

cqc-alec:
Just bumped tket version again after merge.

PR #927, opened 2023-07-11
Add warning about MacOS and conda to package documentation.

cqc-alec:
(This appears on the pypi page. Is there anywhere else we should put this?)

### DIFF
--- a/pytket/package.md
+++ b/pytket/package.md
@@ -18,6 +18,11 @@ To install the pytket extension modules add a hyphen and the extension name to t
 
 For a list of pytket extensions see this page: https://cqcl.github.io/pytket-extensions/api/index.html.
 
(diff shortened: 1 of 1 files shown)


CalMacCQ:
Thanks.

The only other place I can think of is to mention it in "installation troubleshooting"

https://cqcl.github.io/tket/pytket/api/install.html

cqc-alec:
> Thanks.
> 
> The only other place I can think of is to mention it in "installation troubleshooting"
> 
> https://cqcl.github.io/tket/pytket/api/install.html



> Thanks.
> 
> The only other place I can think of is to mention it in "installation troubleshooting"
> 
> https://cqcl.github.io/tket/pytket/api/install.html

Yes, will add it there too.

CalMacCQ (dismissed):

CalMacCQ (approved):

PR #930, opened 2023-07-11
add render option, remove sleep

DNA386:
Revert added file dump for circuits on default render. Not sure how the actual issue it was intended to fix was fixed, but I couldn't reproduce the problem (rendering large circuits having truncated iframe srcdoc).

Also add in system theme option to the renderer defaults.


CalMacCQ (approved):
Looks good to me, thanks :)

PR #934, opened 2023-07-12
Release/1.17.1

cqc-alec:


pytket/tests/requirements.txt
```diff
@@ -6,5 +6,5 @@ pytest-benchmark
 py
 hypothesis
 docker
-jsonschema
+jsonschema==4.18.0
```
cqc-melf, line 9: Is this change pushed on purpose?

cqc-alec: Yes, we need this, 4.18.1 seems to have a bug that causes test failures :-(

cqc-melf (approved):

PR #942, opened 2023-07-18
Feature/depth 2q gates

### LINKED ISSUE #827: Add simple way in pytket to get 2-qubit gate depth
This can be done by explicitly passing all the 2-qubit OpTypes your circuit might contain to `Circuit.depth_by_type()`, but it would be useful to have a dedicated method. Alternatively (or as well) we could expose a method to return all two-qubit OpTypes (which already exists in the C++).

yao-cqc:
resolves #827

### DIFF
--- a/tket/src/Circuit/macro_circ_info.cpp
+++ b/tket/src/Circuit/macro_circ_info.cpp
@@ -708,10 +708,10 @@ unsigned Circuit::depth_by_types(const OpTypeSet& _types) const {
   return count;
 }
 
-unsigned Circuit::depth_by_2q_gates() const {
+unsigned Circuit::depth_2q() const {
   unsigned count = 0;
   std::function<bool(Op_ptr)> skip_func = [&](Op_ptr op) {
-    return (op->n_qubits() != 2 || !is_gate_type(op->get_type()));
+    return (op->n_qubits() != 2 || op->get_type() == OpType::Barrier);
   };
   Circuit::SliceIterator slice_iter(*this, skip_func);
   if (!(*slice_iter).empty()) count++;
--- a/tket/src/Circuit/macro_circ_info.cpp
+++ b/tket/src/Circuit/macro_circ_info.cpp
@@ -708,6 +708,21 @@ unsigned Circuit::depth_by_types(const OpTypeSet& _types) const {
   return count;
 }
 
+unsigned Circuit::depth_by_2q_gates() const {
+  unsigned count = 0;
+  std::function<bool(Op_ptr)> skip_func = [&](Op_ptr op) {
+    return (op->n_qubits() != 2 || !is_gate_type(op->get_type()));
+  };
+  Circuit::SliceIterator slice_iter(*this, skip_func);
+  if (!(*slice_iter).empty()) count++;
+  while (!slice_iter.finished()) {
+    slice_iter.cut_ = this->next_cut(
+        slice_iter.cut_.u_frontier, slice_iter.cut_.b_frontier, skip_func);
+    if (!(*slice_iter).empty()) count++;
+  }
+  return count;
+}
+
 std::map<Vertex, unit_set_t> Circuit::vertex_unit_map() const {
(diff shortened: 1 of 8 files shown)


tket/src/Circuit/macro_circ_info.cpp
```diff
@@ -708,6 +708,21 @@ unsigned Circuit::depth_by_types(const OpTypeSet& _types) const {
   return count;
 }
 
+unsigned Circuit::depth_by_2q_gates() const {
+  unsigned count = 0;
+  std::function<bool(Op_ptr)> skip_func = [&](Op_ptr op) {
+    return (op->n_qubits() != 2 || !is_gate_type(op->get_type()));
```
cqc-melf, line 714: Is `!is_gate_type(op->get_type()));` excluding conditionals and boxes or should they work fine?

tket/test/src/Circuit/test_Circ.cpp
```diff
@@ -1326,6 +1326,41 @@ SCENARIO("Test depth_by_type method") {
...
 }
+SCENARIO("Test depth_by_2q_gates method") {
+  GIVEN("2q OpTypes") {
+    Circuit circ(3, 1);
+    circ.add_op<unsigned>(OpType::CX, {0, 1});
+    circ.add_op<unsigned>(OpType::Z, {0});
+    circ.add_op<unsigned>(OpType::Z, {1});
+    circ.add_op<unsigned>(OpType::CZ, {1, 0});
+    circ.add_conditional_gate<unsigned>(OpType::CY, {}, {1, 2}, {0}, 0);
+    REQUIRE(circ.depth_by_2q_gates() == 2);
```
cqc-melf, line 1337: Maybe I understand this check wrong, but I would expect a depth of 3 because CX, CZ, CY all act on qubit 1?

tket/test/src/Circuit/test_Circ.cpp
```diff
@@ -1326,6 +1326,41 @@ SCENARIO("Test depth_by_type method") {
...
+  }
+  GIVEN("Boxes") {
+    Circuit circ(3);
+    Circuit inner(2);
+    inner.add_op<unsigned>(OpType::CX, {0, 1});
+    CircBox cbox(inner);
+    circ.add_box(cbox, {0, 1});
+    circ.add_op<unsigned>(OpType::Z, {1});
+    REQUIRE(circ.depth_by_2q_gates() == 0);
```
cqc-melf, line 1346: This is also not counting the gate in the box, is that expected?

tket/src/Circuit/macro_circ_info.cpp
yao-cqc: Yes, boxes and conditionals are not counted, but I'm not certain about this. The existing `depth_by_type` and `depth_by_types` methods don't count the conditional version of the given gate types, so I'm not sure if I should keep this behavior.

I named this method `depth_by_2q_gates` with the intention of not counting boxes. I'm not sure if boxes are considered "gates" since they can contain anything (e.g., measurements, conditionals, barriers).

My rationale is that people want a method that is equivalent to passing all 2-qubit gate types to `depth_by_type`, and this method achieves that while also recognising gates such as CnX.

Perhaps I can name this method something like `depth_by_2q_vertices` and count every vertex that has 2 qubits, as it might be less confusing?

I'm open to suggestions.

cqc-melf: I would have expected that boxes and conditionals that only interact with two qubits gates are counted as well. But I am not sure if that is what other people would expect.

Is this implemented based on a user request?

Maybe we can check this with them or add both in two functions and point out the differences in the documentation?

I think my preferred solution (if there is not further feedback) would be to include conditionals and boxes (and maybe rename the function)

yao-cqc:
I have updated the method to count all vertices with 2-legs excluding barriers.

cqc-melf (approved):

cqc-melf (approved):

PR #945, opened 2023-07-25
Fix phase poly box custom registers

### LINKED ISSUE #902: PhasePolyBox doesn't work well with custom registers
```python
c = Circuit()
areg = c.add_q_register("a",2)
breg = c.add_q_register("b",1)
c.CX(areg[0], breg[0]).CX(areg[1], areg[0])
ComposePhasePolyBoxes().apply(c)
```
the last line raises
`RuntimeError: Circuit does not contain unit with id: q[0]`

Also,

```python
c = Circuit()
areg = c.add_q_register("a",2)
breg = c.add_q_register("b",1)
c.CX(areg[0], breg[0]).CX(areg[1], areg[0])
ppbox = PhasePolyBox(c)

d = Circuit(3)
d.add_phasepolybox(ppbox, [0,1,2])
DecomposeBoxes().apply(d)
```
raises
`RuntimeError: Function only allowed for simple circuits (single register)`

I think the second error has something to do with [this line](https://github.com/CQCL/tket/blob/0b83762494e55df121f6613791e540b017d02d35/tket/src/Converters/PhasePoly.cpp#L280C1-L280C1)

trvto:
fixes #902 

Fixes two bugs leading to errors when using custom registers in combination with PhasePolyBoxes:
1. A temporary internal circuit used within `CircToPhasePolyConversion::convert` was being addressed incorrectly
2. PhasePolyBoxes hold internally a map to external qubits (which is either passed in at construction or determined from the circuit the box is constructed from)
    - This map was being used to rename the qubits in the circuit generated by the `generate_circuit` method, but the decompose_boxes pass expects the generated circuit to only contain default registers (the external qubit info is contained in how the box is wired into the circuit). This causes errors if the internal map uses custom registers.
    - However, some of the AAS code (`aas::phase_poly_synthesis`, `steiner_forest.cpp`) expected the mapping to be used and fails if it isn't.
    - Since all other boxes don't contain such info, the best course of action would probably be to refactor the PhasePolyBox code to not have the internal mapping and the AAS code to not rely on it. (I tried to do this quickly but couldn't figure out where all the dependencies where in the affected AAS code. It's not the prettiest).
    - As a quick fix. I have changed the `generate_circuit` method so that it no longer respects the mapping. This gets decompose boxes working again. I added a method `generate_circuit_with_original_placement` that has the same effect as the old `generate_circuit` method and is now used within the affected AAS code.
    - Added some tests to check that the situations from the issue now work as expected


sjdilkes (dismissed):
Sorry for super slow review, LGTM.

sjdilkes (approved):

PR #947, opened 2023-07-26
Update `auto_rebase_pass` to take optional `allow_swaps` argument

sjdilkes:
If `allow_swaps` is true, then when targeting `CX` or `ZZMax` gates, `auto_rebase_pass` will decompose `TK2` gates with wire swaps added if it will reduce the number of `CX`/`ZZMax` gates in the decomposition.

To do this I've added two new methods to CircPool.hpp and the library.cpp binders: `TK2_using_CX_and_SWAP` and `TK2_using_ZZMax_and_SWAP`.

These methods construct circuits via a new method in CircPool.cpp `TK2_Swap_replacement`, which finds cases where wire swaps improve gate count.

To cleanup remaining single qubit gates, I've also updated `standard_rebase_via_tk2` to move the `TK2` gate decomposition into the for loop that replaces all gates with `TK2` operations. This makes it easier apply single qubit cleanup to just the gates as part of that decomposition, shouldn't effect performance of the overall method, and seemed to be necessary to make gates such as `SWAP` return only a wire swap and not any further single qubit gates.

### DIFF
--- a/pytket/binders/circuit/library.cpp
+++ b/pytket/binders/circuit/library.cpp
@@ -44,7 +44,8 @@ void init_library(py::module &m) {
   library_m.def(
       "_TK2_using_CX_and_swap", &CircPool::TK2_using_CX_and_swap,
       "Given expressions α, β and γ, return circuit equivalent to "
-      "TK2(α, β, γ), up to a wire swap, using up to 3 CX and single-qubit "
+      "TK2(α, β, γ), up to a wire swap that is encoded in the implicit "
+      " qubit permutation of the Circuit, using up to 3 CX and single-qubit "
       "gates.\n\n"
       "The decomposition minimizes the number of CX gates.");
   library_m.def(
@@ -74,6 +75,9 @@ void init_library(py::module &m) {
   library_m.def(
       "_CX_using_ZZMax", &CircPool::CX_using_ZZMax,
       "Equivalent to CX, using only ZZMax, Rx and Rz gates");
+  library_m.def(
+      "_CX_using_ZZPhase", &CircPool::CX_using_ZZPhase,
+      "Equivalent to CX, using only ZZPhase, Rx and Rz gates");
   library_m.def(
       "_CX_using_XXPhase_0", &CircPool::CX_using_XXPhase_0,
       "Equivalent to CX, using only XXPhase, Rx, Ry and Rz gates");
@@ -223,6 +227,10 @@ void init_library(py::module &m) {
       "_TK2_using_ZZPhase", &CircPool::TK2_using_ZZPhase,
       "Equivalent to TK2, using 3 ZZPhase gates");
   library_m.def(
+      "_TK2_using_ZZPhase_and_swap", &CircPool::TK2_using_ZZPhase_and_swap,
+      "Equivalent to TK2, up to a wire swap that is encoded in the implicit "
+      " qubit permutation of the Circuit, using up to 3 ZZPhase gates.");
+  library_m.def(
       "_approx_TK2_using_1xZZPhase", &CircPool::approx_TK2_using_1xZZPhase,
       "Approximate equivalent to TK2, using 1 ZZPhase gate and single-qubit "
       "gates. Only requires the first angle of the TK2 gate.");
@@ -235,7 +243,8 @@ void init_library(py::module &m) {
       "Equivalent to TK2, using up to 3 ZZMax gates.");
   library_m.def(
       "_TK2_using_ZZMax_and_swap", &CircPool::TK2_using_ZZMax_and_swap,
-      "Equivalent to TK2, up to a wire swap, using up to 3 ZZMax gates.");
+      "Equivalent to TK2, up to a wire swap that is encoded in the implicit "
+      " qubit permutation of the Circuit, using up to 3 ZZMax gates.");
   library_m.def(
       "_XXPhase3_using_TK2", &CircPool::XXPhase3_using_TK2,
       "Equivalent to XXPhase3, using three TK2 gates");
--- a/pytket/binders/circuit/library.cpp
+++ b/pytket/binders/circuit/library.cpp
@@ -42,6 +42,12 @@ void init_library(py::module &m) {
       "TK2(α, β, γ) using up to 3 CX and single-qubit gates.\n\n"
       "The decomposition minimizes the number of CX gates.");
   library_m.def(
+      "_TK2_using_CX_and_swap", &CircPool::TK2_using_CX_and_swap,
+      "Given expressions α, β and γ, return circuit equivalent to "
+      "TK2(α, β, γ), up to a wire swap, using up to 3 CX and single-qubit "
+      "gates.\n\n"
+      "The decomposition minimizes the number of CX gates.");
+  library_m.def(
       "_approx_TK2_using_1xCX", &CircPool::approx_TK2_using_1xCX,
       "Best approximation of TK2 using 1 CX gate and single-qubit gates, using "
       "squared trace fidelity metric. "
@@ -227,6 +233,9 @@ void init_library(py::module &m) {
   library_m.def(
       "_TK2_using_ZZMax", &CircPool::TK2_using_ZZMax,
       "Equivalent to TK2, using up to 3 ZZMax gates.");
+  library_m.def(
+      "_TK2_using_ZZMax_and_swap", &CircPool::TK2_using_ZZMax_and_swap,
+      "Equivalent to TK2, up to a wire swap, using up to 3 ZZMax gates.");
   library_m.def(
       "_XXPhase3_using_TK2", &CircPool::XXPhase3_using_TK2,
       "Equivalent to XXPhase3, using three TK2 gates");
--- a/pytket/binders/circuit/library.cpp
+++ b/pytket/binders/circuit/library.cpp
@@ -231,6 +231,10 @@ void init_library(py::module &m) {
       "Equivalent to TK2, up to a wire swap that is encoded in the implicit "
       " qubit permutation of the Circuit, using up to 3 ZZPhase gates.");
   library_m.def(
+      "_TK2_using_TK2_or_swap", &CircPool::TK2_using_TK2_or_swap,
+      "Either the exact TK2, or a wire swap encoded in the implicit qubit "
+      "permutation of the Circuit and single qubit gates.");
+  library_m.def(
       "_approx_TK2_using_1xZZPhase", &CircPool::approx_TK2_using_1xZZPhase,
       "Approximate equivalent to TK2, using 1 ZZPhase gate and single-qubit "
       "gates. Only requires the first angle of the TK2 gate.");
--- a/pytket/binders/circuit/library.cpp
+++ b/pytket/binders/circuit/library.cpp
@@ -45,7 +45,7 @@ void init_library(py::module &m) {
       "_TK2_using_CX_and_swap", &CircPool::TK2_using_CX_and_swap,
       "Given expressions α, β and γ, return circuit equivalent to "
       "TK2(α, β, γ), up to a wire swap that is encoded in the implicit "
-      " qubit permutation of the Circuit, using up to 3 CX and single-qubit "
+      "qubit permutation of the Circuit, using up to 3 CX and single-qubit "
       "gates.\n\n"
       "The decomposition minimizes the number of CX gates.");
   library_m.def(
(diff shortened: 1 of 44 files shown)


pytket/binders/circuit/library.cpp
```diff
@@ -41,6 +41,12 @@ void init_library(py::module &m) {
       "Given expressions α, β and γ, return circuit equivalent to "
       "TK2(α, β, γ) using up to 3 CX and single-qubit gates.\n\n"
       "The decomposition minimizes the number of CX gates.");
+  library_m.def(
+      "_TK2_using_CX_and_swap", &CircPool::TK2_using_CX_and_swap,
+      "Given expressions α, β and γ, return circuit equivalent to "
+      "TK2(α, β, γ), up to a wire swap, using up to 3 CX and single-qubit "
```
cqc-alec, line 47: Should probably make it clear here (and similarly below and in the documentation for the C++ functions) that the needed wire swap will be encoded in the implicit qubit permutation of the circuit; otherwise it looks like we may have lost that information.

pytket/pytket/passes/auto_rebase.py
```diff
@@ -44,6 +44,13 @@ def _TK2_using_TK2(a: Param, b: Param, c: Param) -> Circuit:
     OpType.ZZMax: _library._TK2_using_ZZMax,
 }
 
+_TK2_CIRCS_WIRE_SWAP: Dict[OpType, Callable[[Param, Param, Param], "Circuit"]] = {
+    OpType.TK2: _TK2_using_TK2,
+    OpType.ZZPhase: _library._TK2_using_ZZPhase,
```
cqc-alec, line 49: Shouldn't we add a special version of this one too?

pytket/pytket/passes/auto_rebase.py
```diff
@@ -127,12 +141,19 @@ def auto_rebase_pass(gateset: Set[OpType]) -> RebaseCustom:
...
 
+    if allow_swaps:
+        try:
+            return RebaseCustom(
+                gateset, get_tk2_decomposition(gateset, allow_swaps), tk1
+            )
+        except NoAutoRebase:
+            pass
     # if the gateset has CX but not TK2, rebase via CX
     if OpType.CX in gateset and OpType.TK2 not in gateset:
```
cqc-alec, line 152: Possibly this should go above the `if allow_swaps` check?

tket/src/Transformations/Rebase.cpp
```diff
@@ -122,7 +122,7 @@ static bool standard_rebase_via_tk2(
         tk2_replacement) {
   bool success = false;
   VertexSet bin;
-
+  bool TK2_allowed = allowed_gates.contains(OpType::TK2);
```
cqc-alec, line 125:
```suggestion
  const bool TK2_allowed = allowed_gates.contains(OpType::TK2);
```

tket/src/Transformations/Rebase.cpp
```diff
@@ -139,6 +139,27 @@ static bool standard_rebase_via_tk2(
       continue;
     // need to convert
     Circuit replacement = TK2_circ_from_multiq(op);
+    // 2. If TK2 gates are not allowed in the target gateset we find a
```
cqc-alec, line 142:
```suggestion
    // If TK2 gates are not allowed in the target gateset we find a
```

tket/src/Transformations/Rebase.cpp
```diff
@@ -148,32 +169,6 @@ static bool standard_rebase_via_tk2(
...
-        } else {
-          circ.substitute(replacement, v, Circuit::VertexDeletion::No);
-        }
-        bin.insert(v);
-        success = true;
-      }
-    }
-  }
-
   // 3. Replace 0- and 1-qubit gates by converting to TK1 and replacing.
```
cqc-alec, line 172:
```suggestion
  // 2. Replace 0- and 1-qubit gates by converting to TK1 and replacing.
```

tket/test/src/test_CompilerPass.cpp
```diff
@@ -1722,5 +1722,114 @@ SCENARIO("Flatten and relabel registers") {
...
+  PassPtr pp_rebase_zzmax = gen_rebase_pass_via_tk2(
+      allowed_gates_zzmax, CircPool::TK2_using_ZZMax_and_swap,
+      CircPool::tk1_to_PhasedXRz);
+
+  GIVEN("Targeting CX gates, ISWAPMax gate.") {
+    Circuit c(2);
+    c.add_op<unsigned>(OpType::ISWAPMax, {0, 1});
+    CompilationUnit cu(c);
+    CHECK(pp_rebase_cx->apply(cu));
+    REQUIRE(cu.get_circ_ref().count_gates(OpType::CX) == 1);
```
cqc-alec, line 1740: Can we add checks of unitary equivalence to all these tests please?

cqc-alec (changes requested):
Nice, just a few comments.

pytket/binders/circuit/library.cpp
sjdilkes: I've updated the definitions for `_TK2_using_CX_and_swap` and `TK2_using_ZZMax_and_swap`, and the comments describing the `CircPool` methods they use to express this

tket/src/Transformations/Rebase.cpp
sjdilkes: added

tket/src/Transformations/Rebase.cpp
sjdilkes: Done

tket/src/Transformations/Rebase.cpp
sjdilkes: Done

pytket/pytket/passes/auto_rebase.py
sjdilkes: For the TK2 case, it may be best to just check if the angles are equivalent to a SWAP, in which case adding a wire SWAP, else completing like usual. There's no advantage to iSWAPMax or Sycamore gates having a wire swap as they would still have the same TK2 cost.

For ZZPhase, we could do the ZZMax conversion and then convert to ZZPhase. Currently we always convert TK2 to 3 ZZPhase + local Clifford corrections. This logic doesn't fit cleanly with the function added in `CircPool`, but I could refactor it to isolate a check for the the number of two qubit gates predicted. If this is < 3 then we could do the construct as CX and convert to ZZPhase, else we could do the normal version.

cqc-alec: +1 to both suggestions.

tket/test/src/test_CompilerPass.cpp
sjdilkes: Have done -> for some reason Sycamore is not providing equivalent unitaries, I will need to investigate

sjdilkes: I'm not sure why I didn't check this first, but after some bug hunting I have found that actually the issue is not with the new code, but in the top level `standard_rebase_via_tk2`, which even with `CircPool::TK2_using_CX` instead of the swap version, produces incorrect unitary matrices.

My suggestion is to add a new jira task to fix this separately, and push with the Sycamore tests commented out.

pytket/pytket/passes/auto_rebase.py
sjdilkes: This is now done

pytket/pytket/passes/auto_rebase.py
```diff
@@ -37,13 +37,28 @@ def _TK2_using_TK2(a: Param, b: Param, c: Param) -> Circuit:
     return Circuit(2).TK2(a, b, c, 0, 1)
 
 
+def _TK2_using_TK2_or_swap(a: Param, b: Param, c: Param) -> Circuit:
+    if a == b == c == 0.5:
```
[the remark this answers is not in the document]
sjdilkes: I've ended up moving this into the c++ and made it work slightly differently, so no longer an issue.

pytket/binders/circuit/library.cpp
```diff
@@ -227,6 +241,10 @@ void init_library(py::module &m) {
   library_m.def(
       "_TK2_using_ZZMax", &CircPool::TK2_using_ZZMax,
       "Equivalent to TK2, using up to 3 ZZMax gates.");
+  library_m.def(
+      "_TK2_using_ZZMax_and_swap", &CircPool::TK2_using_ZZMax_and_swap,
+      "Equivalent to TK2, up to a wire swap that is encoded in the implicit "
+      " qubit permutation of the Circuit, using up to 3 ZZMax gates.");
```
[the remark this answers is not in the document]
sjdilkes: done

pytket/binders/circuit/library.cpp
```diff
@@ -41,6 +41,13 @@ void init_library(py::module &m) {
       "Given expressions α, β and γ, return circuit equivalent to "
       "TK2(α, β, γ) using up to 3 CX and single-qubit gates.\n\n"
       "The decomposition minimizes the number of CX gates.");
+  library_m.def(
+      "_TK2_using_CX_and_swap", &CircPool::TK2_using_CX_and_swap,
+      "Given expressions α, β and γ, return circuit equivalent to "
+      "TK2(α, β, γ), up to a wire swap that is encoded in the implicit "
+      " qubit permutation of the Circuit, using up to 3 CX and single-qubit "
```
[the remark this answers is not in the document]
sjdilkes: done

tket/src/Circuit/CircPool.cpp
```diff
@@ -935,6 +951,174 @@ Circuit TK2_using_CX(const Expr &alpha, const Expr &beta, const Expr &gamma) {
...
+      angles_eval[last_angle] = *eval;
+    } else {
+      break;
+    }
+  }
+
+  unsigned last_angle_swapped = 0;
+  for (; last_angle_swapped < 3; ++last_angle_swapped) {
+    std::optional<double> eval =
+        eval_expr_mod(angles_swapped[last_angle_swapped]);
```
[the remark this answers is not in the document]
sjdilkes: this code is now removed

tket/src/Circuit/CircPool.cpp
```diff
@@ -935,6 +951,174 @@ Circuit TK2_using_CX(const Expr &alpha, const Expr &beta, const Expr &gamma) {
...
+  std::tie(pre, angles_swapped, post) = normalise_TK2_angles(
+      angles_swapped[0], angles_swapped[1], angles_swapped[2]);
+  pre.add_phase(0.25);
+
+  // Try to evaluate exprs to doubles.
+  std::array<double, 3> angles_eval;
+  std::array<double, 3> angles_eval_swapped;
+  unsigned last_angle = 0;
+  for (; last_angle < 3; ++last_angle) {
+    std::optional<double> eval = eval_expr_mod(angles[last_angle]);
```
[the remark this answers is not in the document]
sjdilkes: this code is now removed

tket/src/Circuit/CircPool.cpp
```diff
@@ -935,6 +951,174 @@ Circuit TK2_using_CX(const Expr &alpha, const Expr &beta, const Expr &gamma) {
...
+    if (eval) {
+      angles_eval[last_angle] = *eval;
+    } else {
+      break;
+    }
+  }
+
+  unsigned last_angle_swapped = 0;
+  for (; last_angle_swapped < 3; ++last_angle_swapped) {
```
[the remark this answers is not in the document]
sjdilkes: We would have needed to assign `last_angle_swapped` from the value i later if we did this. This code is now removed though so no longer an issue.

tket/include/tket/Circuit/CircPool.hpp
```diff
@@ -316,6 +319,18 @@ Circuit normalised_TK2_using_CX(
...
 Circuit TK2_using_CX(const Expr &alpha, const Expr &beta, const Expr &gamma);
 
+/**
+ * @brief Equivalent to TK2(α, β, γ)  up to a wire swap that is encoded in
+ * the implicit qubit permutation of the Circuit with minimal number of
+ * CX gates.
+ *
+ * A TK2-equivalent circuit with as few CX gates as possible (0, 1, 2 or 3 CX).
+ *
+ * @return Circuit Equivalent circuit, up to a wire swap, to TK2(α, β, γ).
```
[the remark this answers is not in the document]
cqc-alec: I think they're all wrong actually: https://www.doxygen.nl/manual/commands.html#cmdreturn ... doxygen can infer the return type from the signature. But let's ignore for this PR.

sjdilkes:
> A couple of pytket tests failing?

I forgot to update the n_gates value for Sycamore for ZZPhase and ZZMax, have fixed now

cqc-alec:
> > A couple of pytket tests failing?
> 
> I forgot to update the n_gates value for Sycamore for ZZPhase and ZZMax, have fixed now

Just the one still failing...

PR #952, opened 2023-07-31
Bugfix/slow get statevector

yao-cqc:
`box_ptr->get_unitary` always returns a value because it will decompose the box then compute the unitary if the box doesn't implement the `get_box_unitary` method. This unnecessarily compute the unitary of the decomposed box during state vector simulation.


cqc-alec (approved):
Thanks!

PR #960, opened 2023-08-07
[infra] Build fixes

doug-q:
My gcc 12 toolchain needs the <algorithm> includes for std::is_sorted.

tkassert must be a PUBLIC dependency of tktokenswap because the public header VectorListHybrid.hpp includes tkassert/Assert.hpp.

tklog must be a PUBLIC dependency of tkassert because Assert.hpp includes tklog/TketLog.hpp.


trvto:
Looks good! I'm not sure but I think this may warrant/require a version bump of the libraries that changed.

trvto (changes requested):
After talking to Alec, we should bump the library versions. Can then update the tket dependencies in a another PR

cqc-alec:
With the version bump we may need two more PRs, one to update the other libraries that depend on tkassert and then one for tket.

PR #972, opened 2023-08-15
Update `TK2_circ_from_multiq` to use predefined decompositions

sjdilkes:
`CircPool` has many predefined definitions for certain `OpType` into `TK2` gates. Currently these are not being used, with the logic called in `with_TK2` rederiving decompositions for gates such as `XXPhase` every time `synthesise_tk` is called.

This PR adds more definitions to the switch statement in `TK2_circ_from_multiq` to avoid this. This can also reduce the number of extraneous single qubit gates: `YYPhase` now decomposes to a single `TK2` with a rotation in the y-basis, rather than an x-basis `TK2` rotation with single qubit wrappers.

### DIFF
--- a/.github/workflows/build-without-conan.yml
+++ b/.github/workflows/build-without-conan.yml
@@ -96,9 +96,9 @@ jobs:
     - name: Install pybind11
       run: |
         cd ${TMP_DIR}
-        wget https://github.com/pybind/pybind11/archive/refs/tags/v2.10.4.tar.gz
-        tar xzvf v2.10.4.tar.gz
-        cd pybind11-2.10.4/
+        wget https://github.com/pybind/pybind11/archive/refs/tags/v2.11.1.tar.gz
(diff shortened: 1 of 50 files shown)


sjdilkes:
> I think this might be caused by a bug in `with_TK2`.
> 
> At the end of `with_TK2` there is a switch statement that has all the `**_using_TK2` functions. However, the `else if (n == 2 && op->free_symbols().empty())` condition above the switch statement seems to catch all 2-qubit ops.
> 
> I might have misread the code.

You're right, that is causing a lot of issues. However, I first moved that switch statement forward and it caused an issue in some tests. So while manually working them out each time is not good, it does seem to be doing something.

I think a full investigation into what's happening should be done as part of the issue you raised and also into any work on merging TK2 and TK1, however for now doing this might be able to help with some other work Vik is doing so I'd be keen to get this small fix for this gate subset merged in sooner.

PR #977, opened 2023-08-18
Feature/add control_state argument to QControlBox

yao-cqc:


cqc-alec (commented):
Looks good, just a few small comments.

pytket/binders/circuit/boxes.cpp
```diff
@@ -293,7 +321,13 @@ void init_boxes(py::module &m) {
...
       .def(
           "get_n_controls", &QControlBox::get_n_controls,
-          ":return: the number of control qubits");
+          ":return: the number of control qubits")
+      .def(
+          "get_control_state",
+          [](QControlBox &qcbox) {
+            return bin_to_dec(qcbox.get_control_state());
+          },
+          ":return: the control state as an integer");
```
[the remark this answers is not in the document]
yao-cqc: I couldn't decide on what to return, but you are right, we can have both

tket/include/tket/Circuit/Boxes.hpp
```diff
@@ -498,7 +498,9 @@ class QControlBox : public Box {
    *
    * @param op op to control
    * @param n_controls number of qubit controls to add
-   * @param control_state control state expressed as a bit vector
+   * @param control_state control state expressed as a bit vector.
+   * The size of control_state should match n_controls. When n_controls > 0,
+   * empty vector will be converted to a vector with all 1s.
```
cqc-alec, line 503:
```suggestion
   * If control_state is non-empty, its size should match n_controls.
   * An empty vector is converted to an all-1s vector of length n_controls.
```

tket/test/src/Circuit/test_Boxes.cpp
```diff
@@ -811,6 +811,24 @@ SCENARIO("QControlBox", "[boxes]") {
...
+    Eigen::MatrixXcd u = random_unitary(8, 1);
+    Unitary3qBox ubox(u);
+    Op_ptr op = std::make_shared<Unitary3qBox>(ubox);
+    QControlBox qcbox(op, 3, {1, 0, 1});
+    std::shared_ptr<Circuit> c = qcbox.to_circuit();
+    const Eigen::MatrixXcd circ_u = tket_sim::get_unitary(*c);
+    Eigen::MatrixXcd V = Eigen::MatrixXcd::Identity(64, 64);
+    for (unsigned i = 0; i < 8; i++) {
+      for (unsigned j = 0; j < 8; j++) {
+        V(40 + i, 40 + j) = u(i, j);
```
cqc-alec, line 824:
```suggestion
    QControlBox qcbox(op, 3, {1, 0, 0});
    std::shared_ptr<Circuit> c = qcbox.to_circuit();
    const Eigen::MatrixXcd circ_u = tket_sim::get_unitary(*c);
    Eigen::MatrixXcd V = Eigen::MatrixXcd::Identity(64, 64);
    for (unsigned i = 0; i < 8; i++) {
      for (unsigned j = 0; j < 8; j++) {
        V(32 + i, 32 + j) = u(i, j);
```
I prefer this test because it catches endianness mistakes.

PR #984, opened 2023-08-24
Feature/conjugation box

yao-cqc:
Adds ConjugationBox to capture any compute-action-uncompute pattern.
By default, uncompute will be derived from compute, but the user has the ability to provide a custom operation.
The current pytket interface has some limitation therefore the three operations need to have the same number of qubits.

### DIFF
--- a/pytket/binders/circuit/Circuit/add_op.cpp
+++ b/pytket/binders/circuit/Circuit/add_op.cpp
@@ -26,6 +26,7 @@
 #include "tket/Circuit/Boxes.hpp"
 #include "tket/Circuit/Circuit.hpp"
 #include "tket/Circuit/ClassicalExpBox.hpp"
+#include "tket/Circuit/ConjugationBox.hpp"
 #include "tket/Circuit/DiagonalBox.hpp"
 #include "tket/Circuit/Multiplexor.hpp"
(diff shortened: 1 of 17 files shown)


sjdilkes (changes requested):
The code all looks great. Do you think it's worth adding case handling for `QControlBox` with this new conjugation box to only control on the action in this PR? It's the obvious functionality that this gives benefit to.

Similarly, `PauliExpBox` could be considered a child of this new box type, do you think its easy to update `PauliExpBox` to have a similar structure and benefit from quantum control off only the action (i.e. rotation)?

yao-cqc:
> The code all looks great. Do you think it's worth adding case handling for `QControlBox` with this new conjugation box to only control on the action in this PR? It's the obvious functionality that this gives benefit to.
> 
> Similarly, `PauliExpBox` could be considered a child of this new box type, do you think its easy to update `PauliExpBox` to have a similar structure and benefit from quantum control off only the action (i.e. rotation)?

Yeah, I plan to do these things in separate PR(s). The `QControlBox` still need some changes for this to work. One possible solution is to overload the `decompose_boxes` method so it can stop the recursion at certain boxes (`ConjugationBox` in this case) so the `QControlBox` can spot the `ConjugationBox`es.

sjdilkes (approved):

PR #994, opened 2023-08-31
make tests not a package again

trvto:
I made `pytket.tests` a package to more easily satisfy `mypy`, but that broke the build wheel and test CI for some reason.
This makes `pytket.tests` not a package again. After reading up [here](https://docs.pytest.org/en/6.2.x/goodpractices.html),
having tests as a package is not good practice anyway (it would be okay if we had the pytket package under `src/pytket` instead of just `pytket`.

### DIFF
--- a/pytket/tests/assertion_test.py
+++ b/pytket/tests/assertion_test.py
@@ -27,7 +27,7 @@ from pytket.passes import (
 )
 
 from pytket.pauli import PauliStabiliser, Pauli
-from .simulator import TketSimShotBackend
+from simulator import TketSimShotBackend  # type: ignore
 
 
 def test_assertion_init() -> None:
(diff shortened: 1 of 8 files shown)


trvto:
The check that this fixes the wheel build is running here: https://github.com/CQCL/tket/actions/runs/6039343076

cqc-alec (approved):
Thanks! LGTM as long as the wheel build passes.

cqc-alec (commented):
Seems some tests are failing, maybe missed some?

cqc-alec (changes requested):
Need to fix some other imports?

cqc-alec (approved):

cqc-alec (approved):

PR #1004, opened 2023-09-05
Fix implementations of symbol_substitution() for boxes that cannot contain symbols.

### LINKED ISSUE #999: Segfault substituting symbols in circuit containing boxes
Running

```python
import pickle

with open("my_circ.pickle", "rb") as f:
    cdict = pickle.load(f)
c = Circuit.from_dict(cdict)
sym_map = {s: 0.2 for s in c.free_symbols()}
c.symbol_substitution(sym_map)
```

with the attached file (unzipped) gives a segmentation fault.

[my_circ.zip](https://github.com/CQCL/tket/files/12512442/my_circ.zip)

cqc-alec:
These were just returning a null `Op_ptr`.

Fixes #999 .

### DIFF
--- a/tket/include/tket/Circuit/ConjugationBox.hpp
+++ b/tket/include/tket/Circuit/ConjugationBox.hpp
@@ -47,7 +47,7 @@ class ConjugationBox : public Box {
 
   Op_ptr symbol_substitution(
       const SymEngine::map_basic_basic &) const override {
-    return std::make_shared<ConjugationBox>(compute_, action_, uncompute_);
+    return std::make_shared<ConjugationBox>(*this);
   }
 
   SymSet free_symbols() const override { return {}; }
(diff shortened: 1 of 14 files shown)


tket/include/tket/Circuit/ConjugationBox.hpp
```diff
@@ -47,7 +47,7 @@ class ConjugationBox : public Box {
 
   Op_ptr symbol_substitution(
       const SymEngine::map_basic_basic &) const override {
-    return Op_ptr();
+    return std::make_shared<ConjugationBox>(compute_, action_, uncompute_);
   }
```
yao-cqc, line 51: `ConjugationBox` should support symbols, but we can fix it in a separate PR

yao-cqc (commented):
Is there a reason not to call the copy constructors?
e.g. `return std::make_shared<Unitary1qBox>(*this);`

tket/include/tket/Circuit/ConjugationBox.hpp
cqc-alec: Made an issue and added a comment.

PR #1017, opened 2023-09-12
update changelog

cqc-melf:


pytket/docs/changelog.rst
```diff
@@ -1,14 +1,21 @@
 Changelog
 =========
 
-1.19.1 (September 2023)
------------------------
+1.20.0 (unreleased)
```
cqc-alec, line 4: I don't think we should bump the version number; this applies to 1.19.1.

cqc-alec: Sorry I see this is not in 1.19.1. But then the line below (first "minor new feature") should also be moved up here.

cqc-alec: Also normally just put "Unreleased" as the heading, without a version number.

cqc-melf: I wanted to point out that this is in the 1.20.0rc1, maybe this is now the way to go?

cqc-melf: The minor new feature should be up already, or is misunderstood what you mean?

cqc-alec: Sorry yes you are right.

Yes we can put 1.20.0rc0 as the release, and change it to 1.20.0 when we do the release.

cqc-melf: see [002cac3](https://github.com/CQCL/tket/pull/1017/commits/002cac30629a5aca22907ceb99fe37b1b13621d1)

cqc-alec (approved):

PR #1021, opened 2023-09-13
fix wasm, update changelog

cqc-melf:
fixes the issues of the wasm parsing on the pytket side.

I would like to put out a release candidate after this is merged


pytket/pytket/wasm/wasm.py
```diff
@@ -68,6 +68,7 @@ def __init__(self, filepath: str, check_file: bool = True, int_size: int = 32):
 
         function_signatures: list = []
         function_names: list = []
+        _func_lookup = {}
```
cqc-alec, line 71: Can we please add a test that shows the issue this is fixing (i.e. fails before the fix)?

cqc-melf: I have added one additional file that is now read in

pytket/tests/classical_test.py
```diff
@@ -458,6 +458,52 @@ def test_wasmfilehandler_repr_64() -> None:
     )
 
 
+def test_wasmfilehandler_repr_2() -> None:
```
cqc-alec, line 461: Looks like this test (and the following one) pass with the existing version, or am I missing something? Do they exhibit the bug?

pytket/tests/wasm-generation/src/lib-2.rs
```diff
@@ -0,0 +1,107 @@
+// src/lib.rs
+
+#[no_mangle]
+fn init() {
```
cqc-melf, line 4: for example here

pytket/tests/wasm-generation/src/lib-2.rs
```diff
@@ -0,0 +1,107 @@
...
+    let mut i = 0;
+
+    while i < limit {
+        i = i * 2;
+    }
+return i
+}
+
+#[no_mangle]
+pub fn mixed_up_2(limit: i32, limit2: i32) -> i32 {
```
cqc-melf, line 65: and here

pytket/tests/wasm-generation/src/lib-2.rs
```diff
@@ -0,0 +1,107 @@
...
+
+    while i < limit2 {
+        i = i * 3;
+    }
+return i
+}
+
+
+#[no_mangle]
+fn mixed_up_3(limit: i32, limit2: i32, limit3: i32) -> i32 {
```
cqc-melf, line 80: and here

pytket/tests/classical_test.py
cqc-alec: Ah OK, thanks!

cqc-alec (approved):

PR #1026, opened 2023-09-15
Bugfix/slow cliffordsimp

### LINKED ISSUE #1018: `FullPeepholeOptimise` hangs
Running `FullPeepholeOptimise` on https://github.com/CQCL/tket_benchmarking/blob/master/active/chem_qasm/LiH_cmplt_JW_sto3g.qasm
never terminates.

Initial investigation suggests that `clifford_reduction` is where it got stuck.

yao-cqc:
Fixes #1018 

I initially thought that the `CliffordSimp` could enter an infinite loop. However, it turned out that a particular gate in the test circuit triggered a series of merging actions of 2-qubit Clifford gates. As the merging points moved closer to the front of the circuit, the time taken for each merge increased and eventually became extremely slow.

After each merging, the algorithm would traverse the causal future of that merging point and remove any potential interactions in this causal cone. The issue was that the code had a bug in it, which caused the traversal to blow up when the cone got larger.

### DIFF
--- a/tket/src/Transformations/CliffordReductionPass.cpp
+++ b/tket/src/Transformations/CliffordReductionPass.cpp
@@ -410,8 +410,12 @@ Subcircuit CliffordReductionPass::substitute(
   // edges.
   EdgeList future_edges;
   VertexSet v_frontier;
+  // keep track of visited vertices
+  VertexSet visited;
   for (unsigned qi = 0; qi < q_width; ++qi) {
-    v_frontier.insert(circ.target(out_edges[qi]));
+    Vertex target = circ.target(out_edges[qi]);
+    visited.insert(target);
+    v_frontier.insert(target);
   }
   while (!v_frontier.empty()) {
     EdgeSet out_edges;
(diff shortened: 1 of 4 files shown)


tket/src/Transformations/CliffordReductionPass.cpp
```diff
@@ -410,8 +410,12 @@ Subcircuit CliffordReductionPass::substitute(
...
   EdgeList future_edges;
   VertexSet v_frontier;
+  // keep track of visited vertices
+  VertexSet visited;
   for (unsigned qi = 0; qi < q_width; ++qi) {
-    v_frontier.insert(circ.target(out_edges[qi]));
+    Vertex target = circ.target(out_edges[qi]);
+    visited.insert(target);
+    v_frontier.insert(target);
   }
```
cqc-alec, line 419: Maybe this would be easier to read if we just assigned
```c++
VertextSet visited = v_frontier;
```
after the original loop? Up to you.

cqc-alec (approved):
Good detective work! Thanks.

tket/src/Transformations/CliffordReductionPass.cpp
yao-cqc: ahh, sorry I didn't see this...

PR #1033, opened 2023-09-18
Set cppstd to gnu17 in default profile on MacOS.

cqc-alec:


trvto:
I guess this is a bug in the newest conan version?

trvto (approved):

cqc-alec:
> I guess this is a bug in the newest conan version?

Given the warnings that the command prints, I guess it doesn't count as a bug:

```
WARN: This profile is a guess of your environment, please check it.
WARN: Defaulted to cppstd='gnu17' for apple-clang.
WARN: The output of this command is not guaranteed to be stable and can change in future Conan versions.
WARN: Use your own profile files for stability.
```

PR #1040, opened 2023-09-19
Feature/selectively decompose boxes

### LINKED ISSUE #1011: Allow `DecomposeBoxes()` to act only on boxes in a specified op group

yao-cqc:
Allow `DecomposeBoxes` to exclude boxes from being decomposed.
Resolves #1011


tket/src/Circuit/macro_manipulation.cpp
```diff
@@ -708,10 +708,19 @@ bool Circuit::substitute_box_vertex(
   return true;
 }
 
-bool Circuit::decompose_boxes() {
+bool Circuit::decompose_boxes_recursively(
```
yao-cqc, line 711: We're using `std::list` to store the graph vertices. Also, according to the Boost documentation, vertices are added to the graph using `std::list::push_back`. So, any vertex added inside the loop will always get looped through in the future. `BGL_FORALL_VERTICES` essentially just goes through the `std::list`, so the code here shouldn't cause any undefined behaviour.

cqc-alec (approved):
LGTM!

PR #1041, opened 2023-09-20
Use clang-format v17

yao-cqc:


tket/src/Characterisation/Cycles.cpp
```diff
@@ -148,7 +148,7 @@ void CycleFinder::order_keys(
             goto new_loop;
           }
 
-  new_loop : {}
+  new_loop: {}
```
yao-cqc, line 151: needed to bump the tket version just for this...

PR #1044, opened 2023-09-20
Avoid setuptools_scm 8

cqc-alec:
Short-term fix until we find a better one.


trvto:
I think I might have a fix

PR #1048, opened 2023-09-21
Feature/pauli exp box decomposes into conjugation box

### LINKED ISSUE #911: More efficient decompositions of (multi-)controlled `PauliExpBox`es
Instead off adding controls to every CX, we should only add a control to the central Rz.

(See https://quantumcomputing.stackexchange.com/questions/24122/how-to-perform-a-controlled-pauli-string-rotation-gate .)

yao-cqc:
resolves #911

### DIFF
--- a/pytket/conanfile.py
+++ b/pytket/conanfile.py
@@ -32,7 +32,7 @@ class pytketRecipe(ConanFile):
         cmake.install()
 
     def requirements(self):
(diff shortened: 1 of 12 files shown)


yao-cqc:
It seems that gnu.org is currently down.

cqc-alec (approved):
LGTM!

PR #1055, opened 2023-09-26
CircBox symbol substitution

### LINKED ISSUE #1014: Make python bindings for op-level symbol substitution
At least for `CircBox`; if easy we should make it work for all op types.

trvto:
- Creates a method for in-place symbol substitution for the circuit within a `CircBox`
- Exposes method in pytket
Fixes #1014


pytket/binders/circuit/boxes.cpp
```diff
@@ -76,7 +76,20 @@ void init_boxes(py::module &m) {
           py::arg("circ"))
       .def(
           "get_circuit", [](CircBox &cbox) { return *cbox.to_circuit(); },
-          ":return: the :py:class:`Circuit` described by the box");
+          ":return: the :py:class:`Circuit` described by the box")
+      .def(
+          "symbol_substitution",
```
cqc-alec, line 81: It may be worth noting here that if a `CircBox` has been added to a circuit before this method is called, calling this method affects the circuit it has been added to. I can see this being a trip hazard.

trvto: I agree it seems like a hazard. Should we try to avoid this behaviour? (Should `add_circbox` add a copy of the circbox to the circuit?)

cqc-alec:
> I added a warning in the docstring of `CircBox.symbol_substitution`. I also removed the "Will not affect boxes..." part in the `Circuit.symbol_substitution` docstrings. I think that is not correct anymore? Please confirm

Good spot, yes it isn't correct.

cqc-alec (approved):

PR #1058, opened 2023-09-27
add aditional check for index out of array functions

cqc-melf:
Add an additional check if the function is within the file


cqc-melf:
I would like to include this in the release candidate, I will add tests later next week?

cqc-alec:
Tests will need to be rerun after the latest develop merge has finished so that the new tket version is uploaded...

cqc-alec (approved):

PR #1061, opened 2023-09-29
Fix/tket version

cqc-melf:
currently broken on develop, try to fix this


cqc-melf:
Not sure if this is the right way to solve this

cqc-melf:
For the problem see https://github.com/CQCL/tket/actions/runs/6349246471

cqc-melf:
I think this needs two PRs to be solved

cqc-alec:
I think this is because of a mismatch of boost revisions (there was a new revision on the conan-center uploaded in the last day or two). For now I will just rebuild the tkwsm and tktokenswap libraries and upload them and then this should work again.

cqc-melf:
Great, thank you. I will close this now?

cqc-alec:
> Great, thank you. I will close this now?

I think so yes.

PR #1072, opened 2023-10-11
Improve typing

### LINKED ISSUE #1067: Stubs contains invalid type `Dict[List[bool], ...]`
Lists are not hashable hence cannot be used as dictionary keys.

There are several cases in `circuit.pyi`, e.g.

```python
class MultiplexedU2Box(Op):
    @overload
    def __init__(self, op_map: Dict[List[bool],Op], impl_diag: bool = ...) -> None: ...
```

It should be `Dict[Tuple[bool],Op]`

trvto:
Sorry in advance! Fixes #1067

Improvements:
  - Use `pybind11-stubgen` instead of `stubgen` for stub generation
     - Since typing was introduced in the compiled modules, `pybind11-stubgen` has been completely rewritten and is no longer in a "pending deprecation" state.
     - New `pybind11-stubgen` does a much better job
        - Puts docstrings into the stubs (so they will be visible when hovering over a method in many text editors)
        - Improved handling of some types (`stubgen` doens't recognize `tuple[T, ...]` for some reason
        - Better typing of method extensions in python (e.g. `__and__` etc. for `Bit`/`BitRegister`). This exposed some typing problems in other places (mostly `pytket.circuit.logic_exp`)
  - Cast `std::vectors<T>` and `std::list<T>` to `typing.Sequence[T]` and `tuple[T, ...]` where it makes sense
     - It took some hacking of pybind11 but it helps in many situations that people have been complaining about, and makes some constructers viable again (ones that wanted a `dict[list[T], V]` (an unviable type) are now typed as `dict[tuple[T, ...], V]`, which was always acceptable at runtime anyway.
  - Big changes were needed in the `logic_exp.py` code in order to get it typed right.
  - Removed a lot of `cast` calls. Cleaned up the typing in some places to make that possible.
  - Added `pytket.circuit.named_types` module with some useful type definitions 

Note: Could possibly cause some `mypy` problems again in the extensions. May need to think about the best way to check/update them

Drive-By: In the build_and_test workflow, distinguish between things that should be done when tket changes only or if tket *or* the workflow changes


pytket/binders/circuit_library.cpp
```diff
@@ -22,6 +22,9 @@ namespace py = pybind11;
 namespace tket {
 
 PYBIND11_MODULE(circuit_library, library_m) {
+  library_m.def(
+      "BRIDGE_using_CX_0", &CircPool::BRIDGE_using_CX_0,
+      "Equivalent to BRIDGE, using four CX, first CX has control on qubit 0");
   library_m.def(
       "_BRIDGE_using_CX_0", &CircPool::BRIDGE_using_CX_0,
```
cqc-alec, line 29: I checked that this isn't used in any of the extensions. So I think we can remove the underscored version.

pytket/binders/include/typecast.hpp
```diff
@@ -25,14 +26,125 @@
...
+// to tuple[T, ...] on python side Instead of list[T]. Can be used as a
+// parameter type or return type.
+template <typename T>
+class TupleVec : public std::vector<T> {
+  using std::vector<T>::vector;
+};
+PYBIND11_NAMESPACE_END(tket_custom)
 PYBIND11_NAMESPACE_BEGIN(detail)
+// This struct is copied from the struct "list_caster" in pybind11/stl.h with
+// some minor customization It adds the ability to customize the type name
```
cqc-alec, line 56:
```suggestion
// some minor customization. It adds the ability to customize the type name
```

pytket/binders/include/typecast.hpp
```diff
@@ -25,14 +26,125 @@
...
+// (using a handle_type_name<T> struct) and specify the python type that the
+// object is cast too Changes to the pybind11 code may warrant/require changes
```
cqc-alec, line 58:
```suggestion
// object is cast to. Changes to the pybind11 code may warrant/require changes
```

pytket/binders/include/typecast.hpp
```diff
@@ -25,14 +26,125 @@
...
+// here The struct is used to define custom type casters for the "tket_custom"
```
cqc-alec, line 59:
```suggestion
// here. The struct is used to define custom type casters for the "tket_custom"
```

pytket/binders/include/typecast.hpp
```diff
@@ -25,14 +26,125 @@
...
+// types
```
cqc-alec, line 60:
```suggestion
// types.
```

pytket/binders/passes.cpp
```diff
@@ -850,7 +864,11 @@ PYBIND11_MODULE(passes, m) {
   m.def(
       "SimplifyInitial",
       [](bool allow_classical, bool create_all_qubits, bool remove_redundancies,
-         std::shared_ptr<const Circuit> xcirc) -> PassPtr {
+         std::optional<std::shared_ptr<const Circuit>> xcirc_opt) -> PassPtr {
+        std::shared_ptr<const Circuit> xcirc = nullptr;
+        if (xcirc_opt.has_value()) {
+          xcirc = xcirc_opt.value();
+        }
```
cqc-alec, line 871: Not sure why we need to make a new `xcirc` rather than just use the parameter directly?

cqc-alec (commented):
Heroic.

I'm just building a wheel from this branch in order to test some of the extensions with it.

pytket/binders/passes.cpp
trvto: well `xcirc_opt` is a `std::optional<>`  so it can't be passed directly, but you are right there is a better way, namely: `xcirc_opt.value_or(nullptr)`. Using that now.

Btw. the reason this is needed is because mypy (now...) complains about setting the default of a non `Optional` value to `None`.

cqc-alec: Ah OK I misread it, thought it was a shared_ptr. Thanks.

cqc-alec:
Tested pytket-qiskit and pytket-quantinuum with the wheel from this branch. Tests passed; mypy checks only failed with

```
error: Unused "type: ignore" comment  [unused-ignore]
```

errors, which are easy to fix!

pytket/binders/architecture.cpp
```diff
@@ -94,12 +101,6 @@ PYBIND11_MODULE(architecture, m) {
       .def(
           "__deepcopy__", [](const Architecture &arc,
                              const py::dict & = py::dict()) { return arc; })
-      .def(
```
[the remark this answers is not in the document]
trvto: Not sure anymore why I deleted it. Looking into it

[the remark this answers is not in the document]
trvto: Oh now I know, it was a duplicate :)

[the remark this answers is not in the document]
trvto: see line 53

trvto:
BTW, I added the `BRIDGE_using_CX_0` without underscore only because `pybind11-stubgen` includes all non underscored methods in the `__all__` attribute the `.pyi` file. When there are no methods without underscore it causes an error. It is a bit weird though maybe that now there is one circuit without underscore and all others have it.

Should I maybe add non-underscored circuits for all of them, so the underscored ones can be phases out? Or leave it like it is?

cqc-alec:
> Should I maybe add non-underscored circuits for all of them, so the underscored ones can be phased out? Or leave it like it is?

I think that's a good idea, yes.

cqc-alec (approved):
Thanks!

PR #1077, opened 2023-10-13
Ccache windows dynamic build

trvto:
`ccache 4.8.3` includes a bugfix that makes this possible 🥳

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -476,11 +476,17 @@ jobs:
         key: tket-dynamic-visual-studio-windows-2022-${{ steps.current_time.outputs.formattedTime }}
         restore-keys: |
           tket-dynamic-visual-studio-windows-2022
+    - name: get ccache version
+      id: ccache-ver
+      shell: bash
+      run: |
+        ccache_ver=$(choco list -e ccache | grep "ccache" | grep -ioE '[0-9]+\.[0-9]+\.[0-9]+')
+        echo "Found ccache version ${ccache_ver}"
+        echo "ccache_ver=${ccache_ver}" >> $GITHUB_OUTPUT
     - name: Build tket
       if: needs.check_changes.outputs.tket_changed == 'true'
       run: |
-        ccache --shimgen-help
-        $env:TKET_VSGEN_CCACHE_EXE = 'C:\\ProgramData\\chocolatey\\lib\\ccache\\tools\\ccache-4.8.3-windows-x86_64\\ccache.exe'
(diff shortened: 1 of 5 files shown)


trvto:
I couldn't figure out how to get the ccache version set dynamically. See the commented out step before the tket build step. If you have any ideas why its not working, please share 😄

cqc-alec:
> I couldn't figure out how to get the ccache version set dynamically. See the commented out step before the tket build step. If you have any ideas why its not working, please share 😄

Is it because the command is running in Powershell and the syntax is different? In which case setting `shell: bash` for this step of the job may work? (I don't know Powershell well.)

cqc-alec:
> Is it because the command is running in Powershell and the syntax is different? In which case setting `shell: bash` for this step of the job may work? (I don't know Powershell well.)

Sorry, I see you did that, ignore me.

trvto:
> > I couldn't figure out how to get the ccache version set dynamically. See the commented out step before the tket build step. If you have any ideas why its not working, please share 😄
> 
> Is it because the command is running in Powershell and the syntax is different? In which case setting `shell: bash` for this step of the job may work? (I don't know Powershell well.)

I am pretty sure I was using `shell: bash`

trvto:
Here you can see what happens: https://github.com/CQCL/tket/actions/runs/6502081399/job/17660502339

cqc-alec:
> Here you can see what happens: https://github.com/CQCL/tket/actions/runs/6502081399/job/17660502339

I wonder if it's something to do with the escaped characters, and whether it would work with the backtick form, i.e.

```shell
ccache_bin="`ccache --shimgen-help | grep -ioE 'C:\\ProgramData\\Chocolatey\\lib\\ccache\\tools\\.+\\ccache.exe'`"
```

cqc-alec:
> > Here you can see what happens: https://github.com/CQCL/tket/actions/runs/6502081399/job/17660502339
> 
> I wonder if it's something to do with the backslashes, and whether it would work with the backtick form, i.e.
> 
> ```shell
> ccache_bin="`ccache --shimgen-help | grep -ioE 'C:\\ProgramData\\Chocolatey\\lib\\ccache\\tools\\.+\\ccache.exe'`"
> ```

OK that didn't work. Sigh, this would be a lot easier with access to a Windows machine...

trvto:
> 
> OK that didn't work. Sigh, this would be a lot easier with access to a Windows machine...

yep

trvto:
Ok I found another route! Working now (at least a version change won't break anything).

cqc-alec (approved):
Thanks!

PR #1078, opened 2023-10-15
fix bug and improve typing of logic expressions

trvto:
- Only allow creation of logic_exps through the various class constructors for each expression (RegOr, RegMul, etc.).
- Only create new LogicExps using create_reg/bit_logic_exp, which use the class constructors.
- Also type the c++ functions better that take and return LogicExps


pytket/tests/classical_test.py
```diff
@@ -1418,12 +1418,9 @@ def test_flatten_registers_with_classical_exps() -> None:
 def test_box_equality_check() -> None:
     exp1 = Bit(2) & Bit(3)
     exp2 = Bit(1) & Bit(3)
-    exp3 = Bit(1)
```
trvto, line 1421: Removed this test because it doesn't fit with the typing of ClassicalExpBox anymore (was `Any` before). As far as I can tell, the only way to add a classical exp box to a circuit is through `circ.add_classicalexpbox_bit/register` and these functions cannot support `Bit` or `BitRegister` as arguments. So I think it doesn't make sense to use a `Bit` or `Bitregister` in the constructor. 

But maybe I'm missing something?

PR #1083, opened 2023-10-17
Make maximum classical register width a parameter of QASM converters

### LINKED ISSUE #1019: Make maximum register width in QASM conversions configurable
Currently it is hard-coded to 32 in a couple of places. We should make this a parameter to the converters.

cqc-alec:
Also improve the docstrings a bit.

Closes #1019 .


yao-cqc (approved):
LGTM!

PR #1087, opened 2023-10-20
Bumped install-nix-action to v23

jake-arkinstall:
This is because the original version (v20) was not compatible with System Identity Protection on Macos

### DIFF
--- a/.github/workflows/build-with-nix.yml
+++ b/.github/workflows/build-with-nix.yml
@@ -15,6 +15,6 @@ jobs:
     runs-on: ${{matrix.os}}
     steps:
     - uses: actions/checkout@v3
-    - uses: cachix/install-nix-action@v20
+    - uses: cachix/install-nix-action@v23
     - name: Build and test tket
       run: nix flake check
--- a/.github/workflows/build-with-nix.yml
+++ b/.github/workflows/build-with-nix.yml
@@ -1,5 +1,12 @@
 name: build with nix
 on:
+  pull_request:
+    branches:
+      - main
+      - develop
+  push:
+    branches:
+      - develop
   schedule:
     # 01:00 every Sunday morning
     - cron: '0 1 * * 0'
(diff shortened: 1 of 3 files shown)


cqc-alec:
```
Segmentation fault: 11  python -m pytest -s .
```
:-(

cqc-alec (commented):
Seems to have solved the first issue but hit a segfault in pytest.

jake-arkinstall:
I can confirm that `nix flake check github:CQCL/tket` succeeds on MacOS 13.6. I don't have access to a mac with macos 12 

I've now added the `-L` flag to `nix flake check` to capture some location information for the segfault.

The issue is happening in the same place as it was on macos13 before the dylib flat namespace patch. I cannot determine whether the issue is the macos version or the CPU architecture.

So this leaves three main options.

One is to resolve the underlying issue, which is that typecast.hpp contains implementation, and that is included into two separate compilation units and is causing the symbol collision. I'm not sure how much pybind11 constrains this, as I believe that implementation does need to be provided in a header.

One is to identify why the vanilla cmake build is working - what's different? _Should_ it be working?

Another is to identify the platforms this currently does work on, and then restrict the support matrix accordingly. This doesn't sound ideal, but might serve as a point to improve on over time.

cqc-alec:
> So this leaves three main options.
> 
> One is to resolve the underlying issue, which is that typecast.hpp contains implementation, and that is included into two separate compilation units and is causing the symbol collision. I'm not sure how much pybind11 constrains this, as I believe that implementation does need to be provided in a header.
> 
> One is to identify why the vanilla cmake build is working - what's different? _Should_ it be working?
> 
> Another is to identify the platforms this currently does work on, and then restrict the support matrix accordingly. This doesn't sound ideal, but might serve as a point to improve on over time.

All those sound like good options, but for now, if it works on macos-13, I'm happy to just change the platform used in the workflow to `macos-13`.

jake-arkinstall:
This is now ready for final review

cqc-alec (approved):

PR #1092, opened 2023-10-24
Bugfix/arg ordering in classical exp box

### LINKED ISSUE #1086: Arguments of commands for register expressions converted from QASM are in random order
```python

qasm = """
OPENQASM 2.0;
include "hqslib1.inc";

qreg q[1];

creg a[4];
creg b[4];
creg c[4];

c = a ^ b;
"""

circ = circuit_from_qasm_str(qasm)
cmd = circ.get_commands()[0]
print(cmd.args)
```

Output:

```
[b[2], a[1], b[1], a[0], b[0], a[3], b[3], a[2], c[0], c[1], c[2], c[3]]
```

As far as I can tell this doesn't affect the semantics: at least, conversion back to QASM with `circuit_to_qasm_str()` recovers the original QASM.

trvto:
- Added `LogicalExp.all_inputs_ordered` and use it where `all_inputs` was used in the qasm code
- Would have just changed the `all_inputs` method but it would require a breaking change (returning `list` instead of `set`)

Fixes #1086


cqc-melf (commented):
It would be good to add something to the changelog, besides that it is looking good

trvto:
> It would be good to add something to the changelog, besides that it is looking good

done

cqc-melf (approved):

Issue #836, opened 2023-04-24
Upgrade to sphinx 6.2.0


yao-cqc:
Related to #834

CalMacCQ:
Updated to sphinx <7 in https://github.com/CQCL/tket/pull/1093

Our sphinx book theme currerty requires sphinx<7

PR #1093, opened 2023-10-25
Remove docs build warning from `pydata-sphinx-theme` 0.14.2 and upgrade sphinx version

CalMacCQ:
This fixes a API docs build issue which is caused by the new `pydata-sphinx-theme` [0.14.2 release](https://github.com/pydata/pydata-sphinx-theme/releases/tag/v0.14.2). The pydata-sphinx-theme is the parent theme of the sphinx book theme which we use for the docs.

This release introduced a new warning which caused the docs build to fail. The warning is the following...

________________________________________
#69 39.94 WARNING: The default value for `navigation_with_keys` will change to `False` in the next release. If you wish to preserve the old behavior for your site, set `navigation_with_keys=True` in the `html_theme_options` dict in your `conf.py` file.Be aware that `navigation_with_keys = True` has negative accessibility implications:https://github.com/pydata/pydata-sphinx-theme/issues/1492
________________________________________

I tested this by running the CI with the first commit. The docs build failed
https://github.com/CQCL/tket/actions/runs/6642169421/job/18046321766?pr=1093

The second commit sets the `navigation_with_keys` key to `true` which fixes the docs build
https://github.com/CQCL/tket/actions/runs/6642374353/job/18047022783?pr=1093.

Other changes...

I've removed the restriction on the sphinx version https://github.com/CQCL/tket/issues/836 . This restriction was due to deprecated intersphinx syntax. We are not using intersphinx at the moment in this repository.

I've removed some unused pygments settings as well

### DIFF
--- a/pytket/docs/conf.py
+++ b/pytket/docs/conf.py
@@ -109,8 +109,6 @@ html_theme_options = {
     },
 }
 
-html_show_sphinx = True
-
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
--- a/pytket/docs/conf.py
+++ b/pytket/docs/conf.py
@@ -55,7 +55,6 @@ release = "1.21.0"
 extensions = [
     "sphinx.ext.autodoc",
     "sphinx.ext.autosummary",
-    "sphinx.ext.intersphinx",
     "sphinx.ext.mathjax",
     "sphinx.ext.viewcode",
     "sphinx_copybutton",
--- a/pytket/docs/conf.py
+++ b/pytket/docs/conf.py
@@ -102,6 +102,7 @@ html_theme_options = {
     "repository_url": "https://github.com/CQCL/tket",
     "use_repository_button": True,
     "use_issues_button": True,
+    "navigation_with_keys": True,
     "logo": {
         "image_light": "_static/Quantinuum_logo_black.png",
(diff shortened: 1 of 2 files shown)


pytket/docs/conf.py
```diff
@@ -102,16 +97,20 @@
...
     "use_repository_button": True,
     "use_issues_button": True,
+    "navigation_with_keys": True,
     "logo": {
         "image_light": "_static/Quantinuum_logo_black.png",
         "image_dark": "_static/Quantinuum_logo_white.png",
     },
 }
 
+html_show_sphinx = True
```
CalMacCQ, line 112: This shows the sphinx version in the footer of the API docs

pytket/docs/requirements.txt
```diff
@@ -1,4 +1,4 @@
-sphinx >= 4.5, < 6.2.0
+sphinx >= 4.5
```
cqc-melf, line 1: Do we want to set a new restriction for example < 7 or so? Just to make sure we don't get surprised by api changes?

CalMacCQ: Sure I can do that

CalMacCQ: Done

pytket/docs/conf.py
CalMacCQ: Turns out this doesn't do anything with our current theme so I removed it.

cqc-melf (approved):

PR #1095, opened 2023-10-26
Remove docs build warning from `pydata-sphinx-theme` 0.14.2 and upgra…

cqc-melf:
We are having some problems with the docs not building on another repo which pulls from main.
I have cherry picked the fix for the build problem, it should work now.

It is not intended to have a release after this is merged.

…de sphinx version (#1093)

* remove unused pygment

* try navigation_with_keys fix

* remove unused intersphinx extension for now

* remove sphinx version restriction

* Show sphinx version in doc footer

* use sphinx <7

* remove show_sphinx flag

### DIFF
--- a/pytket/docs/conf.py
+++ b/pytket/docs/conf.py
@@ -55,7 +55,6 @@ release = "1.21.0"
 extensions = [
     "sphinx.ext.autodoc",
     "sphinx.ext.autosummary",
-    "sphinx.ext.intersphinx",
     "sphinx.ext.mathjax",
     "sphinx.ext.viewcode",
     "sphinx_copybutton",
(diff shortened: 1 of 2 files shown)


cqc-melf:
This PR should be merged via a merged commit, after that we should do another PR from main into develop. That might have the effect that the [7dbc7de](https://github.com/CQCL/tket/pull/1095/commits/7dbc7de6de9808a065c0ff64e6464724021f4966) commit will be listed twice on develop, but I think that should not be an issue.

pytket/docs/conf.py
```diff
@@ -55,7 +55,6 @@
 extensions = [
     "sphinx.ext.autodoc",
     "sphinx.ext.autosummary",
-    "sphinx.ext.intersphinx",
```
CalMacCQ, line 58: I don't think we need to remove this actually. We can probably add this back in.

cqc-melf: If we changes this here now we will have a merge conflict later on, is it okay to keep it like this?

trvto (approved):

pytket/docs/conf.py
CalMacCQ: Yes should be fine.

CalMacCQ (approved):

PR #1105, opened 2023-10-31
Add pytket extensions index page to pytket API docs

CalMacCQ:
Currently the [pytket-extensions repo](https://github.com/CQCL/pytket-extensions) currently only contains the source for a single documentation page.

I was going to put this page in the `tket-site` repository but this requires adding an additonal build stage. I think that adding a whole extra build for a single index page is a bit overkill

 I think it is much simpler to just have the extensions index as part of the pytket API docs. This is already linked from the API docs anyway.
 
 Note that most of the links go to the relevant API docs at the https://tket.quantinuum.com domain.
 
The pytket-qujax, pytket-ionq, pytket-aqt and pytket-qsharp docs are not yet built on the website so the links for these extensions go to the cqcl.github.io site.


CalMacCQ:
Looks like there are some warnings from duplicate links that are casuing the docs build to fail on unbuntu.

pytket/docs/index.rst
```diff
@@ -81,14 +81,14 @@ Licensed under the `Apache 2 License <http://www.apache.org/licenses/LICENSE-2.0
...
-    faqs.rst 
+    faqs.rst
 
 .. toctree::
     :caption: More Documentation:
     :maxdepth: 1
     
     Manual <https://cqcl.github.io/pytket/manual/index.html>
-    Extensions <https://cqcl.github.io/pytket-extensions/api/index.html>
+    extensions_index.rst
```
cqc-melf, line 91: Have you checked if this is showing up as intended? Do we want to update the links back from the extensions to this overview as well?

pytket/docs/extensions_index.rst
```diff
@@ -0,0 +1,165 @@
...
+* **Unitary Simulators** - Unitary simulators calculate the unitary operator that is applied by a circuit. A unitary matrix/ndarray is returned `AerUnitaryBackend <https://tket.quantinuum.com/extensions/pytket-qiskit/api/api.html#pytket.extensions.qiskit.AerUnitaryBackend>`_ is an example of such a simulator.
+* **Density Matrix Simulators** - These simulators compute the density matrix prepared by a circuit. The result can be a statistical mixture of states in contrast to statevector simulation. E.g. `CirqDensityMatrixSampleBackend <https://tket.quantinuum.com/extensions/pytket-cirq/api/api.html#pytket.extensions.cirq.CirqDensityMatrixSampleBackend>`_ .
+* **Other specialised simulators** - There are extensions for simulating specific types of circuit. For instance the `SimplexBackend <https://tket.quantinuum.com/extensions/pytket-pysimplex/api/pysimplex/api.html#pytket.extensions.pysimplex.SimplexBackend>`_ is designed to simulate Clifford circuits. 
+
+A full list of available pytket backends is shown below.
+
+QPUs
+====
+`IBMQBackend <https://tket.quantinuum.com/extensions/pytket-qiskit/api.html#pytket.extensions.qiskit.IBMQBackend>`_
```
cqc-melf, line 32: In which order are the QPUs sorted here?

pytket/docs/index.rst
CalMacCQ: Here's what it looks like before and after I click. I would like to avoid the dropdown as this looks really bad.

Any ideas how to fix this?

I've tried the following
```
.. toctree::
    :caption: More Documentation:
    :maxdepth: 0
```
But that doesn't work

Screenshot 2023-10-31 at 18 50 33

Screenshot 2023-10-31 at 18 50 23

pytket/docs/extensions_index.rst
CalMacCQ: Screenshot 2023-10-31 at 18 55 00

They are now in this order. I have now put Quantinuum first and mentioned that H2-1 is supported and H1-2 is not. Is this correct?

cqc-melf: I thought they are all three now available? I think we should list all three

pytket/docs/index.rst
cqc-melf: The dropdown looks not like the best solution, but showing them always is taking a lot of space also... Do you have any other ideas how to handle this?

pytket/docs/extensions_index.rst
CalMacCQ: H1-2 is retired from my understanding.

Is there another one besides H1-1 and H2-1?

CalMacCQ: I think the H1-2 can be left out.

cqc-melf (approved):

PR #1106, opened 2023-11-01
Make architecture field optional in BackendInfo

yao-cqc:
We should avoid using `Architecture([])` to specify unconstrained architectures for some simulator backends. This PR makes the architecture field in `BackendInfo` optional.

~~I think this is a API breaking change, but I'm not sure whether or how to provide a deprecation path. Any ideas are welcome!~~ I'm not sure if this is an API breaking change. I'm also not so sure about what the correct behaviour of `nodes()` and `n_nodes()` should be when the `architecture` field is missing . Potentially we can allow them to return `None` as well.


cqc-alec (approved):
LGTM. After merging this we should make a release candidate and check if it breaks any of the extensions.

Ideally I would say `nodes` and `n_nodes` should be `None`, but that increases the chances of breaking dependencies' assumptions, so for now I would leave as is (so that they match current values).

PR #1111, opened 2023-11-08
[tkassert] Link tklog as public

cqc-alec:
(See #960 .)


doug-q:
My notes in the prior PR include:

tkassert must be a PUBLIC dependency of tktokenswap because the public
header VectorListHybrid.hpp includes tkassert/Assert.hpp.

Should we include this here?

doug-q (approved):

cqc-alec:
> My notes in the prior PR include:
> 
> tkassert must be a PUBLIC dependency of tktokenswap because the public header VectorListHybrid.hpp includes tkassert/Assert.hpp.
> 
> Should we include this here?

Was planning to do that in a second PR, where we also bump the version of tkassert that tktokenswap depends on.

PR #1120, opened 2023-11-10
Add DummyBox

cqc-alec:


schemas/circuit_v1.json
```diff
@@ -778,6 +778,23 @@
...
+              "type": {
+                "const": "DummyBox"
+              }
+            }
+          },
+          "then": {
+            "required": [
+              "n_qubits",
+              "n_bits",
+              "resource_data"
```
yao-cqc, line 793: I think we might also need to define `resource_data`.

pytket/tests/circuit_test.py
```diff
@@ -1257,6 +1260,98 @@ def test_phase_order() -> None:
...
+    c.add_dummybox(dbox, [0, 2, 3], [1])
+    cmds = c.get_commands()
+    assert len(cmds) == 1
+    op = cmds[0].op
+    assert type(op) is DummyBox
+    resource_data1 = op.get_resource_data()
+    op_type_count = resource_data1.get_op_type_count()
+    assert op_type_count[OpType.T].get_min() == 10
+    assert op_type_count[OpType.T].get_max() == 20
+
```
yao-cqc, line 1281: Could you also call `json_validate` on `c`?

tket/src/Circuit/Circuit.cpp
```diff
@@ -320,4 +328,104 @@ bool Circuit::in_causal_order(
...
+              static_cast<const DummyBox&>(*get_Op_ptr_from_Vertex(v));
+          data = dbox.get_resource_data();
+          for (const auto& pair : data.OpTypeCount) {
+            op_type_count[pair.first].min += pair.second.min;
+            op_type_count[pair.first].max += pair.second.max;
+          }
+        } else {
+          data.GateDepth = {1};
+          data.OpTypeDepth[optype] = {1};
+          if (OpDesc(optype).n_qubits() == 2) {
```
yao-cqc, line 412: How should we treat types with dynamic n_qubits? e.g. CnRy, boxes.

yao-cqc (changes requested):
Looks good! Only a few minor comments

pytket/tests/circuit_test.py
cqc-alec: Done.

tket/src/Circuit/Circuit.cpp
cqc-alec: Changed to check for actual number of qubits.

PR #1121, opened 2023-11-13
Skip non-module libraries when generating stubs.

cqc-alec:
For some reason `pkgutil` is including these libraries in its modules list (with latest version on Linux).

Also apply black formatting (first commit here).


trvto (approved):
👍

trvto:
Could consider moving the mypy checks to macos-13 runners?

trvto:
> Could consider moving the mypy checks to macos-13 runners?

Since they run much faster and would therefore also give earlier feedback

cqc-alec:
> Could consider moving the mypy checks to macos-13 runners?

Yes, that would be good.

PR #1122, opened 2023-11-13
PauliGraph: Core data structure and synthesis for arbitrary circuits

willsimmons1465:
In order to break down the review of content from the PauliGraph rework project whilst accepting that it isn't useful in the main codebase until a critical feature mass is provided, here is a PR into a staging branch where we can accumulate additions we are content with. After a long time in the making (and spinning off into #941 and #1081), this PR covers the base data structure for the new PauliGraph and a naive synthesis method (handling each vertex individually).

This new data structure generalises the existing PauliGraph from just rotations and a Clifford tableau to also include:
- Specialised vertices for Clifford-angled rotations (to help realise Clifford-specific rewrites)
- Measurements
- Decoherence (i.e. OpType::Collapse)
- Resets
- Classically conditioned operations
- Black boxes (abstractions representing an arbitrary operation for which we don't know the semantics and don't apply any rewrites, but still track the subspace in which it acts after conjugations with Clifford circuits)
- Stabiliser assertions
- Clifford tableaux at both inputs and outputs, which can incorporate initialisations, post-selections, and discarding (i.e. built using ChoiMixTableau)

Most notably, with tracking classical data and incorporation of black boxes, passes using this data structure can be applied to any circuit and will require minimal maintenance when OpTypes are added/altered.

The graph tracks non-commutation relations between every pair of Pauli strings (from any vertices) across the graph in a matrix, so that common rewrites can be applied locally within the graph with simple update rules rather than requiring completely rebuilding the graph as before. Classical data dependencies are tracked separately but can similarly be updated locally.

I'm keen to get some feedback on this before committing to it for upcoming improvements.


willsimmons1465:
I have added some extra synthesis methods that use the same logic as the existing PauliGraph implementation, restricted to unitary circuits with end-of-circuit measurements. There may still be subtle differences in output between the two since the old PauliGraph merged rotations on construction whereas merging is left for the rewrite engine in the new refactor. These synthesis methods are there to help maintain similar outputs over the refactor as well as support early experimentation with new techniques that assume this simpler model (or at least aren't immediately obvious as to how to generalise them for arbitrary operations).

willsimmons1465:
Review was sent externally:

> Unfortunately I don't have much feedback to go with my slowness - it basically all looks suitable & good to me! The only real note I have is on file structure - I'd move PGOp and PGBox to separate hpp & cpp files as PauliGraph and PauliGraphConverters are intimidatingly large atm

This is done, so will merge into the staging branch now. Next step: adapting gray-synth for set-wise synthesis of the wide variety of different PGOp kinds

PR #1123, opened 2023-11-13
add job to check all build and test jobs successful

### LINKED ISSUE #1115: Branch protection rules prevent PRs that don't change tket or pytket from being merged
The issue is the "build and test pytket (macos)" job, which is required, and should be skipped in such cases. For some reason, while the job is not run, the PR does not show it as skipped but as waiting for a runner (indefinitely), preventing merging.

For an example see this PR: #1114 .

trvto:
Fixes #1115, also see comments there

Added a job to build and test workflow whose only job is to check that all other jobs did not explicitly fail or were not canceled (skipped jobs are ok). How this fixes the problem:

- After merging, only use the new job as a required check in the github settings -> No longer need to define a required check for the individual jobs
- No longer have to worry about naming issues that occur with matrix strategy jobs

After merging: In a follow-up PR, should be able to remove the dummy job `build_test_tket_not_required` since it should no longer be required

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -577,3 +577,39 @@ jobs:
       run: |
         wget https://cqcl.github.io/tket/pytket/test-coverage/cov.xml -O oldcov.xml
         ./.github/workflows/compare-pytket-coverage oldcov.xml pytket-test-coverage/cov.xml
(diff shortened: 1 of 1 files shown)


cqc-alec (approved):
Thanks!

PR #1124, opened 2023-11-14
remove tket build dummy job

trvto:
Removed no longer necessary job `build_test_tket_not_required`

- Shouldn't forget to change the required checks for the build and test workflow to only include the "All jobs succeeded" job. (I don't have the permissions to do this)


cqc-alec (approved):
I've updated the list of required checks.

PR #1125, opened 2023-11-14
Upload new package only if tket has changed

cqc-alec:


trvto:
Does this cause an error or any other problems if `tket` didn't change?

cqc-alec:
> Does this cause an error or any other problems if `tket` didn't change?

No, it just made the CI runs take longer than necessary.

trvto:
> > Does this cause an error or any other problems if `tket` didn't change?
> 
> No, it just made the CI runs take longer than necessary.

Okay. Because technically, if the conan build command changes, the package that is built could be different even if the `tket` directory doesn't have changes

cqc-alec:
> > > Does this cause an error or any other problems if `tket` didn't change?
> > 
> > 
> > No, it just made the CI runs take longer than necessary.
> 
> Okay. Because technically, if the conan build command changes, the package that is built could be different even if the `tket` directory doesn't have changes

I guess that's true. Maybe we should keep it then. Ideally we'd want to skip it for anything not related to the tket build, but that's probably too awkward to do at the moment.

trvto:
> > > > Does this cause an error or any other problems if `tket` didn't change?
> > > 
> > > 
> > > No, it just made the CI runs take longer than necessary.
> > 
> > 
> > Okay. Because technically, if the conan build command changes, the package that is built could be different even if the `tket` directory doesn't have changes
> 
> I guess that's true. Maybe we should keep it then. Ideally we'd want to skip it for anything not related to the tket build, but that's probably too awkward to do at the moment.

Yes, I mean it's a bit of a weird case but could cause some confusion. One way to cover that could be to move the build/upload part to an external script and add a check on changes to it.

PR #1130, opened 2023-11-16
Drop support for MacOS 11

cqc-alec:
Release workflow tested [here](https://github.com/CQCL/tket/actions/runs/6891015932).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -73,7 +73,7 @@ jobs:
 
   build_macos_wheels:
     name: Build macos wheels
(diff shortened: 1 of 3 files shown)


cqc-melf (commented):
Should we add that somewhere in the dev part of the readme?

cqc-alec:
> Should we add that somewhere in the dev part of the readme?

I don't think it's that relevant to developers; we do sat that apple-clang 14 is the recommended compiler on MacOS.

cqc-melf (approved):

PR #1138, opened 2023-11-21
Add mypy support to nix builds of pytket

jake-arkinstall:
The previous nix derivations for pytket did not include the .pyi files for the _tket directory. This meant that mypy checks could not establish the type information for the binders. With this PR, those files are included and mypy runs successfully. I have thus added mypy checks to the checkPhase of pytket.

This issue was discovered when running mypy checks on a local pytket-qir clone I'm working on adding nix support to.

I have also .gitignored the 'result' directory, in case someone runs `nix build .#[package]` from within the source directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,6 @@ pytket/tests/qasm_test_files/testout6.qasm
 # IDEs
 .idea
 .vscode
+
+#nix
+result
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -31,3 +31,6 @@ pytket/tests/qasm_test_files/testout6.qasm
 # IDEs
 .idea
 .vscode
+
+#nix
```
cqc-alec, line 35: Maybe delete the commented line?

cqc-alec (approved):

PR #1139, opened 2023-11-22
Link to home page in TOC and update extensions

CalMacCQ:
I've added the API docs home to the docs sidebar following -> https://stackoverflow.com/questions/16123951/how-do-i-include-the-homepage-in-the-sphinx-toc

Screenshot 2023-11-22 at 13 47 57

I've also added the qulacs density matrix simulation to the extensions index.

### DIFF
--- a/pytket/docs/index.rst
+++ b/pytket/docs/index.rst
@@ -75,9 +75,10 @@ Licensed under the `Apache 2 License <http://www.apache.org/licenses/LICENSE-2.0
 .. _Quantinuum: https://www.quantinuum.com/
 
 .. toctree::
-    :caption: Introduction:
+    :caption: Overview:
     :maxdepth: 1
 
+    self
     getting_started.rst
     changelog.rst
     install.rst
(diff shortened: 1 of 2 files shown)


pytket/docs/index.rst
```diff
@@ -75,18 +75,19 @@ Licensed under the `Apache 2 License <http://www.apache.org/licenses/LICENSE-2.0
...
 .. toctree::
-    :caption: Introduction:
+    :caption: Overview:
     :maxdepth: 1
 
+    API docs home <self> 
     getting_started.rst
     changelog.rst
     install.rst
+    TKET website <https://tket.quantinuum.com/>
```
cqc-melf, line 85: Why are we not adding extension, examples and manual to this overview?

CalMacCQ: Do you think its more apporiate to merge the "Overview" and "More documentation" lists?

cqc-melf: Yes, but if you have good reasons to not do that, I am happy to discuss that. But I am not sure what our general plan is how and if we want to add the general navigation bar on top?

CalMacCQ: I was thinking we add these links here as an easy near term solution and then figure out the navbar later on.

Issue #1015, opened 2023-09-11
`PauliFrameRandomisation` causes segmentation fault


daniel-mills-cqc:
Example
```
from pytket import OpType
from pytket.tailoring import PauliFrameRandomisation
from pytket.passes.auto_rebase import auto_rebase_pass
from pytket import Circuit

nq = 5
circuit = Circuit(nq)
circuit.add_gate(OpType.CnX, [j for j in range(nq)])    

auto_rebase_pass({OpType.H, OpType.CX, OpType.S, OpType.Rz}).apply(circuit)
PauliFrameRandomisation().sample_circuits(circuit=circuit, samples=5)
```

cqc-alec:
A minimal example illustrating the issue:

```python
from pytket.circuit import Circuit
from pytket.tailoring import PauliFrameRandomisation

circ = Circuit(4).CX(0, 2).Rz(0.5, 0).CX(0, 3).Rz(0.5, 3).CX(0, 3).CX(0, 1)

PauliFrameRandomisation().sample_circuits(circuit=circ, samples=1)
```

balewski:
is there perhaps any progress here?

cqc-alec:
We are investigating this, no fix yet but hopefully in the next release.

PR #1147, opened 2023-11-27
Fix Cycle handling in FrameRandomisation

sjdilkes:
Fixes https://github.com/CQCL/tket/issues/1015.

FrameRandomisations works by first finding a set of subcircuits (or "cycles") of the given set of OpType and then wiring an OpType::noop vertex in the in and out edges of each of these cycles. To add the "frames" for randomisation, these OpType::noop vertices have their type changed.

"Cycles" are found in causal order and "noop frames" are added in causal order. 

The first fix is for finding cycles. The cycle finding works by iterating through the circuit slice-by-slice, adding vertices to cycles if they have the permitted `OpType`. However, the original code had an incorrect assumption that all `Edge` held in a `Slice` object were connected to vertices held in the slice. This is not true and could lead to incorrect cycle objects. This is fixed by confirming that vertices are in a slice before adding to a cycle.

In some cases, a single edge can be the in edge of one cycle and the out edge of another cycle. Given frames are added in causal order, the code had a hardcoded, but not documented, assumption that while adding new noop frames, an encountered in edge to a cycle may have already had a noop vertex added for a previous cycle, and so no longer be valid. To work around this, there is a std::map that provides a look up for which edge should instead have a noop vertex wired into it.

The segmentation fault occurred because it is possible to have two cycles in causal order where the in edge of a cycle encountered first was also the out edge of a cycle encountered later. The second fix is to update the logic for updating in edges to also update out edges.

### DIFF
--- a/tket/src/Characterisation/Cycles.cpp
+++ b/tket/src/Characterisation/Cycles.cpp
@@ -219,7 +219,7 @@ void CycleFinder::order_keys(
             goto new_loop;
           }
 
-  new_loop : {}
+  new_loop: {}
   }
 
   for (const unsigned& key : bad_keys) old_keys.erase(key);
--- a/tket/src/Characterisation/Cycles.cpp
+++ b/tket/src/Characterisation/Cycles.cpp
@@ -170,16 +170,15 @@ std::pair<unsigned, std::set<unsigned>> CycleFinder::make_cycle(
         if (std::find(
                 history_uids.begin(), history_uids.end(), not_present_uid) !=
             history_uids.end()) {
-          // => the not present uid is in this cycle
-          // now we check the overlap
-          //
-          for (const UnitID& a :
-               this->cycle_history.history[candidate_cycle_key]) {
-            for (const UnitID& b : history_uids) {
-              if (a == b) {
-                not_mergeable_keys.insert(candidate_cycle_key);
-              }
-            }
+          std::vector<UnitID> intersection;
+          std::vector<UnitID> candidates =
+              cycle_history.history[candidate_cycle_key];
+          std::set_intersection(
+              candidates.begin(), candidates.end(), history_uids.begin(),
+              history_uids.end(), std::back_inserter(intersection));
+          if (!intersection.empty()) {
+            not_mergeable_keys.insert(candidate_cycle_key);
+            break;
           }
         }
       }
--- a/tket/src/Characterisation/Cycles.cpp
+++ b/tket/src/Characterisation/Cycles.cpp
@@ -79,33 +79,54 @@ void CycleFinder::erase_keys(
   for (const unsigned& key : all_erased) erase_from.erase(key);
 }
 
-// Adds a new cycle to bt.key_to_cycle
+// Adds a new cycle to this->cycle_history.key_to_cycle
 std::pair<unsigned, std::set<unsigned>> CycleFinder::make_cycle(
     const Vertex& v, const EdgeVec& out_edges, const CutFrontier& cut) {
   // Make a new boundary, add it to "all_boundaries", update "boundary_key" map
   std::vector<edge_pair_t> new_cycle_boundary;
-  std::vector<UnitID> new_boundary_uids;
+  std::set<UnitID> new_boundary_uids;
   std::set<unsigned> old_boundary_keys;
-  std::set<unsigned> banned_keys;
+  std::set<unsigned> not_mergeable_keys;
+  std::set<UnitID> not_mergeable_uids;
 
   // Get UnitID, add to uids for new boundary, keep note of all old boundary
   // keys for merging set new boundary key, add new edges to new boundary update
   // cycle out edges if the edge can't be found
+  // std::cout << "\nMake Cycle for " << circ.index_map()[v] << std::endl;
+
   for (const Edge& e : out_edges) {
     UnitID uid = unitid_from_unit_frontier(cut.u_frontier, e);
-    new_boundary_uids.push_back(uid);
-    old_boundary_keys.insert(bt.uid_to_key[uid]);
+    new_boundary_uids.insert(uid);
+    // bool r = this->cycle_history.uid_to_key.find(uid) ==
+    //          this->cycle_history.uid_to_key.end();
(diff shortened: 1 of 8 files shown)


tket/test/src/test_FrameRandomisation.cpp
```diff
@@ -358,5 +360,32 @@ SCENARIO(
...
+  // https://github.com/CQCL/tket/issues/1015
+  PauliFrameRandomisation pfr;
+  Circuit c(4);
+  c.add_op<unsigned>(OpType::CnX, {0, 1, 2, 3});
+  CompilationUnit cu(c);
+  RebaseTket()->apply(cu);
+  std::vector<Circuit> circs =
+      pfr.sample_randomisation_circuits(cu.get_circ_ref(), 1);
+  CHECK(circs.size() == 1);
+  TKET_ASSERT(circs[0].get_commands().size() == 107);
```
cqc-alec, line 387: Shouldn't this be a `CHECK`, rather than ` `TKET_ASSERT`?

sjdilkes: Ah yes, sorry I went quite heavy with the `TKET_ASSERT` while debugging.

sjdilkes: Updated

cqc-alec:
Hmm, I just built pytket from this branch and ran the python from the original bug report [here](https://github.com/CQCL/tket/issues/1015#issue-1890734778), and I got a segmentation fault. (The "minimal example" from the comment below was OK, however.) Looks like we haven't quite got to the bottom of the problem?

sjdilkes:
> Hmm, I just built pytket from this branch and ran the python from the original bug report [here](https://github.com/CQCL/tket/issues/1015#issue-1890734778), and I got a segmentation fault. (The "minimal example" from the comment below was OK, however.) Looks like we haven't quite got to the bottom of the problem?

I'll increase the size of the CnX gate in the tests and see what crops up.

tket/src/Characterisation/Cycles.cpp
```diff
@@ -79,49 +79,119 @@ void CycleFinder::erase_keys(
...
     // i.e. e's associated in edge isn't a cycle out edge
     // i.e. some other non-cycle gate has been passed
+
+    // if the edge going into the vertex is not in a previous cycle
+    // then it can't be merged with that cycle
     if (cycle_out_edges.find(circ.get_last_edge(v, e)) ==
         cycle_out_edges.end()) {
-      banned_keys.insert(bt.uid_to_key[uid]);
+      not_mergeable_keys.insert(this->cycle_history.uid_to_key[uid]);
+      // not_mergeable_uids.insert(uid);
```
cqc-alec, line 109: Delete this commented line?

tket/src/Characterisation/Cycles.cpp
```diff
@@ -79,49 +79,119 @@ void CycleFinder::erase_keys(
...
+    this->cycle_history.uid_to_key[uid] = this->cycle_history.key;
     new_cycle_boundary.push_back({circ.get_last_edge(v, e), e});
     update_cycle_out_edges(uid, e);
   }
 
+  for (const unsigned& key : old_boundary_keys) {
+    for (const UnitID& uid : this->cycle_history.history[key]) {
+      if (not_mergeable_uids.find(uid) != not_mergeable_uids.end()) {
+        not_mergeable_keys.insert(key);
+        // break;
```
cqc-alec, line 120: Delete commented line?

tket/src/Characterisation/Cycles.cpp
```diff
@@ -79,49 +79,119 @@ void CycleFinder::erase_keys(
...
+        continue;
```
cqc-alec, line 121: This `continue;` is unnecessary isn't it?

cqc-alec (commented):
Thanks, just some minor comments.

tket/src/Characterisation/Cycles.cpp
sjdilkes: Removed

tket/src/Characterisation/Cycles.cpp
sjdilkes: Removed

tket/src/Characterisation/Cycles.cpp
sjdilkes: removed

tket/src/Characterisation/Cycles.cpp
```diff
@@ -79,49 +79,119 @@ void CycleFinder::erase_keys(
...
+      if (std::find(
+              candidate_cycle_uid.begin(), candidate_cycle_uid.end(),
+              candidate_uids) == candidate_cycle_uid.end()) {
+        not_present_uids.insert(candidate_uids);
+      }
+    }
+    // not_present_uids now contains UnitID in the new cycle that are not in the
+    // cycle represented by "candidate_cycle_key"
+
+    // We now iterate through all cycles between these UnitID's most recent
```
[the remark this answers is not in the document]
sjdilkes: Changed

tket/src/Characterisation/Cycles.cpp
```diff
@@ -117,8 +116,7 @@ std::pair<unsigned, std::set<unsigned>> CycleFinder::make_cycle(
     for (const UnitID& uid : this->cycle_history.history[key]) {
       if (not_mergeable_uids.find(uid) != not_mergeable_uids.end()) {
         not_mergeable_keys.insert(key);
-        // break;
-        continue;
+        break;
```
cqc-alec, line 119: Did you mean to add the `break;` here? Seems like a change to the logic?

sjdilkes: I meant to have the `break` in the first place, the `continue` was checking for a different error that I forgot about. 

Essentially once `key` is known to be "not_mergeable"  we then don't need to check the remaining UnitID. It ends up not being a change to the logic as `not_mergeable_keys` is a Set: if we don't `continue` or don't have the `break` then it will just check extra UnitID to add the same `key`.

cqc-alec (approved):
OK I see why the `break;` is there now. (Doesn't change the result, just makes it more efficient.)

PR #1149, opened 2023-11-28
Feature/unitarymul

cqc-alec:


cqc-alec:
The coverage check is failing because it isn't counting template code defined in header files. I've added tests that exercise the SVD-computation paths; and the happy paths are tested already by passes such as `FullPeepholeOptimise`, so I'm pretty sure code coverage hasn't gone down.

I've also tested with the enormous circuit that Andrew gave me (previously failing), where the SVD had to be used about 5 times. Compilation was successful.

yao-cqc (approved):
Nice!

PR #1164, opened 2023-12-07
Add PR template.

cqc-alec:
Open to suggestions as to anything else worth including here.


sjdilkes (approved):
LGTM as a starting point!

Issue #922, opened 2023-07-10
[Question] `NoiseAwarePlacement` for single-qubit gates 


1tnguyen:
When using `NoiseAwarePlacement`, if I only have circuits with one-qubit gates, it seems like the  `NoiseAwarePlacement` module doesn't work.

For example, in the below example, I'd have expected it to map the two H gates to qubits 1 and 2 (much lower node errors) but it'd return `unplaced`. 

How should I use  `NoiseAwarePlacement` in these scenarios?

```
std::vector<std::pair<unsigned, unsigned>> edges = {{0, 1}, {1, 2}, {0, 2}};
Architecture architecture(edges);
Circuit circuit(2);
circuit.add_op<unsigned>(OpType::H, {0});
circuit.add_op<unsigned>(OpType::H, {1});

avg_node_errors_t op_node_errors;
op_node_errors[Node(0)] = 0.25;
op_node_errors[Node(1)] = 0.01;
op_node_errors[Node(2)] = 0.01;

DeviceCharacterisation characterisation(op_node_errors, {}, {});
NoiseAwarePlacement placement(architecture);
placement.set_characterisation(characterisation);
auto map = placement.get_placement_map(circuit);

```

sjdilkes:
Hello,

In TKET, when "Mapping" a virtual quantum `Circuit` to a physical quantum circuit for some `Architecture` we use two stages:
1) "Placement", i.e. deciding a mapping between a `Circuit` virtual `Qubit` and a `Architecture` physical `Node`.
2) "Routing", i.e. adding gates such as `SWAP` to permute virtual `Qubit` on `Architecture` `Node` and so make the full circuit valid for the given `Architecture`.

I mention this to try and motivate why a `Placement` method might leave a `Qubit` labelled as `Placement::unplaced_reg()`. Routing is able to dynamically assign virtual `Qubit` to a physical `Node` as a `Circuit` is routed: If a virtual `Qubit` first undergoes a 2-qubit gate deep into the `Circuit` structure, then it can be dynamically assigned to the closest unassigned physical `Node` in the `Architecture` to the other Qubit (assuming it is placed, else both are placed) it is interacting with. In general we find this helps with compilation. Note that if a virtual `Circuit` `Qubit` has no edges, we don't want to assign it to an `Architecture` `Node` that is better placed for some dynamic assignment later, so instead we leave it.

[...]

PR #1165, opened 2023-12-08
Update Noise Aware Placement for Circuits with only single qubit gates

sjdilkes:
https://github.com/CQCL/tket/issues/922 brings up that qubits without 2-qubit gates aren't placed by `NoiseAwarePlacement`. This is fine for `GraphPlacement` as these qubits will be dynamically assigned during routing, with their allocation unrelated to performance as it assumes "no noise", but for `NoiseAwarePlacement` they could still be assigned to `Node` in the `Architecture` that have smaller error rates.

However, the expectation is that typical circuits won't have qubits without any 2-qubit gates, i.e. this is a special case fix. My proposed solution is:
• If the Circuit has qubits with 2-qubit gates and qubits without 2-qubit gates, don't assign the qubits without 2-qubit gates to any `Node`, as these `Node` may be better used as dynamically assigned later.
• If the Circuit **only** has qubits without 2-qubit gates, then assign them to the `Node` with best reported error-rates.

This changes two previous tests that were seeing how `NoiseAwarePlacement` worked on circuits with no 2-qubit gates. In particular, these tests assumed they would be assigned to the `unplaced` register, while now they are assigned to _any_ `Architecture` `Node`. I have considered this a fine change & minor change in behaviour as this not an impactful change for any realistic circuits.

# Checklist

- [X] I have performed a self-review of my code.
- [X] I have commented hard-to-understand parts of my code.
- [X] I have made corresponding changes to the public API documentation.
- [X] I have added tests that prove my fix is effective or that my feature works.
- [X] I have updated the changelog with any user-facing changes.

### DIFF
--- a/pytket/docs/changelog.rst
+++ b/pytket/docs/changelog.rst
@@ -14,12 +14,11 @@ Deprecations:
   
 Fixes:
 
-* Fix `PauliFrameRandomisation.sample_circuits`.
-
-Fixes:
-
 * Ensure that squashing long sequences of gates via unitary multiplication does
   not produce non-unitary results due to rounding errors.
+* Fix `PauliFrameRandomisation.sample_circuits`.
+* For `Circuit` with no 2-qubit gates, `NoiseAwarePlacement` now assigns `Qubit` to `Node` in `Architecture``
+  with lowest reported error rates.
 
 
 1.22.0 (November 2023)
--- a/pytket/docs/changelog.rst
+++ b/pytket/docs/changelog.rst
@@ -17,7 +17,7 @@ Fixes:
 * Ensure that squashing long sequences of gates via unitary multiplication does
   not produce non-unitary results due to rounding errors.
 * Fix `PauliFrameRandomisation.sample_circuits`.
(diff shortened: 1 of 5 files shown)


pytket/docs/changelog.rst
```diff
@@ -14,12 +14,11 @@ Deprecations:
...
 Fixes:
 
-* Fix `PauliFrameRandomisation.sample_circuits`.
-
-Fixes:
-
 * Ensure that squashing long sequences of gates via unitary multiplication does
   not produce non-unitary results due to rounding errors.
+* Fix `PauliFrameRandomisation.sample_circuits`.
+* For `Circuit` with no 2-qubit gates, `NoiseAwarePlacement` now assigns `Qubit` to `Node` in `Architecture``
```
cqc-alec, line 20:
```suggestion
* For `Circuit` with no 2-qubit gates, `NoiseAwarePlacement` now assigns `Qubit` to `Node` in `Architecture`
```

cqc-alec (changes requested):
Just a typo in the changelog.

cqc-alec (approved):

PR #1184, opened 2023-12-18
[feature] add a Circuit method for Reset

ss2165:
Commonly used enough that having to switch to add_gate confuses people.

### DIFF
--- a/pytket/pytket/_tket/circuit.pyi
+++ b/pytket/pytket/_tket/circuit.pyi
@@ -471,6 +471,20 @@ class Circuit:
         :return: the new :py:class:`Circuit`
         """
     @typing.overload
+    def Reset(self, qubit: int, **kwargs: Any) -> Circuit:
+        """
+        Appends a Reset operation. Sets a qubit to the Z-basis 0 state. Non-unitary operation.
+        
+        :return: the new :py:class:`Circuit`
+        """
+    @typing.overload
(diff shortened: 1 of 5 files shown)


cqc-melf (approved):
Are there any examples in which we want to add this?

cqc-melf:
You need to update the version in the binders as well: https://github.com/CQCL/tket/blob/feature/reset-method/pytket/conanfile.py#L35

But I am not sure if that solves the problem.

If you need help I could take a more detailed look later today

pytket/pytket/_tket/circuit.pyi
```diff
@@ -471,20 +471,6 @@ class Circuit:
         :return: the new :py:class:`Circuit`
         """
     @typing.overload
```
cqc-melf, line 473: The stub generation removes this, will look into this again tomorrow.

ss2165: yes I think I mistakenly put those in manually, should be taken out and generated I think?

cqc-melf: Should be working now

cqc-melf (approved):

pytket/pytket/_tket/circuit.pyi
ss2165: thank you!

PR #1212, opened 2024-01-12
Don't build docs in release workflow.

cqc-alec:
# Description

Stop building the docs in the release workflow. They are already built in the PR workflow (to check that they build successfully); and are now published [separately](https://github.com/CQCL-DEV/tket-site/blob/main/.github/workflows/build_and_publish.yml) from the `docmain` branch.

# Checklist


cqc-melf:
I think we are still using this to update https://cqcl.github.io/tket/pytket/api/ which is still the page of the docs google links to? I think it would be good to add a doc build for a push to `docmain` but that is not added yet?

cqc-melf:
Or a better solution would be to put a 301 to redirect and remove everything?

cqc-alec:
> Or a better solution would be to put a 301 to redirect and remove everything?

Yes, I think that's a good idea.

cqc-alec:
> > Or a better solution would be to put a 301 to redirect and remove everything?
> 
> Yes, I think that's a good idea.

Will make a separate PR to do that.

cqc-alec:
> > > Or a better solution would be to put a 301 to redirect and remove everything?
> > 
> > 
> > Yes, I think that's a good idea.
> 
> Will make a separate PR to do that.

#1213

cqc-melf (approved):

PR #1264, opened 2024-02-21
Cherry-pick doc fixes onto `docmain`

CalMacCQ:
# Description

Cherry-picking the fixes from #1258 and #1261 onto `docmain` so we don't have to wait for the next release for the fixes to be in the API docs.

# Checklist

### DIFF
--- a/pytket/conanfile.py
+++ b/pytket/conanfile.py
@@ -32,7 +32,7 @@ class pytketRecipe(ConanFile):
         cmake.install()
 
     def requirements(self):
-        self.requires("tket/1.2.90@tket/stable")
+        self.requires("tket/1.2.91@tket/stable")
         self.requires("tklog/0.3.3@tket/stable")
         self.requires("tkrng/0.3.3@tket/stable")
         self.requires("tkassert/0.3.4@tket/stable")
--- a/pytket/docs/circuit.rst
+++ b/pytket/docs/circuit.rst
@@ -41,6 +41,9 @@ pytket.circuit
 .. autoclass:: pytket.circuit.PauliExpPairBox
     :special-members:
     :members:
+.. autoclass:: pytket.circuit.PauliExpCommutingSetBox
+    :special-members:
+    :members:
 .. autoclass:: pytket.circuit.ToffoliBox
     :special-members:
     :members:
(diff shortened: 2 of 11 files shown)


.github/workflows/build-without-conan.yml
```diff
@@ -85,9 +85,6 @@ jobs:
     - name: Install rapidcheck
       run: |
         cd ${TMP_DIR}
-        wget https://github.com/emil-e/rapidcheck/archive/1c91f40e64d87869250cfb610376c629307bf77d.zip
```
CalMacCQ, line 88: Hmm I don't think the changes to these to `.yml` files should be here.

pytket/conanfile.py
```diff
@@ -32,7 +32,7 @@ def package(self):
         cmake.install()
 
     def requirements(self):
-        self.requires("tket/1.2.90@tket/stable")
```
CalMacCQ, line 35: This change is becasue I changed a link in a code comment which apparently still requrires bumping the version

CalMacCQ, line 35: This change is required because I updated a link in a comment in the source code. Apparently still need a version bump.

cqc-alec:
I don't understand why the "Check tket version bump" step is running, since it is only supposed to run for PRs into `develop`: https://github.com/CQCL/tket/blob/cef4c069383e92c5a03c2642a2536cd0724c1ebb/.github/workflows/build_and_test.yml#L66 .

In any case I think you can just omit the change to the hpp file. It won't affect the generated docs at all. Then you shouldn't need the version bump.

cqc-alec:
Actually, why is this workflow running at all? It should only be run on PRs to `main` and `develop`.

CalMacCQ:
> Actually, why is this workflow running at all? It should only be run on PRs to `main` and `develop`.

I mistakenly made the PR wanting to merge into develop, then switched in the UI. Thats probably why.

cqc-alec:
> Actually, why is this workflow running at all? It should only be run on PRs to `main` and `develop`.

Ah, looks like you originally had the PR going into `develop` which is why the workflow is running. In which case we can just ignore it.

PR #1273, opened 2024-03-06
Update to clang-format 18

cqc-alec:
v17 is no longer available from homebrew and the default has been changed to 18; therefore use v18 and format code accordingly.

### DIFF
--- a/tket/src/Predicates/Predicates.cpp
+++ b/tket/src/Predicates/Predicates.cpp
@@ -52,9 +52,11 @@ static std::string auto_name(const T&) {
   return predicate_name(typeid(T));
 }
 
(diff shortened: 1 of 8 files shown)


tket/src/Predicates/Predicates.cpp
```diff
@@ -52,9 +52,11 @@ static std::string auto_name(const T&) {
   return predicate_name(typeid(T));
 }
 
+#define SET_PRED_NAME(a) \
```
cqc-alec, line 55: Had to move this macro definition up here to work around a bug in the formatter: otherwise it would reformat these lines but then reformat them back again.

cqc-melf (approved):

PR #1285, opened 2024-03-12
Fix subcircuit finding algorithm

### LINKED ISSUE #1279: Occasional proptest failures relating to `gen_clifford_resynthesis_pass()`
E.g. https://github.com/CQCL/tket/actions/runs/8232300297/job/22509445073?pr=1278

cqc-alec:
# Description

The failures were due to the fact that we were implicitly identifying the vertices of the graph constructed by `boost::transitive_closure()` with the vertices of the original graph. I tried to fix this by providing an associative map; but for some reason the resulting generated graph was always the complete graph on all vertices. In the end I implemented transitive closure myself, which we can do fairly efficiently for DAGs if we start with a topological sort.

# Related issues

Fixes #1279 .

# Checklist

### DIFF
--- a/tket/src/Circuit/SubcircuitFinder.cpp
+++ b/tket/src/Circuit/SubcircuitFinder.cpp
@@ -98,18 +98,38 @@ static subcircuit_info_t convex_union(
   return {verts, preds, succs};
 }
 
+static std::set<std::pair<Vertex, Vertex>> order_relations(
+    Circuit *circ /*const*/) {
+  // Put the vertices in reverse topological order:
+  circ->index_vertices();
+  VertexVec verts;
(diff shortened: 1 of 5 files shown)


tket/src/Circuit/SubcircuitFinder.cpp
```diff
@@ -98,18 +98,38 @@ static subcircuit_info_t convex_union(
   return {verts, preds, succs};
 }
 
+static std::set<std::pair<Vertex, Vertex>> order_relations(
+    Circuit *circ /*const*/) {
```
cqc-alec, line 102: "Morally" `const`, but cannot be `const` because `index_vertices()` is not.

yao-cqc (approved):
Looks good

PR #1297, opened 2024-03-19
Remove deprecated `SynthesiseHQS` pass

cqc-alec:
And other associated functions.

Removing `CircPool::two_Rz1` even though it's strictly part of the public API. No extensions are using it, and I'd be amazed if anyone is.


yao-cqc (approved):
Thanks. `two_Rz1` is a strange one :)

PR #1313, opened 2024-03-26
Add Transform.PushCliffordsThroughMeasures

sjdilkes:
Adds a new Transform that reduces Circuit 2qb gate count by resynthesising an end of Circuit Clifford circuit before end of circuit measures. The method works by acting on each Z tensor I measurement operator with the Clifford circuit to produce a new set of measurement operators. As these necessarily commute, they can then be implemented identically to measurement reduction, finding a Clifford circuit that mutually diagonalises them (using pre-existing `mutual_diagonalise` in the c++) and then correcting the measurements via classical post-processing. In this implementation the classical post-processing is implemented in the Circuit using Classical operations.

# Checklist


tket/src/Transformations/CliffordOptimisation.cpp
```diff
@@ -699,6 +705,250 @@ Transform singleq_clifford_sweep() {
...
+
+    // add back in end of circuit measurements
+    TKET_ASSERT(qubits.size() == bits.size());
+    for (unsigned i = 0; i < qubits.size(); i++) {
+      circ.add_measure(qubits[i], bits[i]);
+    }
+
+    // Add classical logic to permute output measurements to correct result
+    register_t scratch_r =
+        circ.add_c_register("permutation_scratch", bits.size() + 1);
```
yao-cqc, line 884: Can we get an exception for "register already exists"? For example, running the pass twice. Maybe we can use `permutation_scratch` just as a prefix.

Also this might be a bit annoying, but could we move this register name into UnitID.hpp/cpp?

yao-cqc (commented):
Thanks, I've left some comments. Any reason not making this a pass?

pytket/binders/transform.cpp
```diff
@@ -414,6 +415,17 @@ PYBIND11_MODULE(transform, m) {
...
+          "PushCliffordsThroughMeasures",
+          &Transforms::push_cliffords_through_measures,
+          "Derives a new set of end-of-Circuit measurement operators "
+          "by acting on end-of-Circuit measurements with a Clifford "
+          "subcircuit. The new set of measurement operators is necessarily "
+          "commuting and is implemented by adding a new mutual diagonalisation "
+          "Clifford subcirciuit to the end of the Circuit and implementing the "
+          "remaining diagonal measurement operators by measuring and permuting "
+          "the "
+          "output.")
```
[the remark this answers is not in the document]
sjdilkes: odd, thanks! will change

tket/src/Transformations/CliffordOptimisation.cpp
sjdilkes: Good idea, I've updated the code toget the permutation_scratch name from `UnitID.hpp` and to call `Circuit::get_next_c_reg_name` to make sure a novel register is used each time.

sjdilkes:
> Thanks, I've left some comments. Any reason not making this a pass?

No good reason - I've also exposed it as a pass now

yao-cqc (approved):
Looks good, thanks!

PR #1319, opened 2024-03-28
Use latest pybind11-stubgen

### LINKED ISSUE #1135: Stub generation fails with pybind11-stubgen 2.4
It produces a lot of errors similar to this:

```
pybind11_stubgen - [  ERROR] In pytket._tket.circuit.Circuit.get_statevector : Can't find/import 'm'
```

cqc-alec:
Fixes #1135 .


cqc-alec:
For some reason, when I run stubgen locally (on Linux) I get one small discrepancy:
```
diff --git a/pytket/pytket/_tket/circuit.pyi b/pytket/pytket/_tket/circuit.pyi
index 56a36cb97..103680cbc 100644
--- a/pytket/pytket/_tket/circuit.pyi
+++ b/pytket/pytket/_tket/circuit.pyi
@@ -942,7 +942,7 @@ class Circuit:
         :param n_bits: The number of classical bits in the circuit
         :param name: Optional name for the circuit.
         """
-    def __iter__(self) -> typing.Iterator[Command]:
+    def __iter__(self) -> typing.Iterator:
         """
         Iterate through the circuit, a Command at a time.
         """
```
But I've updated the stubs to what the CI expects (i.e with `[Command]` in there).

cqc-melf (approved):
Should we take a deeper look on the difference between local and ci runs?

cqc-alec:
> Should we take a deeper look on the difference between local and ci runs?

It would be nice to understand it but I don't know if it's worth spending much time on.

Issue #1202, opened 2024-01-08
Restore "Unitary Synthesis of ChoiMixTableau for Diagonalisation" and fix regression


cqc-alec:
#1201 reverted #941 because it caused some regressions in compilation. For example, with this circuit and sequence of passes:

```python
import json
from pytket.circuit import Circuit, OpType
from pytket.passes import (
    DecomposeBoxes,
    FullPeepholeOptimise,
    auto_rebase_pass,
    auto_squash_pass,
    NormaliseTK2,
    DecomposeTK2,
    RemoveRedundancies,
)

[...]

Issue #1199, opened 2024-01-05
Reported regression in compilation performance from pytket 1.21 to 1.22


cqc-alec:
Apparently the issue reported [here](https://github.com/CQCL/tket/issues/1196) is having an adverse effect on compilation performance (significantly higher two-qubit gate count when compiled with quantinuum backend at optimization level 2).

We need to investigate what is going on.

PR #1320, opened 2024-04-02
Unitary Synthesis of ChoiMixTableau for Diagonalisation (Again)

willsimmons1465:
# Description

#941 implemented a whole bunch of stuff: new features, bug fixes, reorganised code... and some unexpected regression for circuits of pauli gadgets. #1201 reverted the whole PR to temporarily fix the regression. This PR reintroduces the new features, bug fixes, and reorganised code without changing any of the observable behaviour for pytket users.

# Related issues

#1199 and #1202.

Note that in #1202, the examples show that both the old and new synthesis methods had situational benefits over each other. The new behaviour would still be available (toggle the `allow_matching_final` parameter of `reduce_overlap_of_paulis`), though analysis of which is better for which situation is left for future experimentation (likely best done through naively running with both and comparing effectiveness).

# Checklist


tket/src/Converters/ChoiMixTableauConverters.cpp
```diff
@@ -30,637 +30,730 @@ ChoiMixTableau circuit_to_cm_tableau(const Circuit& circ) {
...
-          tab.tab_.zmat_.bottomRows(post_selected_size));
-  // These gates will also swap qubits to isolate the post-selections on the
-  // first few qubits - this is fine as can be cleaned up later with peephole
-  // optimisations
-  for (const std::pair<unsigned, unsigned>& op : col_ops) {
-    tab.tab_.apply_CX(op.first, op.second);
-    ChoiMixTableau::col_key_t ctrl = tab.col_index_.right.at(op.first);
-    ChoiMixTableau::col_key_t trgt = tab.col_index_.right.at(op.second);
-    in_circ.add_op<Qubit>(OpType::CX, {ctrl.first, trgt.first});
+// DEBUG METHOD: Ignore this for coverage checks
```
cqc-alec, line 222: Are we ignoring it?

cqc-alec (commented):
Looks good, just a tiny question on a comment. Thankfully didn't take too long to review as I was mostly able to visually compare this and the previous reviewed PR.

tket/src/Converters/ChoiMixTableauConverters.cpp
willsimmons1465: This comment was more intended for any developer who is scanning through coverage reports to check for ways to improve their coverage. I don't know of any simple method for ignoring methods from gcovr - if you have any suggestions please let me know! I'd rather not have to lift this entire class structure up to a header and move this to test files, as the class is just a helper utility to perform a single method

cqc-alec: Ah I see. You should be able to use exclusion markers: https://gcovr.com/en/stable/guide/exclusion-markers.html .

willsimmons1465: Thanks for the suggestion, I have added the markers and checked it ignores it locally. The CI errors seem to be from a quimb update which looks independent from this PR and will need some looking into

cqc-alec: > Thanks for the suggestion, I have added the markers and checked it ignores it locally. The CI errors seem to be from a quimb update which looks independent from this PR and will need some looking into

Ah yes, we had this issue on the pytket-docs repo as well. Should get this fixed ASAP.

cqc-alec (approved):

PR #1335, opened 2024-04-11
Relax scipy constraint and require latest quimb

cqc-alec:
The quimb 1.8.0 release fixes the issue with scipy 1.13 so we don't need to pin to scipy 1.12 anymore.


cqc-melf (commented):
Should that be added to the changelog?

cqc-alec:
> Should that be added to the changelog?

Done; also relaxed the scipy constraint (forgot to commit this change earlier).

cqc-melf (approved):

PR #1338, opened 2024-04-14
fix: Handle classical control in PhasedXFrontier

### LINKED ISSUE #1306: SIGABRT Error with `GlobalisePhasedX` Pass on Classically Controlled Circuits
### Overview

An issue analogous to the one reported in #1305 occurs when applying the `GlobalisePhasedX` pass to circuits containing classically conditioned gates. 

### Example and reproduction steps

Consider the following Python code:

```python
import pytket
from pytket.passes import GlobalisePhasedX

circuit_tket = pytket.Circuit(2, 1)
circuit_tket.CX(0, 1, condition=pytket.Bit(0))
circuit_tket.Y(0)
circuit_tket.SX(1)
print(GlobalisePhasedX().apply(circuit_tket))
print(circuit_tket.get_commands())
```
#### Observed

Executing the above code results in a critical assertion failure and the process terminates unexpectedly with a SIGABRT signal:

```
True
[tket] [critical] Assertion '!"slice is empty"' (../../src/Circuit/macro_circ_info.cpp : operator++ : 935) failed.  Aborting.

Process finished with exit code 134 (interrupted by signal 6:SIGABRT)
```

lmondada:
Closes #1306

This solves the issue by treating any classically controlled gate as a barrier when squashing and turning gates to global gates. This is the only simple solution I could think of, as squashing classically controlled operation is always a bit tricky and comes with many edge cases.

### DIFF
--- a/pytket/conanfile.py
+++ b/pytket/conanfile.py
@@ -32,7 +32,7 @@ class pytketRecipe(ConanFile):
         cmake.install()
 
     def requirements(self):
-        self.requires("tket/1.2.114@tket/stable")
+        self.requires("tket/1.2.115@tket/stable")
         self.requires("tklog/0.3.3@tket/stable")
         self.requires("tkrng/0.3.3@tket/stable")
         self.requires("tkassert/0.3.4@tket/stable")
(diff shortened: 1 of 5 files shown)


cqc-alec (approved):
Thanks!

PR #1365, opened 2024-04-27
ci: Add automated benchmarking workflow

daniel-mills-cqc:
# Description

Adds a workflow to run benchmarks on each new PR. The version of pytket built from the PR is used in the benchmarks. At the moment this is a simple comparison to Qiskit on quantum volume circuits.

I would like to have a wider conversation about the compilers to compare agains, and the circuits to compile. The outcome of those conversation would feature in later PRs.

I also note that the benchmarks are quite slow. I will experiment with parallelisation in future PRs.

# Related issues

N/A

# Checklist

### DIFF
--- a/.github/workflows/pytket_benchmarking.yml
+++ b/.github/workflows/pytket_benchmarking.yml
@@ -78,11 +78,6 @@ jobs:
           pip install pytket-*.whl
           pip install --pre --index-url https://github_actions:${{ secrets.PRIVATE_PYPI_PASS }}@cqcpythonrepository.azurewebsites.net/simple/ pytket_benchmarking
 
-      - name: Checkout tket
-        uses: actions/checkout@v4
-        with:
-          path: tket
-
       - name: Checkout pytket-benchmarking-store
         uses: actions/checkout@v4
         with:
--- a/.github/workflows/pytket_benchmarking.yml
+++ b/.github/workflows/pytket_benchmarking.yml
@@ -56,8 +56,8 @@ jobs:
         name: pytket_wheel
         path: wheelhouse/
 
-  Write-Env-Var:
-    name: Write Env Var
+  compile-and-compare:
+    name: Compile and compare
     runs-on: macos-14
     needs: build_wheels
 
@@ -68,7 +68,7 @@ jobs:
         with:
           python-version: '3.11'
 
-      - name: Download Wheel
+      - name: Download wheel
         uses: actions/download-artifact@v4
         with:
           name: pytket_wheel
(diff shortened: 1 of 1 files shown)


.github/workflows/pytket_benchmarking.yml
```diff
@@ -0,0 +1,112 @@
...
+
+      - name: Install dependencies
+        run: |
+          pip install pytket-*.whl
+          pip install --pre --index-url https://github_actions:${{ secrets.PRIVATE_PYPI_PASS }}@cqcpythonrepository.azurewebsites.net/simple/ pytket_benchmarking
+
+      - name: Checkout tket
+        uses: actions/checkout@v4
+        with:
+          path: tket
```
cqc-alec, line 84: Why is this necessary, as we are just using the downloaded wheel?

.github/workflows/pytket_benchmarking.yml
```diff
@@ -0,0 +1,112 @@
...
+    - name: Build tket C++
+      run: conan create tket --user tket --channel stable --build=missing -o boost/*:header_only=True -o tklog/*:shared=True -o tket/*:shared=True -tf ""
+
+    - name: Build wheel
+      run: |
+        conan create recipes/pybind11
+        conan create recipes/pybind11_json/all --version 0.2.13
+        cd pytket
+        # Ensure wheels are compatible with MacOS 12.0 and later:
+        export WHEEL_PLAT_NAME=macosx_12_0_arm64
```
cqc-alec, line 48: This shouldn't be necessary as we're testing on `macos-14`.

cqc-alec (commented):
Looks good, just a couple of comments.

.github/workflows/pytket_benchmarking.yml
daniel-mills-cqc: Ah yes, a bit of a left over from testing. Cheers!

.github/workflows/pytket_benchmarking.yml
daniel-mills-cqc: Would you mind saying a little bit more about that. I'm afraid I'm not so familiar with the steps required to build the wheels.
I tried removing these two lines but see that the build then fails. I'm not sure if I should have expected that?

cqc-alec: Hmm, that's a surprise. Ah well, we can keep the lines there, it's not an issue.

cqc-alec (approved):

ferbetanzo (approved):
LGTM!

PR #1371, opened 2024-05-03
Switch workflows to target `main` only

cqc-alec:
# Description

Update CI workflows to run on push and PR to `main` instead of `develop`.

Once approved, I would:
- change the branch protection rules on the repo (note that some of the "Required" checks below will not run on this PR);
- merge this into `develop`;
- make `main` the default branch;
- make a PR from `develop` into `main`;
- merge that.

We will also need to update the workflows on `tket-site`, so that they only publish released versions (which will be identifiable by tag).

# Checklist

### DIFF
--- a/pytket/docs/backends.rst
+++ b/pytket/docs/backends.rst
@@ -5,7 +5,7 @@ Contains :py:class:`Backend` abstract class and associated methods. In pytket a 
 
 There are several `example notebooks <https://tket.quantinuum.com/examples>`_ on pytket :py:class:`Backend`\s. If you are interested in developing your own :py:class:`Backend` or pytket extension then see the `creating backends <https://tket.quantinuum.com/examples/creating_backends.html>`_ tutorial.
 
-Notebook tutorials specific to the :py:class:`QuantinuumBackend` can be found `here <https://github.com/CQCL/pytket-quantinuum/tree/main/examples>`_.
+Notebook tutorials specific to the :py:class:`QuantinuumBackend` can be found `here <https://github.com/CQCL/pytket-quantinuum/tree/develop/examples>`_.
 
 See also the `Running on backends <https://tket.quantinuum.com/user-manual/manual_backend.html>`_ section of the pytket user manual.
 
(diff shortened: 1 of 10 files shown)


pytket/docs/backends.rst
```diff
@@ -5,7 +5,7 @@ Contains :py:class:`Backend` abstract class and associated methods. In pytket a
 
 There are several `example notebooks <https://tket.quantinuum.com/examples>`_ on pytket :py:class:`Backend`\s. If you are interested in developing your own :py:class:`Backend` or pytket extension then see the `creating backends <https://tket.quantinuum.com/examples/creating_backends.html>`_ tutorial.
 
-Notebook tutorials specific to the :py:class:`QuantinuumBackend` can be found `here <https://github.com/CQCL/pytket-quantinuum/tree/develop/examples>`_.
+Notebook tutorials specific to the :py:class:`QuantinuumBackend` can be found `here <https://github.com/CQCL/pytket-quantinuum/tree/main/examples>`_.
```
cqc-melf, line 8: Is this change intended? We have not updated that repo yet?

cqc-melf (commented):
can you maybe outline what the plan for the releases is?
Do we want to create a release from main or from a release branch?

cqc-melf:
Have you checked if there are any open PRs that might be effected by this?

pytket/docs/backends.rst
cqc-alec: You're right, that was a mistake, I'll revert it.

cqc-alec:
> can you maybe outline what the plan for the releases is? Do we want to create a release from main or from a release branch?

We can do either, but normally we will just release from main, as at present. If we want to do a hotfix release that doesn't have all the commits from main then we will create a special release branch for that.

cqc-alec:
> Have you checked if there are any open PRs that might be effected by this?

This should just be a matter of changing the target branch in the PR.

cqc-melf (approved):

PR #1373, opened 2024-05-03
Show diff from clang-format.

cqc-alec:
# Description

Previously the formatting check would fail without indicating exactly what was wrong. This PR makes it print a diff which the developer can then apply to their branch to fix the issues.

This became necessary since clang-format began changing its rules on updates to minor versions that are not easily available on all platforms.

Also actually fix the latest formatting errors.

# Checklist


sjdilkes (approved):
Big approve!

PR #1386, opened 2024-05-09
Use `lark` package instead of deprecated `lark-parser`

### LINKED ISSUE #1356: Updating `lark-parser` depedency to `lark`
Hi,

I am using `pytket` in a project where also `lark` is an dependency additionally to `lark-parser` that is required by `pytket`. Here, I face some compatibility issues resulting in errors since apparently having both `lark` and `lark-parser` installed with different versions is problematic. 

Therefore, I was wondering whether the (apparently deprecated) `lark-parser` dependency could be upgraded to `lark` as well.

cqc-alec:
# Description

Needs a couple of small tweaks to the qasm and quipper parsers.

# Related issues

Closes #1356 .

# Checklist

### DIFF
--- a/nix-support/pytket.nix
+++ b/nix-support/pytket.nix
@@ -39,7 +39,7 @@ in {
     inherit version;
     propagatedBuildInputs = with super.python3.pkgs; [
       self.binders
-      super.lark-parser
+      super.lark
       super.types-pkg_resources
       super.qwasm
       graphviz
--- a/nix-support/third-party-python-packages.nix
+++ b/nix-support/third-party-python-packages.nix
@@ -19,15 +19,17 @@ self: super: {
       sha256 = "sha256:g/QA5CpAR3exRDgVQMnXGIH8bEGtwGFBjjSblbdXRkU=";
     };
   };
-  lark-parser = super.python3.pkgs.buildPythonPackage {
-    pname = "lark-parser";
-    version = "0.12.0";
+  lark = super.python3.pkgs.buildPythonPackage {
+    pname = "lark";
+    version = "1.1.9";
+    format = "pyproject";
     src = super.fetchFromGitHub {
       owner = "lark-parser";
       repo = "lark";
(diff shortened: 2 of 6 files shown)


jake-arkinstall:
Keeping the old hash can confuse nix and make it use the wrong package - it's usually best to give the hash an empty string on an update. Then the build fails, but it tells you what the hash should have been.

Additionally the latest versions of lark use pyproject formatting rather than a setup.py process, so buildPythonPackage needs to be told to use that. I've adjusted that and committed the change.

cqc-alec:
> Keeping the old hash can confuse nix and make it use the wrong package - it's usually best to give the hash an empty string on an update. Then the build fails, but it tells you what the hash should have been.
> 
> Additionally the latest versions of lark use pyproject formatting rather than a setup.py process, so buildPythonPackage needs to be told to use that. I've adjusted that and committed the change.

@jake-arkinstall-quantinuum  Oops I think I just force-pushed over your change sorry. I was hoping it would tell me the correct hash. Please could you fix again? Thanks!

jake-arkinstall:
> > Keeping the old hash can confuse nix and make it use the wrong package - it's usually best to give the hash an empty string on an update. Then the build fails, but it tells you what the hash should have been.
> > Additionally the latest versions of lark use pyproject formatting rather than a setup.py process, so buildPythonPackage needs to be told to use that. I've adjusted that and committed the change.
> 
> @jake-arkinstall-quantinuum Oops I think I just force-pushed over your change sorry. I was hoping it would tell me the correct hash. Please could you fix again? Thanks!

No worries, I should have warned you! I'll patch it now.

cqc-melf:
Happy to approve this, are the nix questions sorted @jake-arkinstall @cqc-alec ?

jake-arkinstall:
Yes, I am happy with the nix side here

cqc-melf (approved):

PR #1438, opened 2024-06-06
Run ZX test with all Python versions

### LINKED ISSUE #1242: `zx_diagram_test` aborts when run on `ubuntu-22.04` with python 3.12
OK on other platforms.

cqc-alec:
# Description

Tested here: https://github.com/CQCL/tket/actions/runs/9396857013/job/25878869049

There was an issue with the ZX test running on Ubuntu with Python 3.12 but this seems to be working now.

# Related issues

Closes #1242 .

# Checklist

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -282,15 +282,9 @@ jobs:
       with:
         python-version: '3.12'
     - name: Build pytket
-      if: github.event_name != 'schedule'
       run: |
         cd pytket
         pip install -e .[ZX] -v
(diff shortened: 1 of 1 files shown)


cqc-melf (approved):

cqc-melf:
I think it would be good to update the filter, so that all tests are run when there are changes in the workflows?

cqc-alec:
> I think it would be good to update the filter, so that all tests are run when there are changes in the workflows?

I think this is already the case? (But PR runs don't use Python 3.12, hence the need for a manual test.)

PR #1470, opened 2024-06-26
Update to pybind11 2.13.0.

cqc-alec:


cqc-melf (approved):

jake-arkinstall:
At current, the nix version still runs pybind11-2.12, and this is the latest version in nixpkgs. If there is an upcoming need for this version update, let me know and I will package it manually.

PR #1471, opened 2024-06-26
Make compatible with numpy 2.0

cqc-alec:
Update version requirements on dependencies, removing upper bounds.

A few fixes to allow compatibility with numpy 2.0.


pytket/docs/changelog.rst
```diff
@@ -7,6 +7,7 @@ Unreleased
 * Support classical transforms and predicates, and QASM registers, with up to 64
   bits. Add an attribute to the pytket module to assert this.
 * Add `PauliSynthStrat.Greedy` strategy to `TermSequenceBox`.
+* Update version requirements on dependencies, including allowing numpy 2.0.
```
cqc-melf, line 10: Maybe we should specify this to make sure it is understood what this means?

pytket/setup.py
```diff
@@ -190,21 +190,22 @@ def finalize_options(self):
     license="Apache 2",
     packages=setuptools.find_packages() + ["pytket.qasm.includes"],
     install_requires=[
-        "sympy ~=1.6",
-        "numpy >=1.21.4, <2.0",
-        "lark ~=1.1",
-        "scipy ~=1.13",
+        "sympy >= 1.12.1",
```
cqc-melf, line 193: Are we sure that we want to do this now and not wait a bit how this works on pytket-qiskit?

cqc-alec: I actually think it's worth doing now. Updating the lower bounds is something we've always done as a matter of course (when we remember), and removing the upper bounds (and fixing issues as they arise) is better than forcing users to have a lower version than they need (and than may be required by other packages).

pytket/docs/changelog.rst
cqc-alec: Do you mean give the exact new version requirements on all the packages? Happy to do that.

cqc-melf: I mean to point out that there are no upper bounds any more for all packages. (In case any new updates break something it should be clear that it might be possible to fix it by downgrading dependencies)

pytket/setup.py
```diff
@@ -190,21 +190,22 @@ def finalize_options(self):
...
+        "numpy >= 1.26.4",
```
cqc-melf, line 194: Have you tested if everything still works with 1.26? What is our future testing plan? When using feature only available in 2.0 the ci would not show any problems?

cqc-alec: Yes I have checked this locally.

This is a question we face already with other packages (e.g. how do we know it still works with sympy 1.6). I don't think there's an easy answer; we can't test every version of everything. This is part of the trade-off between full determinism and compatibility... But from what I've seen of the changes in numpy 2.0, I think it's unlikely.

cqc-melf (approved):

PR #1476, opened 2024-07-02
Ae/fix cliffresynth

cqc-alec:
# Description

Please summarise the changes.

# Related issues

# Checklist


tket/src/Transformations/CliffordResynthesis.cpp
```diff
@@ -49,18 +54,46 @@ static Circuit resynthesised_circuit(
...
+}
+
 static bool resynthesise_cliffords(
     Circuit &circ,
     std::optional<std::function<Circuit(const Circuit &)>> transform = nullptr,
     bool allow_swaps = true) {
   bool changed = false;
-  for (const VertexSet &verts :
-       circ.get_subcircuits([](Op_ptr op) { return op->is_clifford(); })) {
+  while (true) {
```
cqc-melf, line 74: Should this maybe better contain the break condition in the while line?

cqc-alec: It's a matter of preference. That would mean introducing an extra boolean variable outside the loop to control it, and modifying the variable inside the loop, whereas this way we don't need to do that. On balance I think the function is sufficiently small that the logic is apparent.

cqc-melf (approved):

PR #1480, opened 2024-07-04
Remove deprecated `SynthesiseOQC` and deprecate `SynthesiseUMD`

### LINKED ISSUE #1472: Remove `SynthesiseOQC`, `RebaseUFR` and `DecomposeBridges`; deprecate `SynthesiseUMD`

cqc-alec:
Also remove deprecated `ConstPredicate`.

Not removing `RebaseUFR` as it's used internally, or `DecomposeBridges` as it seems potentially useful even if we don't use it anywhere except in tests.

Closes #1472 .


pytket/pytket/passes/auto_rebase.py
```diff
@@ -19,7 +19,10 @@
 
 
 class NoAutoRebase(Exception):
-    """Automatic rebase could not be found."""
+    """Automatic rebase could not be found.
+
+    "DEPRECATED: will be removed after pytket 1.33."
```
sjdilkes, line 24: Other deprecations have suggested pytket 1.32 while this says pytket 1.33 - is this deliberate?

tket/src/Transformations/Decomposition.cpp
```diff
@@ -420,23 +420,6 @@ Transform decompose_tk1_to_rzrx() {
   });
 }
 
-Transform decompose_CX_to_ECR() {
```
sjdilkes, line 423: Could it be worth keeping this still?

sjdilkes (changes requested):
Just a couple of queries

pytket/pytket/passes/auto_rebase.py
cqc-alec: Yeah just wanted to give a bit longer since some of the extensions still use `auto_rebase_pass`.

tket/src/Transformations/Decomposition.cpp
cqc-alec: It's not used anwhere and we can easily restore if needed, so I think it's fine to go.

sjdilkes (approved):

PR #1481, opened 2024-07-04
only remove empty wires at the end of a register

cqc-melf:
# Description

Solves https://github.com/CQCL/tket/issues/1458

Update `FlattenRelabelRegistersPass` and `remove_blank_wires` to only  remove unused classical bits at the end of a register

# Related issues

# Checklist


tket/src/Circuit/basic_circ_manip.cpp
```diff
@@ -33,27 +33,51 @@ namespace tket {
 // this method removes them and removes the vertices
 // from boundaries
```
cqc-alec, line 34: Please correct this comment as it isn't quite accurate any more.

tket/src/Circuit/basic_circ_manip.cpp
```diff
@@ -33,27 +33,51 @@ namespace tket {
...
+          }
+        }
+
+        if (remove_op) {
+          dag[in].op = noop;
+          bin.push_back(in);
+          dag[out].op = noop;
+          bin.push_back(out);
+          unused_units.push_back(el.id_);
+        }
```
cqc-alec, line 73:
```suggestion
        // check if unitid is an unused bit in a register of dim 1
        if (succs.front() == out && succs.size() == 1 &&
            el.type() == UnitType::Bit && el.id_.reg_dim() == 1) {
          // check if the empty bit is at the end of a register
          if (get_reg(el.id_.reg_name()).size() == (el.id_.index()[0] + 1)) {
            // rerun while loop
            found_empty_bit_at_end = true;
            // remove op
            dag[in].op = noop;
            bin.push_back(in);
            dag[out].op = noop;
            bin.push_back(out);
            unused_units.push_back(el.id_);
          }
        }
```
I would get rid of the `remove_op` variable and rewrite like this (if we keep this logic).

tket/src/Circuit/basic_circ_manip.cpp
```diff
@@ -33,27 +33,51 @@ namespace tket {
...
-      Vertex out = el.out_;
-      VertexVec succs = get_successors(in);
-      if (succs.size() == 1 && succs.front() == out) {
-        dag[in].op = noop;
-        bin.push_back(in);
-        dag[out].op = noop;
-        bin.push_back(out);
-        unused_units.push_back(el.id_);
+  bool found_empty_bit_at_end = true;
+  while (found_empty_bit_at_end) {
```
cqc-alec, line 37: The logic here is quite hard to understand, and seems inefficient: we are looping over all the boundary elements multiple times. Is there a better way, perhaps iterating over registers and removing unused bits from each of them working backwards from the end?

tket/src/Circuit/basic_circ_manip.cpp
cqc-melf: This is now outdated, see below

tket/src/Circuit/basic_circ_manip.cpp
cqc-melf: I have tried to do this in a different way to make it easier to understand. I think `find` is still iterating some part of the boundary, so I would assume that this is a bit less sufficient then the other option for large circuits.

cqc-melf: See [cc8d629](https://github.com/CQCL/tket/pull/1481/commits/cc8d6293e567307084a5887029219c795d29e9db)

cqc-melf: see @cqc-melf
[update loop over bitnames](https://github.com/CQCL/tket/pull/1481/commits/9193185e984d8d2a923bd1d60670f8c928d3d004)
[9193185](https://github.com/CQCL/tket/pull/1481/commits/9193185e984d8d2a923bd1d60670f8c928d3d004)

tket/src/Circuit/basic_circ_manip.cpp
cqc-melf: done in @cqc-melf
[update comment in void Circuit::remove_blank_wires](https://github.com/CQCL/tket/pull/1481/commits/cc8d6293e567307084a5887029219c795d29e9db)
[cc8d629](https://github.com/CQCL/tket/pull/1481/commits/cc8d6293e567307084a5887029219c795d29e9db)

cqc-alec (approved):
Thanks!

PR #1492, opened 2024-07-10
Update `docmain` for pytket 1.30 release

CalMacCQ:
# Description

Updating `docmain` from the latest `main` for the pytket 1.30.0 release.


pytket/docs/conf.py
```diff
@@ -43,7 +43,6 @@
 release = "1.30.0"
 
 
-
```
cqc-melf, line 45: are this changes that are not on main?

cqc-melf (commented):
Might be better to resolve the issues with black on main and not here, because that would cause further issues next time? Is the docmain branch protected? It might be good to just force push the current state of main?

CalMacCQ:
> Might be better to resolve the issues with black on main and not here, because that would cause further issues next time? Is the docmain branch protected? It might be good to just force push the current state of main?

Yeah the `docmain` branch is protected.

cqc-melf:
There are still differences between the update branch and `main` is that intended? https://github.com/CQCL/tket/compare/main...update_docs/1.30

PR #1503, opened 2024-07-29
feat: Refactor WASM module

johnchildren:
- Add a new WasmModuleHandler class that takes raw wasm_module bytes on construction, but has the same interface as WasmFileHandler.
- WasmFileHandler now inherits from WasmModuleHandler.
- Split checking of the WASM module into a new function so it can be called after initialisation if required.
- Adds new cached properties, bytecode, bytecode_base64 and module_uid which allow lazy loading of the module in encoded form, or computation of a unique identifier for the module.
- Add deprecated properties _check_file, wasm_file_encoded and _wasmfileuid to maintain compatibility with existing versions of pytket-qir and pytket-quantinuum.
- Remove some exceptions that are no longer possible.

# Description

The primary intention behind this PR is to allow usage of WASM modules without having to access the file system, enabling use-cases such as loading WASM from a network location and using it in a workflow. However I also ended up making a lot more changes to discourage usage of private attributes of instances of the class while ensuring that the class stays compatible with existing versions of pytket-quantinuum and pytket-qir. These packages should also be updated to use the public versions of the properties that I have added instead to guarantee a stable API.

Additionally I have made features like the unique identifier and base64 encoded bytecode optional functionality that are supplied by cached properties. Hopefully this should allow users who do not need these features to avoid them but I am very open to feedback on this approach.

# Related issues

# Checklist

### DIFF
--- a/pytket/pytket/circuit/__init__.py
+++ b/pytket/pytket/circuit/__init__.py
@@ -135,7 +135,7 @@ def overload_add_wasm_to_reg(
     if args_wasm is None:
         args_wasm = [0]
 
-    if filehandler._check_file:
+    if filehandler.checked:
         for reg in list_i:
             if reg.size > 32:
                 raise ValueError(
--- a/pytket/pytket/circuit/__init__.py
+++ b/pytket/pytket/circuit/__init__.py
@@ -150,7 +150,11 @@ please use only registers of at most 32 bits"""
 please use only registers of at most 32 bits"""
                 )
 
-    if filehandler.check_function(funcname, len(list_i), len(list_o)):
+    # If the filehandler has not been checked we allow it to
+    # be added without checking the function arity.
+    if not filehandler.checked or filehandler.check_function(
+        funcname, len(list_i), len(list_o)
+    ):
         if (len(args_wasm)) > 0:
             self._add_w_register(max(args_wasm) + 1)
         return self._add_wasm(
--- a/pytket/pytket/wasm/wasm.py
+++ b/pytket/pytket/wasm/wasm.py
@@ -266,7 +266,8 @@ class WasmModuleHandler:
         :return: true if the signature and the name of the function is correct"""
         if not self.checked:
             raise ValueError(
-                "Cannot retrieve functions from an unchecked wasm module. Please call .check() first."
+                "Cannot retrieve functions from an unchecked wasm module."
+                " Please call .check() first."
             )
 
         return (
@@ -284,7 +285,8 @@ class WasmModuleHandler:
         """
         if not self.checked:
             raise ValueError(
-                "Cannot retrieve functions from an unchecked wasm module. Please call .check() first."
+                "Cannot retrieve functions from an unchecked wasm module."
+                " Please call .check() first."
             )
         return self._functions
 
@@ -297,7 +299,8 @@ class WasmModuleHandler:
         """
         if not self.checked:
             raise ValueError(
-                "Cannot retrieve functions from an unchecked wasm module. Please call .check() first."
+                "Cannot retrieve functions from an unchecked wasm module."
+                " Please call .check() first."
             )
         return self._unsupported_functions
 
--- a/pytket/pytket/wasm/wasm.py
+++ b/pytket/pytket/wasm/wasm.py
@@ -32,7 +32,7 @@ from qwasm import (  # type: ignore
 
 
 class WasmModuleHandler:
-    """Construct and optionally check a WASM module for use in Pytket WASM Ops."""
+    """Construct and optionally check a WASM module for use in WASM Ops."""
 
     checked: bool
     _int_size: int
@@ -284,7 +284,7 @@ class WasmModuleHandler:
 
 
 class WasmFileHandler(WasmModuleHandler):
-    """Construct and optionally check a WASM module from a file for use in Pytket WASM Ops."""
+    """Construct and optionally check a WASM module from a file for use in WASM Ops."""
 
     def __init__(self, filepath: str, check_file: bool = True, int_size: int = 32):
         """
--- a/pytket/pytket/wasm/wasm.py
+++ b/pytket/pytket/wasm/wasm.py
@@ -32,7 +32,7 @@ from qwasm import (  # type: ignore
 
 
 class WasmModuleHandler:
-    """Construct and optionally check a WASM module for use in WASM Ops."""
+    """Construct and optionally check a wasm module for use in wasm Ops."""
 
     checked: bool
     _int_size: int
@@ -54,7 +54,7 @@ class WasmModuleHandler:
         """
         Construct a wasm module handler
 
-        :param wasm_module: A WASM module in binary format.
+        :param wasm_module: A wasm module in binary format.
         :type wasm_module: bytes
         :param check: If ``True`` checks file for compatibility with wasm
           standards. If ``False`` checks are skipped.
@@ -96,7 +96,7 @@ class WasmModuleHandler:
         """
         if self.checked:
             return
-        
+
         function_signatures: list = []
         function_names: list = []
         _func_lookup = {}
@@ -195,6 +195,7 @@ class WasmModuleHandler:
         # to be checked again.
         self.checked = True
 
+    @property
     @deprecated("Use public property `checked` instead.")
     def _check_file(self) -> bool:
         return self.checked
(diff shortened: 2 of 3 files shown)


pytket/pytket/circuit/__init__.py
```diff
@@ -135,7 +135,7 @@ def overload_add_wasm_to_reg(
     if args_wasm is None:
         args_wasm = [0]
 
-    if filehandler._check_file:
+    if filehandler.checked:
```
johnchildren, line 138: I guess this might be semantically different and therefore not correct as a change

pytket/pytket/wasm/wasm.py
```diff
@@ -42,13 +48,15 @@ class WasmFileHandler:
         LANG_TYPE_EMPTY: None,
     }
 
-    def __init__(self, filepath: str, check_file: bool = True, int_size: int = 32):
```
johnchildren, line 45: changing check_file to check is actually a breaking change and I probably should not have done this.

johnchildren: ah it's actually still called `check_file` for WasmFileHandler so it could be fine? A bit confusing however

pytket/pytket/wasm/wasm.py
```diff
@@ -236,8 +256,52 @@ def check_function(
...
+            self.check()
+        return self._unsupported_functions
+
+
+class WasmFileHandler(WasmModuleHandler):
+    """Construct and optionally check a WASM module from a file for use in Pytket WASM Ops."""
+
+    def __init__(self, filepath: str, check_file: bool = True, int_size: int = 32):
+        """
+        Construct a wasm file handler
```
johnchildren, line 291: This comment needs expansion

pytket/pytket/wasm/wasm.py
```diff
@@ -64,169 +72,181 @@ def __init__(self, filepath: str, check_file: bool = True, int_size: int = 32):
...
-                    f"unsupported function with invalid "
-                    f"parameter or result type: '{x}' \n"
-                )
-
-            return result
-        else:
-            raise ValueError(
-                """the content of the wasm file can only be printed if the
-wasm file was checked"""
+        Will implicitly check the module if it has not been checked already.
```
johnchildren, line 203: Not sure if this is desirable behavior, but it seemed more useful than just throwing an exception

pytket/pytket/wasm/wasm.py
```diff
@@ -64,169 +72,181 @@ def __init__(self, filepath: str, check_file: bool = True, int_size: int = 32):
...
+            raise ValueError("init function should not have any parameter")
 
-            if self._functions["init"][0] != 0:
-                raise ValueError("init function should not have any parameter")
+        if self._functions["init"][1] != 0:
+            raise ValueError("init function should not have any results")
 
-            if self._functions["init"][1] != 0:
-                raise ValueError("init function should not have any results")
+        self.checked = True
```
johnchildren, line 190: I probably should include a comment about the importance of this line

pytket/pytket/wasm/wasm.py
```diff
@@ -236,8 +256,52 @@ def check_function(
...
             and (self._functions[function_name][1] == number_of_returns)
         )
+
+    @cached_property
+    def functions(self) -> dict[str, tuple[int, int]]:
+        """Retrieve the names of functions with the number of input and out arguments.
+
+        Will check the module if it has not been checked already.
+        """
+        if not self.checked:
```
johnchildren, line 271: Possibly this block could be moved into check() itself but it seemed a bit more explicit to do it here, open to suggestions.

pytket/pytket/wasm/wasm.py
```diff
@@ -64,169 +72,181 @@ def __init__(self, filepath: str, check_file: bool = True, int_size: int = 32):
...
+    def bytecode_base64(self) -> bytes:
+        """The WASM content as base64 encoded bytecode."""
+        return base64.b64encode(self._wasm_module)
+
+    @deprecated("Use public property bytecode_base64 instead.")
+    def _wasm_file_encoded(self) -> bytes:
+        return self.bytecode_base64
+
+    @cached_property
+    def module_uid(self) -> str:
```
johnchildren, line 233: Possibly this should just be called `uid` instead?

pytket/pytket/wasm/wasm.py
```diff
@@ -236,8 +256,53 @@ def check_function(
         :type number_of_returns: int
         :return: true if the signature and the name of the function is correct"""
 
-        return not self._check_file or (
+        return not self.checked or (
```
johnchildren, line 259: This behavior does seem a bit weird to me as well, but I'm not sure about the intention behind checking versus not checking, presumably if you are calling `check_function` you do actually want to check?

pytket/pytket/wasm/wasm.py
```diff
@@ -42,13 +48,15 @@ class WasmFileHandler:
...
+        Construct a wasm module handler
 
-        :param filepath: Path to the wasm file
-        :type filepath: str
-        :param check_file: If ``True`` checks file for compatibility with wasm
+        :param wasm_module: A WASM module in binary format.
+        :type wasm_module: bytes
+        :param check: If ``True`` checks file for compatibility with wasm
           standards. If ``False`` checks are skipped.
         :type check_file: bool
```
cqc-melf, line 61:
```suggestion
        :type check: bool

File: pytket/pytket/wasm/wasm.py [hunk 10]
```diff
@@ -64,169 +72,181 @@ def __init__(self, filepath: str, check_file: bool = True, int_size: int = 32):
...
+        if check:
+            self.check()
+
+    def check(self) -> None:
+        """Collect functions from the module that can be used with pytket.
+
+        Populates the internal list of supported and unsupported functions
+        and marks the module as checked so that subsequent checking is not
+        required.
+        """
```
cqc-melf, line 96: Do we maybe want to make sure this has not run before?
Maybe not run it if `self.checked` is already true

pytket/pytket/wasm/wasm.py
cqc-melf: I think this is the structure we need to be compatible with the way `add_wasm` is using the check function. When the file handler is constructed without enabling the checks, `check_function` should always return true.

I see that this is not really intuitive, previously the `self._check_file` stored the parameter given by the user, and now the `self.check` stores if the check has been run, so even if the user does not want to check the file in some situations it does get checked.

pytket/pytket/wasm/wasm.py
cqc-melf: I am not fully sure that it is good to run the checks here (and on other places) secretly ignoring that the user first requested the file not to be checked. Should we at least give a warning that this is done?

cqc-melf:
pytket-qir and pytket-quantinuum are still expecting a `WasmFileHandler` so we would still need to update them to get the benefits of this changes?

johnchildren:
> pytket-qir and pytket-quantinuum are still expecting a `WasmFileHandler` so we would still need to update them to get the benefits of this changes?

I think the only advantage would be for type checking, the interface would be the same. I'm not sure how publicly you would want to make the change, but it might be good for me to make PRs against both those repos as well to check that they can accept the new class as well?

cqc-melf:
> edit: I'm not sure how easy it is to get CI running with branches of tket though.

You can't test against branches, the usual step would be to put out a pytket release candidate and use that for making a PR.

We usually only do the release candidates from `main` (so after this is merged) and then make a new PR into tket if that is required.

cqc-melf:
> > pytket-qir and pytket-quantinuum are still expecting a `WasmFileHandler` so we would still need to update them to get the benefits of this changes?
> 
> I think the only advantage would be for type checking, the interface would be the same. I'm not sure how publicly you would want to make the change, but it might be good for me to make PRs against both those repos as well to check that they can accept the new class as well?

Yes, would be good for type checking, and maybe adding a testcase just to be sure.

pytket/pytket/wasm/wasm.py
cqc-melf: Yes :+1:

pytket/pytket/wasm/wasm.py
cqc-melf: Yes would be good

pytket/pytket/wasm/wasm.py
cqc-melf: Yes, see the comment below. I think the problematic point is that this starts the type checking of all wasm functions that are added later on and it ignores that the user initially requested that they should not be checked.
Maybe it would be good to give a warning?

pytket/pytket/wasm/wasm.py
johnchildren: I'm not really sure I understand the use-case behind not checking the module, is the intention to save compute, to avoid erroneous checks or to allow submission of broken code?

cqc-melf: The idea is to avoid erroneous checks and allow the submission of broken code.

pytket/pytket/wasm/wasm.py
cqc-melf: That sounds like a good idea to me, but I am not sure how this solved the problem that the value of `filehandler.checked` might be different then the user expected it to be?

cqc-melf (approved):
Thank you for updating! Looks good now :)

PR #1512, opened 2024-08-02
remove types setuptools and release 1.31

cqc-melf:
# Description

Please summarise the changes.

# Related issues

# Checklist


cqc-melf:
Tries to solve https://github.com/CQCL/tket/issues/1513

pytket/docs/changelog.rst
```diff
@@ -1,9 +1,10 @@
 Changelog
 =========
 
-(unreleased)
-------------
+1.31.0 (August 2024)
+--------------------
 
+* remove types-pkg_resources dependency
```
CalMacCQ, line 7: GIven that the removal of this package from pypi renders pytket uninstallable I think we should announce this more prominently in the docs. What do you think?

CalMacCQ: I can do this in a separate PR

sjdilkes:
Looks like we might be in luck here - have you also tried installing in a fresh local environment where we can be sure there's no lingering installation of `types-pkg_resources`? I assume the CI is set up to do this, but a sanity check would be helful.

pytket/docs/changelog.rst
sjdilkes, line 7:
```suggestion
* Remove unnecessary types-pkg_resources dependency due to package  being yanked from pypi
```

jake-arkinstall:
A couple of thoughts of potential relevance:
* The version and release strings in pytket/docs/conf.py (I read the version from there when providing nix with the version information) still reference "1.30" - if this PR is for bumping the version should those be changed as part of it?
* poetry.lock still mentions pkg-resources

cqc-melf:
> Looks like we might be in luck here - have you also tried installing in a fresh local environment where we can be sure there's no lingering installation of `types-pkg_resources`? I assume the CI is set up to do this, but a sanity check would be helful.

Yes I did that:
- [ ] install pytket
- [ ] run python tests
- [ ] run mypy checks
- [ ] make sure `types-pkg-resources` is still not installed

pytket/docs/changelog.rst
cqc-melf: Yes we should do that. Feel free to make a second PR for that.

cqc-melf:
The doc test run currently in an older version and I don't see and easy way to solve that at the moment. I would suggest removing the doc checks for now and try to add them back in later.

The main problem is:
Docs need poetry
Poetry does not like setup.py
(I think the solution should be building a wheel first and loading that device specific into poetry is a to complicated change for a release PR, and not fully sure that would work)

cqc-melf:
@sjdilkes @CalMacCQ should be ready for review now. If there are any questions or anything we should discuss, happy to have a call.

CalMacCQ (approved):
Looks good to me.

The TKET webiste will pick up the v1.31 from pypi once the release is put out and will build the docs.

After the 1.31 release is out we can re-enable the docs test

sjdilkes (approved):

PR #1591, opened 2024-09-20
Update to catch2 3.7.1 and pybind11 2.13.6

cqc-alec:
With the pybind11 update, a new stub function is added in many places in the `.pyi` files generated by pybind11-stubgen. To avoid mypy errors I modified the stub-generation script to add a `# type: ignore` to this function (as I have no idea what the return type should be).

### DIFF
--- a/.github/workflows/build-without-conan.yml
+++ b/.github/workflows/build-without-conan.yml
@@ -74,9 +74,9 @@ jobs:
     - name: Install catch2
       run: |
         cd ${TMP_DIR}
-        wget https://github.com/catchorg/Catch2/archive/refs/tags/v3.7.0.tar.gz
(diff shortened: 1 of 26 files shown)


cqc-melf (approved):
Happy to approve this, but maybe this is a bug that all this functions are added and we should better skip this version and wait for the functions to be removed again? Or is there a need to update?

cqc-alec:
> Happy to approve this, but maybe this is a bug that all this functions are added and we should better skip this version and wait for the functions to be removed again? Or is there a need to update?

I think it's a "feature" rather than a bug. See: https://pybind11.readthedocs.io/en/stable/changelog.html#version-2-13-6-september-13-2024 .

PR #1607, opened 2024-10-07
Fix initialization of subcircuit in `GuidedPauliSimp` implementation

### LINKED ISSUE #1578: `GuidedPauliSimp` errors out when circuit contains CircuitBox and classical bits.
```
inner_circ_nobit = Circuit(2).H(0).H(1)
full_circ_nobit = Circuit(2)
full_circ_nobit.add_circbox(CircBox(inner_circ_nobit),inner_circ_nobit.qubits)
Out[59]: [CircBox q[0], q[1]; ]
GuidedPauliSimp().apply(full_circ_nobit)
Out[60]: True


inner_circ_withbit = Circuit(2,1).H(0).H(1)
full_circ_withbit = Circuit(2,1)
full_circ_withbit.add_circbox(CircBox(inner_circ_withbit), inner_circ_withbit.qubits + inner_circ_withbit.bits)
Out[63]: [CircBox q[0], q[1], c[0]; ]
GuidedPauliSimp().apply(full_circ_withbit)
  File "/home/andrew/.local/lib/python3.12/site-packages/IPython/core/interactiveshell.py", line 3577, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-64-fa2ebbbbd18c>", line 1, in <module>
    GuidedPauliSimp().apply(full_circ_withbit)
RuntimeError: Subcircuit boundary mismatch to hole
```

cqc-alec:
# Description

Subcircuit was being constructed assuming that all wires were quantum. Make it use classical wires as well.

# Related issues

Fixes #1578 .

# Checklist

### DIFF
--- a/tket/test/src/test_PauliGraph.cpp
+++ b/tket/test/src/test_PauliGraph.cpp
@@ -28,6 +28,7 @@
 #include "tket/PauliGraph/PauliGraph.hpp"
 #include "tket/Transformations/PauliOptimisation.hpp"
 #include "tket/Transformations/Rebase.hpp"
+#include "tket/Utils/UnitID.hpp"
 
 namespace tket {
 namespace test_PauliGraph {
@@ -692,6 +693,14 @@ SCENARIO("Test mutual diagonalisation of fully commuting sets") {
     test1.symbol_substitution(symbol_map);
     test2.symbol_substitution(symbol_map);
     REQUIRE(test_statevector_comparison(test1, test2));
+  }
+  GIVEN("A circuit with a classical bit") {
+    // https://github.com/CQCL/tket/issues/1578
+    Circuit c0(0, 1);
+    CircBox cbox(c0);
+    Circuit c(0, 1);
+    c.add_box<CircBox, Bit>(cbox, {Bit("c", 0)});
+    REQUIRE_NOTHROW(Transforms::special_UCC_synthesis().apply(c));
   }
   GIVEN(
       "Clifford merges requires removing from start line without segfault "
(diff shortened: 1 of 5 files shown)


tket/test/src/test_PauliGraph.cpp
```diff
@@ -693,6 +694,14 @@ SCENARIO("Test mutual diagonalisation of fully commuting sets") {
     test2.symbol_substitution(symbol_map);
     REQUIRE(test_statevector_comparison(test1, test2));
   }
+  GIVEN("A circuit with a classical bit") {
+    // https://github.com/CQCL/tket/issues/1578
+    Circuit c0(0, 1);
+    CircBox cbox(c0);
+    Circuit c(0, 1);
+    c.add_box<CircBox, Bit>(cbox, {Bit("c", 0)});
+    REQUIRE_NOTHROW(Transforms::special_UCC_synthesis().apply(c));
```
sjdilkes, line 703: I didn't know `GuidedPauliSimp` could handle classically controlled CircBox. Is this fix just to make sure that the method doesn't complain that there are classically controlled CircBox that it won't optimise anyway (in which case I think this test is fine), or is it also decomposing them (in which case maybe we need the CircBox to include something that gets optimised)?

sjdilkes (changes requested):

tket/test/src/test_PauliGraph.cpp
cqc-alec: In this case there's no classical control, it's just a classical wire being passed through the `CircBox`. Previously this caused an error because the `Subcircuit` constructor was assuming all wires were quantum.

sjdilkes (approved):

PR #1613, opened 2024-10-09
fix: update wasm functions to accept `WasmModuleHandler`

qartik:
# Description

The `add_wasm` and `add_wasm_to_reg` functions should now accept `WasmModuleHandler` which is a super class of `WasmFileHandler`.

# Related issues

Follow up to #1503

# Checklist

### DIFF
--- a/pytket/pytket/circuit/__init__.py
+++ b/pytket/pytket/circuit/__init__.py
@@ -63,7 +63,7 @@ from .logic_exp import (
 def overload_add_wasm(
     self: Circuit,
     funcname: str,
-    filehandler: wasm.WasmFileHandler,
+    filehandler: wasm.WasmModuleHandler,
     list_i: Sequence[int],
     list_o: Sequence[int],
     args: Union[Sequence[int], Sequence[Bit]],
@@ -72,7 +72,7 @@ def overload_add_wasm(
 ) -> Circuit:
     """Add a classical function call from a wasm file to the circuit.
     \n\n:param funcname: name of the function that is called
-    \n:param filehandler: wasm file handler to identify the wasm file
+    \n:param filehandler: wasm file or module handler to identify the wasm module
     \n:param list_i: list of the number of bits in the input variables
     \n:param list_o: list of the number of bits in the output variables
     \n:param args: vector of circuit bits the wasm op should be added to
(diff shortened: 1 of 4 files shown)


qartik:
Hi @cqc-melf, unsure if the errors reported on the CI are because of the changes in the PR. Could you help take a look?

cqc-alec:
The `.pyi` file needs to be generated using the `generate_stubs` script (I can do that and push the change...).

cqc-alec:
I don't know why the coverage check is failing. I think there is some random variation in it, but I haven't seen it fail for a long time. Will try rerunning jobs just in case it's a random outlier.

cqc-alec:
OK, I don't know why the coverage has gone down, maybe the previous coverage metric was abnormally high. Happy to ignore that failure, if @cqc-melf is otherwise happy with the changes.

qartik:
> I don't know why the coverage has gone down

it could possibly be because all the existing tests are still passing `WasmFileHandler`, while the function accepts `WasmModuleHandler`, let me add a test for the latter as well.

cqc-melf (approved):
From my perspective this looks good now :)

cqc-melf:
@cqc-alec Could you merge this maybe?

pytket/pytket/circuit/__init__.py
```diff
@@ -72,7 +72,7 @@ def overload_add_wasm(
 ) -> Circuit:
     """Add a classical function call from a wasm file to the circuit.
     \n\n:param funcname: name of the function that is called
-    \n:param filehandler: wasm file handler to identify the wasm file
+    \n:param filehandler: wasm file or module handler to identify the wasm module
```
cqc-alec, line 75: Shouldn't we change the type annotation to a `Union` if both types are accepted (not sure why mypy didn't complain)?

pytket/pytket/circuit/__init__.py
```diff
@@ -113,15 +113,15 @@ def overload_add_wasm(
...
+    filehandler: wasm.WasmModuleHandler,
     list_i: Sequence[BitRegister],
     list_o: Sequence[BitRegister],
     args_wasm: Optional[Sequence[int]] = None,
     **kwargs: Any,
 ) -> Circuit:
     """Add a classical function call from a wasm file to the circuit.
     \n\n:param funcname: name of the function that is called
-    \n:param filehandler: wasm file handler to identify the wasm file
+    \n:param filehandler: wasm file or module handler to identify the wasm module
```
cqc-alec, line 124: As above.

pytket/pytket/circuit/__init__.py
qartik: Like I mentioned in the OP, `WasmFileHandler` is a subclass of `WasmModuleHandler`. If we are accepting the superclass, any of its subclasses is acceptable.

qartik: But if you prefer the Union for better documentation from types, I am happy to make that change, let me know.

cqc-alec (approved):

PR #1634, opened 2024-10-23
Drop support for MacOS 12

cqc-alec:


.github/workflows/release.yml
```diff
@@ -65,8 +65,6 @@ jobs:
   build_macos_x86_wheels:
     name: Build macos x86 wheels
     runs-on: macos-13
-    env:
-        MACOSX_DEPLOYMENT_TARGET: '12.0'
```
cqc-alec, line 69: I'll put this back with 13.0.

cqc-melf (approved):

PR #1636, opened 2024-10-23
Updated nixpkgs. Added custom mypy build.

### LINKED ISSUE #1635: nix builds on CI failing
Seems they aren't using the latest version of mypy.

jake-arkinstall:
# Description

Updated nixpkgs, which was quite overdue (perhaps I should add an auto updater bot?).

Mypy 1.13.0 isn't in nixpkgs yet, so I've added its derivation directly into the nix-support directory (and pushed it to the tket cachix from my machine in advance, because it's a fairly hefty build).

# Related issues

Resolves #1635 

# Checklist

- [ ] I have commented hard-to-understand parts of my code (it's nix, it's all hard to understand!)
- [ ] I have made corresponding changes to the public API documentation. (N/A)
- [ ] I have added tests that prove my fix is effective or that my feature works. (N/A)
- [ ] I have updated the changelog with any user-facing changes (N/A).


nix-support/mypy.nix
```diff
@@ -0,0 +1,139 @@
+{
```
johnchildren, line 1: rather than writing a new definition, can you just use overrideAttrs with a new `src` and `version` attr?

johnchildren: I guess it might be tricky due to the way the python packaging infra works?

jake-arkinstall: I wrote it that way expecting it to be awkward to update given all that's going on in the original, but it turned out to be easier than expected.

I've changed it to an overrideAttrs approach now. The annoying part is that I need to override the name manually.

cqc-alec (approved):

PR #1646, opened 2024-10-31
Add utility method to check register alignment in `ClExprOp`

### LINKED ISSUE #1644: Add utility function to check that all `ClExprOp`s are register-aligned
This is a requirement for conversion to QASM, QIR and PHIR. Rather than implement the check in three different places, provide a utility function in pytket for others to use.

cqc-alec:
Closes #1644 .

Includes a fix to check the suitability of a pytket circuit for conversion to QASM _after_ calling `DecomposeBoxes()`.

### DIFF
--- a/pytket/docs/changelog.rst
+++ b/pytket/docs/changelog.rst
@@ -4,6 +4,11 @@ Changelog
 Unreleased
 ----------
 
+Features:
+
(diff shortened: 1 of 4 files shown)


yao-cqc (approved):
LGTM

PR #1647, opened 2024-11-01
Add custom_deserialisation argument to BasePass.from_dict

sjdilkes:
# Description

Replaces the use of `nlohmann::json` implicit construction of c++ objects from json using `json` and `from_json` with a new method `deserialise` that directly produces the `PassPtr` object. This allows a new argument `custom_deserialisation` to be supplied, which is a map between `std::string` corrresponding to `CustomPass` label and a Circuit->Circuit function that can be used to generate a `customPass`.

# Related issues

# Checklist


pytket/binders/passes.cpp
```diff
@@ -264,24 +264,26 @@ PYBIND11_MODULE(passes, m) {
...
+          [](const BasePass &base_pass) { return serialise(base_pass); },
           ":return: A JSON serializable dictionary representation of the Pass.")
       .def_static(
           "from_dict",
-          [](const py::dict &base_pass_dict) {
-            return json(base_pass_dict).get<PassPtr>();
+          [](const py::dict &base_pass_dict,
+             const std::map<
+                 std::string, std::function<Circuit(const Circuit &)>>
+                 &custom_deserialisation) {
```
cqc-alec, line 274: This argument should be optional, then we don't break the API.

pytket/pytket/_tket/passes.pyi
```diff
@@ -54,7 +54,7 @@ class BasePass:
         :param after_apply: Invoked after a pass is applied. The CompilationUnit and a summary of the pass configuration are passed into the callback.
         :return: True if pass modified the circuit, else False
         """
-    def to_dict(self) -> dict:
+    def to_dict(self) -> json:
```
cqc-alec, line 57: Why can't this still return a `dict`?

tket/include/tket/Predicates/CompilerPass.hpp
```diff
@@ -301,6 +299,14 @@ class RepeatUntilSatisfiedPass : public BasePass {
   PredicatePtr pred_;
 };
 
+nlohmann::json serialise(const BasePass& bp);
+nlohmann::json serialise(const PassPtr& pp);
+nlohmann::json serialise(const std::vector<PassPtr>& pp);
+
+PassPtr deserialise(
+    const nlohmann::json& j,
+    const std::map<std::string, std::function<Circuit(const Circuit&)>>&
```
cqc-alec, line 308: If we called these `to_json` and `from_json` as before, can we avoid some of the changes in `CompilerPass.cpp` (not sure how the magic in `nlohmann::json` works).

pytket/tests/passes_serialisation_test.py
cqc-alec, line 1: Need to add new tests for cusom-pass round serialization.

tket/include/tket/Predicates/CompilerPass.hpp
cqc-alec: (And also change the signature to match the signatures in `JSON_DECL` but with the addition of the one optional argument.)

yao-cqc:
What about typing `custom_deserialisation` `std::map<std::string, std::function<PassPtr(const ConfigType &)>>`?
Here `ConfigType` can be something like `nlohmann::json` (I don't know what is the best), and it's used for parametrise the pass.

Then we can do something like this:
```python
config ={
            "optimisation_level":optimisation_level,
            "seed":seed
        }
def get_routing_pass(config:Dict) -> CustomPass:
    def transformation(circ: Circuit) -> Circuit:
        # code for transforming the circuit
        # that uses the config
        return Circuit()
    return CustomPass(
        transformation,
        label="MyPass",
        config=config
        )

# serialisation
serialised_pass = get_routing_pass(config).to_dict()
# deserialisation
deserialised_pass = BasePass.from_dict(
    serialised_pass,
    custom_deserialisation={
        "MyPass": get_routing_pass
    }
)   
```
This does require adding a config field to `CustomPass`.

pytket/pytket/_tket/passes.pyi
sjdilkes: This is auto generated based on the `serialise` method now returning `json` in the c++

tket/include/tket/Predicates/CompilerPass.hpp
sjdilkes: My look through the nlohmann docs/code suggested that this wasn't possible, with the nlohmann magic relying one using well defined to_json/from_json with two arguments - the json and the object. So I don't think this is possible

pytket/tests/passes_serialisation_test.py
sjdilkes: done

pytket/binders/passes.cpp
sjdilkes: done

sjdilkes:
> What about typing `custom_deserialisation` `std::map<std::string, std::function<PassPtr(const ConfigType &)>>`? Here `ConfigType` can be something like `nlohmann::json` (I don't know what is the best), and it's used for parametrise the pass.
> 
> Then we can do something like this:
> 
> ```python
> config ={
>             "optimisation_level":optimisation_level,
>             "seed":seed
>         }
> def get_routing_pass(config:Dict) -> CustomPass:
>     def transformation(circ: Circuit) -> Circuit:
>         # code for transforming the circuit
>         # that uses the config
>         return Circuit()
>     return CustomPass(
>         transformation,
>         label="MyPass",
>         config=config
>         )
> 
> # serialisation
> serialised_pass = get_routing_pass(config).to_dict()
> # deserialisation
> deserialised_pass = BasePass.from_dict(
>     serialised_pass,
>     custom_deserialisation={
>         "MyPass": get_routing_pass
>     }
> )   
> ```
> 
> This does require adding a config field to `CustomPass`.

I think we would have to force the config to be std::map<std::string, int>` to make the `type` work at the c++ level. Personally I don't think not being able to pass parameters is a problem, as the user has fine control over what they pass to `custom_deserialisation`

tket/src/Predicates/CompilerPass.cpp
```diff
@@ -576,29 +588,41 @@ void from_json(const nlohmann::json& j, PassPtr& pp) {
...
       pp = RoundAngles(n, only_zeros);
+    } else if (passname == "CustomPass") {
+      std::string label = content.at("label").get<std::string>();
+      auto it = custom_deserialise.find(label);
+      if (it != custom_deserialise.end()) {
+        pp = CustomPass(it->second, label);
+      } else {
+        throw JsonError(
+            "Cannot load CustomPass without passing constructor in "
+            "custom_deserialisation map.");
```
cqc-alec, line 599:
```suggestion
            "Cannot deserialise CustomPass without passing a custom_deserialisation map "
            "with a key corresponding to the pass's label.");
```

pytket/binders/passes.cpp
```diff
@@ -265,23 +265,29 @@ PYBIND11_MODULE(passes, m) {
...
-            return json(base_pass_dict).get<PassPtr>();
+          [](const py::dict &base_pass_dict,
+
+             std::map<std::string, std::function<Circuit(const Circuit &)>>
+                 &custom_deserialisation) {
+            return deserialise(base_pass_dict, custom_deserialisation);
           },
           "Construct a new Pass instance from a JSON serializable dictionary "
-          "representation.")
+          "representation.",
```
cqc-alec, line 280: This needs more documentation of the `custom_deserialization` parameter.

cqc-alec (commented):
Thanks, just a couple of small comments.

Please also update changelog.

pytket/binders/passes.cpp
sjdilkes: added

cqc-alec (approved):

PR #1648, opened 2024-11-01
docs: link to pytket-quest extensions docs

CalMacCQ:
# Description

Adding a crosslink to the extensions index to mention pytket-quest. Should I also mention quest explicitly as a density matrix simulator?

Driveby: mention `AerDensityMatrixBackend` from pytket-qiskit which was left out until now.

### DIFF
--- a/pytket/docs/extensions.md
+++ b/pytket/docs/extensions.md
@@ -69,6 +69,8 @@ A full list of available pytket backends is shown below.
 [ProjectQBackend](inv:#*.ProjectQBackend)
 \- Backend for running statevector simulations on the ProjectQ simulator.
 
+[QuESTBackend](https://docs.quantinuum.com/tket/extensions/pytket-quest/api.html#pytket.extensions.quest.QuESTBackend) Interface to the [QUEST simulator](https://quest.qtechtheory.org/docs/).
+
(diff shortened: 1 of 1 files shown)


ferbetanzo (approved):
I'm not sure if we need to mention QuEST in the list of "Density Matrix Simulators" too, given that you use the same `QuESTBackend` for performing both kinds of simulation and by default it will do state vector simulation.

I can see that the `QulacsBackend` can also do state vector and density matrix simulations, and by default it also does state vector simulations, however we are listing it as a density matrix simulator in the docs. Do you know why this is?

CalMacCQ:
> I'm not sure if we need to mention QuEST in the list of "Density Matrix Simulators" too, given that you use the same `QuESTBackend` for performing both kinds of simulation and by default it will do state vector simulation.
> 
> I can see that the `QulacsBackend` can also do state vector and density matrix simulations, and by default it also does state vector simulations, however we are listing it as a density matrix simulator in the docs. Do you know why this is?

I take your point. I guess you have to explicitly enable DM simulation with qulacs. Maybe we should remove the separate mention for qulacs for consistency. Not sure.

I'm happy just mentioning quest as a SV simulator

PR #1657, opened 2024-11-05
Deconstruct classical conditions in terms of `ClExprOp` instead of `ClassicalExpBox`

cqc-alec:
Also change int type in expressions to `uint64_t`, since otherwise pybind11 assumes 32 bits.


pytket/tests/classical_test.py
```diff
@@ -1115,178 +1115,6 @@ def check_serialization_roundtrip(circ: Circuit) -> None:
     assert circ_from_dict.to_dict() == circ_dict
 
 
-def test_decomposition_known() -> None:
```
cqc-alec, line 1118: Next step after this will be to write a new decomposition pass, at which point I'll add a corresponding test to replace this one.

sjdilkes (approved):
LGTM

PR #1664, opened 2024-11-08
Add timeout parameter to GreedyPauliSimp

sjdilkes:
# Description

Use std::async to add a timeout to `greedy_pauli_optimisation`.

A segmentation fault was occurring when modifying the python expression held in `ClassicalExpBox` to relabel the expression bits to match a relabelling.

To work around this, I've added a new argument to `Circuit::flatten_registers`, `Circuit::rename_units` and `FlattenRelabelRegistersPass` to optionally relabel expressions in `ClassicalExpBox`. This is fine for the `greedy_pauli_optimisation` case as the synthesis method does not effect the held expressions and we relabel back to the original assignment after re-synthesising the circuit.

# Checklist


yao-cqc (changes requested):
Can you add a test for the timeout?

sjdilkes:
> Can you add a test for the timeout?

Do the test changes in `predicates_test.py` and `test_greedypauli` not cover it, where I add a 0 timeout and confirm that the `false` branch is returned? Happy to add more

PR #1667, opened 2024-11-08
Clifford representation incorporating global phase

### LINKED ISSUE #1659: Clifford representation incorporating global phase
Efficient Clifford representations via stabiliser theory can never capture global phase, since all of the data is based on externally observable behaviour (the stabilisers) and global phase is never observable. Nevertheless, since global phase tracking is situationally useful, we should aim for all areas of compilation to support it, even if it comes at a small computational cost (ideally toggleable if the cost is not negligible).

For example, in settings where we may wish to apply a controlled-Clifford unitary, the "global" phase in the Clifford unitary becomes local under control. We can't just handle this with stabiliser methods because the controlled-Clifford may longer be Clifford.

Solutions to this problem come from fixing a standardised statevector interpretation of each stabiliser process, from which we can track the relative phase whenever we apply gates. The "reduced affine with phases form" (reduced AP-form) described in [Chapter 5, “Picturing Quantum Software”, Kissinger & van de Wetering] would handle this nicely (relating closely to the CH-form of [“Simulation of quantum circuits by low-rank stabilizer decompositions”, Bravyi et al] and the quadratic form of ["The Cliﬀord group, stabilizer states, and linear and quadratic operations over GF(2)", Dehaene & De Moor]).

Updating this data structure will be more expensive than a standard tableau as it takes more effort to keep in this standard form, so this is something we should provide alongside faster tableau methods for when it is needed rather than in place of.

willsimmons1465:
# Description

This PR adds a new C++ class for the efficient simulation of Clifford states, including global phase. This is based on the "reduced affine with phases form" (reduced AP-form) described in [Chapter 5, “Picturing Quantum Software”, Kissinger & van de Wetering], tracking a normal form for the state which has a well-defined reference phase, from which we can track global phase as the relative difference.

This class comes with direct simulation in the basis {CZ, S, V} where these were found to give some of the simplest proofs of correctness, though this still involves many cases and the derivations are non-trivial. Proofs which track the global phase are available internally within Quantinuum, and can be provided to external colleagues on request. Other gate types are handled by reducing into this gate set, but I can add additional methods for them directly if that is preferred (e.g. Pauli gates are much simpler to perform directly than via S and V).

Due to the constraints of maintaining the normal form, this is noticeably more expensive than tableau simulation, since each gate takes O(n^2) time to apply (where n is the number of qubits in the circuit), as opposed to the O(n) time for tableau updates. Because of the increased cost, I am yet to incorporate this into any live optimisation pass. I'd appreciate some suggestions on where we think it is worth it to add cost for global phase tracking or if there is a nice way to make that optional depending on the user's requirements. Similarly, the complexity of this makes me not want to expose it to Python users yet unless others can think of a good use case where people want direct access to it.

Converters are included to go between `APState` and `Circuit` or `SymplecticTableau` (from which one can wrap it in a `ChoiMixTableau`, constrained to pure states).

I'd like to be able to use this in the new `PauliGraph` class, but `ChoiMixTableau` still provides more utility, from supporting mixed processes and the rows having semantic meaning to the rest of the class. The only way I can think of incorporating global phases would be to duplicate tracking of the Clifford regions via both a `ChoiMixTableau` (for the semantic value of the rows and simplifying rewrite strategies on the graph) and an `APState` to track global phase changes (in this case, where it is enough to just track the phase without trying to maintain the entire semantics, it is sufficient to simulate the result of applying the linear map to the |0> state, rather than simulating the entire isometry as an open process, which does reduce the cost). I'm keen to hear any other suggestions for how to handle this, and if there is a way to do it without duplication.

# Related issues

Closes #1659

# Checklist


willsimmons1465:
I may have been a little stupid in creating this. I only just realised that there is actually no need to keep it in a unique normal form, and relaxing the constraints on the form massively reduces the number of cases we need to consider (especially once I have also extended it to work with mixed states). I'm currently updating the document with the maths, and then will update the code to this simpler and more flexible version

cqc-alec:
@willsimmons1465 Should we close this?

PR #1681, opened 2024-11-18
Add convenience method `Circuit.add_clexpr_from_logicexp()`

cqc-alec:


pytket/binders/circuit/Circuit/add_op.cpp
```diff
@@ -513,6 +513,35 @@ void init_circuit_add_op(py::class_<Circuit, std::shared_ptr<Circuit>> &c) {
...
+          ">>> x_reg = c.add_c_register('x', 3)\n"
+          ">>> y_reg = c.add_c_register('y', 3)\n"
+          ">>> z_reg = c.add_c_register('z', 3)\n"
+          ">>> c.add_clexpr_from_logicexp(x_reg | y_reg, z_reg.to_list())\n"
+          ">>> [ClExpr x[0], x[1], x[2], y[0], y[1], y[2], z[0], z[1], z[2]; "
+          "]\n\n"
+          ":param exp: logical expression\n"
+          ":param output_bits: list of bits in output\n"
+          ":return: the updated circuit",
+          py::arg("exp"), py::arg("output_bits"))
```
cqc-melf, line 544:
```suggestion
          py::arg("exp"), py::arg("output"))

Comment (cqc-melf) [reply pytket/binders/circuit/Circuit/add_op.cpp:544]: And allow sequence[bits] and register?

Comment (cqc-alec) [reply pytket/binders/circuit/Circuit/add_op.cpp:544]: I think I prefer `output_bits` as it is more explicit.

Any sequence of bits should work here.

In principle the output bits do not need to form a register (though that would be the case usually), so I'm not sure it's worth adding an overload.

Comment (cqc-melf) [reply pytket/binders/circuit/Circuit/add_op.cpp:544]: I think the question is, would this work, and if not do we want it to?
```
c = Circuit()
x_reg = c.add_c_register('x', 3)
y_reg = c.add_c_register('y', 3)
z_reg = c.add_c_register('z', 3)
c.add_clexpr_from_logicexp(x_reg | y_reg, z_reg)

cqc-alec: Yes that does work; the register is coerced to a list.

cqc-alec: I'm not sure how to do that?

cqc-melf: I think we would need a second c++ function in the binders for that, but not sure if there might be a easier way?
If you think that is not worth the effort happy to approve this as it is.

PR #1683, opened 2024-11-18
Deprecate `GlobalisePhasedX`

### LINKED ISSUE #1682: Depreceate `GlobalisePhasedX`

cqc-alec:
Closes #1682 .

### DIFF
--- a/pytket/binders/passes.cpp
+++ b/pytket/binders/passes.cpp
@@ -494,8 +494,8 @@ PYBIND11_MODULE(passes, m) {
         PyErr_WarnEx(
             PyExc_DeprecationWarning,
             "The GlobalisePhasedX pass is unreliable and deprecated. It will "
-            "be "
-            "removed after pytket version 1.38.",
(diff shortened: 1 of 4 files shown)


pytket/binders/passes.cpp
```diff
@@ -499,6 +508,7 @@ PYBIND11_MODULE(passes, m) {
       "performance. If squashing is disabled, each non-global PhasedX gate "
       "will be replaced with two global NPhasedX, but any other gates will "
       "be left untouched."
+      "\n\nDEPRECATED: This pass will be removed after pytket version 1.39."
```
cqc-melf, line 511: Do we want to put a specific version, or maybe better a more specific date?

cqc-alec: Intentionally put "after" so as not to be too committal, if we don't want to actually remove it.

cqc-alec: Maybe a date would be better though.

cqc-melf (approved):

PR #1685, opened 2024-11-19
Remove `OpType::Collapse` from `gen_auto_rebase_pass` `GateSetPredicate`

sjdilkes:
`OpType::Collapse` isn't available at the python layer - there is no `OpType.Collapse` - and is not a supported operation in any `Backends` we have. Currently, I don't think it's even possible to construct a pytket `Circuit` with `Collapse` operations, having briefly looked through where it is used.

However, it's proving a blocker on `SequencePass` construction in some places. As an example, `GreedyPauliSimp` supports all the `OpType` required to compile to qiskit & quantinuum hardware, but if added to a `SequencePass` after an `AutoRebasePass`, the `SequencePass` construction fails due to `OpType::Collapse` always being included in the `GateSetPredicate`.

# Checklist

### DIFF
--- a/pytket/binders/passes.cpp
+++ b/pytket/binders/passes.cpp
@@ -540,7 +540,7 @@ PYBIND11_MODULE(passes, m) {
   m.def(
       "RebaseTket", &RebaseTket,
       "Converts all gates to CX, TK1 and Phase. "
-      "(Any Measure, Reset and Collapse operations are left untouched; "
+      "(Any Measure and Reset operations are left untouched; "
       "Conditional gates are also allowed.)");
   m.def(
       "RemoveRedundancies", &RemoveRedundancies,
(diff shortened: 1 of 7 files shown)


cqc-alec (approved):

cqc-alec (changes requested):
Sorry, please update the documentation of `AutoRebase` in the binders as well.

cqc-alec:
> Sorry, please update the documentation of `AutoRebase` in the binders as well.

And `auto_rebase_pass()` (even though it's deprecated).

sjdilkes:
> > Sorry, please update the documentation of `AutoRebase` in the binders as well.
> 
> And `auto_rebase_pass()` (even though it's deprecated).

done both!

cqc-alec (approved):

Issue #1675, opened 2024-11-12
Add method to passes returning operations in gateset predicate


daniel-mills-cqc:
At present extracting the gate set a compiler pass can handle is quite manual. This list is only revealed by the exception raised when the gate set predicate is not met. It would be great to be able to extract the operations supported. The main use case here is to construct from that list a rebase pass which can be applied before the pass itself.
Relates to discussion here -> https://cambridgequantum.slack.com/archives/CTAQUR7M0/p1730193783655379

PR #1689, opened 2024-11-21
Add `BasePass.get_pre_conditions` and `BasePass.get_post_conditions`

sjdilkes:
Adds two new methods to `BasePass` in passes.cpp to expose pre conditions and post conditions for a pass to the python layer as `List[Predicate]`.

# Related issues

https://github.com/CQCL/tket/issues/1675

# Checklist


pytket/binders/passes.cpp
```diff
@@ -268,6 +268,32 @@ PYBIND11_MODULE(passes, m) {
             return py::cast(serialise(base_pass));
           },
           ":return: A JSON serializable dictionary representation of the Pass.")
+      .def(
+          "get_pre_conditions",
```
cqc-alec, line 272:
```suggestion
          "get_preconditions",
```

pytket/binders/passes.cpp
```diff
@@ -268,6 +268,32 @@ PYBIND11_MODULE(passes, m) {
...
+          [](const BasePass &base_pass) {
+            std::vector<PredicatePtr> pre_conditions;
+            for (const std::pair<
+                     const std::type_index, std::shared_ptr<tket::Predicate>>
+                     &p : base_pass.get_conditions().first) {
+              pre_conditions.push_back(p.second);
+            }
+            return pre_conditions;
+          },
+          "Returns the pre condition Predicates for the given pass."
```
cqc-alec, line 282:
```suggestion
          "Returns the precondition Predicates for the given pass."
```

pytket/binders/passes.cpp
```diff
@@ -268,6 +268,32 @@ PYBIND11_MODULE(passes, m) {
...
+          "\n:return: A list of Predicate")
+      .def(
+          "get_post_conditions",
```
cqc-alec, line 285:
```suggestion
          "get_postconditions",
```

pytket/binders/passes.cpp
```diff
@@ -268,6 +268,32 @@ PYBIND11_MODULE(passes, m) {
...
+          [](const BasePass &base_pass) {
+            std::vector<PredicatePtr> post_conditions;
+            for (const std::pair<
+                     const std::type_index, std::shared_ptr<tket::Predicate>> &
+                     p : base_pass.get_conditions().second.specific_postcons_) {
+              post_conditions.push_back(p.second);
+            }
+            return post_conditions;
+          },
+          "Returns the post condition Predicates for the given pass."
```
cqc-alec, line 295:
```suggestion
          "Returns the postcondition Predicates for the given pass."
```

pytket/binders/passes.cpp
```diff
@@ -268,6 +268,32 @@ PYBIND11_MODULE(passes, m) {
...
+          "\n:return: A list of Predicate")
```
cqc-alec, line 296:
```suggestion
          "\n\n:return: A list of :py:class:`Predicate`")
```

pytket/binders/passes.cpp
cqc-alec, line 1: I think it is also worth adding a function specifically to get the gate-set requirements of a pass; i.e. which gets the preconditions, checks for any `GateSetPredicate`s, and returns the set intersection of their gate sets (or `None` if there are no `GateSetPredicate`s.)

tket/conanfile.py
```diff
@@ -23,7 +23,7 @@
 
 class TketConan(ConanFile):
     name = "tket"
-    version = "1.3.48"
+    version = "1.3.49"
```
cqc-alec, line 26: This version bump shouldn't be necessary as the change only touches pytket.

pytket/binders/passes.cpp
sjdilkes: done

pytket/binders/passes.cpp
sjdilkes: done

tket/conanfile.py
sjdilkes: oops ofc - will revert

pytket/binders/passes.cpp
sjdilkes: done

pytket/binders/passes.cpp
```diff
@@ -145,6 +145,31 @@ const PassPtr &DecomposeClassicalExp() {
   return pp;
 }
 
+std::optional<OpTypeSet> get_gate_set(const BasePass &base_pass) {
+  OpTypeSet allowed_ops;
+  for (const std::pair<const std::type_index, std::shared_ptr<tket::Predicate>>
+           &p : base_pass.get_conditions().first) {
+    if (p.second->to_string().substr(0, 17) == "GateSetPredicate:") {
```
cqc-alec, line 152: I don't think this check is necessary. We can just use `dynamic_cast` and ignore it if it returns `nullptr`.

pytket/binders/passes.cpp
```diff
@@ -145,6 +145,31 @@ const PassPtr &DecomposeClassicalExp() {
...
+      std::shared_ptr<GateSetPredicate> gsp_ptr =
+          std::dynamic_pointer_cast<GateSetPredicate>(p.second);
+      if (allowed_ops.empty()) {
```
cqc-alec, line 155: This will end up giving the wrong result in the (admittedly unlikely) case that we have already encountered two `GateSetPredicate`s with empty intersection. I suggest using a `std::optional<OpTypeSet>`, where no-value means all all ops are allowed.

pytket/binders/passes.cpp
```diff
@@ -268,6 +292,37 @@ PYBIND11_MODULE(passes, m) {
...
+            }
+            return post_conditions;
+          },
+          "Returns the postcondition Predicates for the given pass."
+          "\n\n:return: A list of :py:class:`Predicate`")
+      .def(
+          "get_gate_set", &get_gate_set,
+          "Returns the intersection of all set of OpType for all "
+          "GateSetPredicate in the `BasePass` preconditions.",
+          "\n\n:return: A set of allowed OpType")
```
cqc-alec, line 325:
```suggestion
          "Returns the intersection of all set of OpType for all "
          "GateSetPredicate in the `BasePass` preconditions, or `None` if there are no gate-set predicares.",
          "\n\n:return: A set of allowed OpType, or `None`")
```

pytket/conanfile.py
```diff
@@ -38,7 +38,7 @@ def requirements(self):
         self.requires("pybind11_json/0.2.14")
         self.requires("symengine/0.13.0")
         self.requires("tkassert/0.3.4@tket/stable")
-        self.requires("tket/1.3.49@tket/stable")
+        self.requires("tket/1.3.50@tket/stable")
```
cqc-alec, line 41: This shouldn't need updating.

pytket/tests/predicates_test.py
```diff
@@ -1092,6 +1092,26 @@ def test_greedy_pauli_synth() -> None:
...
+def test_get_post_conditions() -> None:
+    gate_set = {OpType.CX, OpType.Rz, OpType.H, OpType.Reset, OpType.Measure}
+    post_cons = AutoRebase(gate_set).get_postconditions()
+    assert post_cons[0].gate_set == gate_set  # type: ignore
+
+
+def test_get_gate_set() -> None:
+    gate_set = GreedyPauliSimp().get_gate_set()
+    assert OpType.CX in gate_set  # type: ignore
+    assert OpType.Measure in gate_set  # type: ignore
```
cqc-alec, line 1111: Should be able to avoid the `type: ignore`s by first asserting that `gate_set is not None`.

tket/conanfile.py
```diff
@@ -23,7 +23,7 @@
 
 class TketConan(ConanFile):
     name = "tket"
-    version = "1.3.49"
+    version = "1.3.50"
```
cqc-alec, line 26: Shouldn't be needed.

cqc-alec (changes requested):

pytket/binders/passes.cpp
sjdilkes: done

pytket/binders/passes.cpp
sjdilkes: done (I think)

pytket/conanfile.py
sjdilkes: I've reverted - let's see if the same issue comes up

cqc-alec (approved):

PR #1690, opened 2024-11-21
Remove all references to dead vertices in ZX extraction

### LINKED ISSUE #1566: `ZXGraphlikeOptimisation()` can cause RunTime error
[Benny](https://github.com/Bennybenassius) and I found this by generating a random `Pytket` circuit.

When we run this code:
``` Python
from pytket import Circuit
from pytket.passes import ZXGraphlikeOptimisation

main_circ = Circuit(6, 0, "main_circ")

# Applying gates 
main_circ.Rx(-0.78,3)
main_circ.Rx(1.069,5)
main_circ.CX(0, 3)
main_circ.CX(3, 5)
main_circ.CX(0, 5)
main_circ.Rx(1.069,3)
main_circ.Rx(-5.958,4)
main_circ.CX(3, 4)
main_circ.CX(3, 5)
main_circ.CX(0, 3)

ZXGraphlikeOptimisation().apply(main_circ)
```

We get this runtime error:
``` bash
RuntimeError: Error during extraction from ZX diagram: diagram does not have gflow
```

We tried to minimize the test case but deleting any one of the gates would make it work. Additionally, if any of the arguments to the rotation gates are integers, it also works.

willsimmons1465:
# Description

During extraction, vertices are gradually removed from a ZX graph as we extract gates. Some additional tracking info is maintained via sets and maps of vertices. When vertices are removed, this shouldn't cause issues with looking up remaining vertices in these sets and maps.

HOWEVER, if needed it will perform an input extension which adds new vertices to the graph. Depending on how boost is feeling with memory management, it may reuse the memory locations of removed vertices for these new ones. Because the vertex descriptors are void* pointers under the hood, this clash of memory locations means we can get false positives when looking up vertices in the sets/maps.

#1566 and a related circuit found internally were shown to hit errors during extraction which at first seemed device-dependent, but was found to just be non-deterministic on any device. These were due to one of these false positive lookups which cause some vertices to be ignored when looking for a flow (hence the error message `Error during extraction from ZX diagram: diagram does not have gflow`).

Given the bug only occurs with some probability (that may be device-dependent), the tests for this fix run the circuits a bunch of times to reduce the likelihood of a false positive for CI.

# Related issues

Closes #1566 

# Checklist

### DIFF
--- a/pytket/conanfile.py
+++ b/pytket/conanfile.py
@@ -38,7 +38,7 @@ class pytketRecipe(ConanFile):
         self.requires("pybind11_json/0.2.14")
         self.requires("symengine/0.13.0")
         self.requires("tkassert/0.3.4@tket/stable")
-        self.requires("tket/1.3.48@tket/stable")
+        self.requires("tket/1.3.49@tket/stable")
         self.requires("tklog/0.3.3@tket/stable")
         self.requires("tkrng/0.3.3@tket/stable")
         self.requires("tktokenswap/0.3.9@tket/stable")
--- a/tket/conanfile.py
+++ b/tket/conanfile.py
@@ -23,7 +23,7 @@ from conan.errors import ConanInvalidConfiguration
 
 class TketConan(ConanFile):
     name = "tket"
-    version = "1.3.48"
+    version = "1.3.49"
     package_type = "library"
     license = "Apache 2"
     homepage = "https://github.com/CQCL/tket"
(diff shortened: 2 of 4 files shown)


sjdilkes (changes requested):
LGTM! Except I don't think we need to loop the tests in `test_ZXSimp` 50 times -  if you've confirmed it manually locally then I'm happy to assume it's all fixed and just run each case once.

sjdilkes (approved):

Issue #1641, opened 2024-10-29
Using `circuit.bits` as `args` to `add_wasm` with more registers causes a `Fatal Python error`.


johnchildren:
Using a test case like this in `test_classical.py` 

```python
def test_wasm_circuit_bits() -> None:
    w = wasm.WasmFileHandler("testfile.wasm")
    c = Circuit(0, 6)
    c.add_wasm("add_one", w, [1], [1], c.bits)
    assert c.depth() == 1
```

Results in an error during the tests of:

```
classical_test.py ................[2024-10-29 15:13:09] [tket] [critical] Assertion 'sum_of_i32 == n_' (/build/tket-sources/src/Ops/ClassicalOps.cpp : WASMOp : 266) failed>
Fatal Python error: Aborted

Current thread 0x00007ffff7f8f740 (most recent call first):
  File "/nix/store/sldmq9pa6aw0plb54q6lc9maydx6a35k-python3.12-pytket-1.34.0/lib/python3.12/site-packages/pytket/circuit/__init__.py", line 103 in overload_add_wasm
  File "/build/test_root/tests/classical_test.py", line 363 in test_wasm_circuit_bits
```

Not that reducing the number of classical registers in the circuit or using a different `args` avoids this issue.

johnchildren:
You can also trigger this bug with:

```python
def test_wasm_circuit_bits() -> None:
    w = wasm.WasmFileHandler("testfile.wasm")
    c = Circuit(0, 1)
    c.add_wasm("add_one", w, [1], [1], [0])
    assert c.depth() == 1

```

cqc-alec:
I noticed recently that the unit ID we use for the wasm state (which has the special name `_w`) has type `Bit`, which seems wrong. Could this be related @cqc-melf ?

cqc-alec:
> [@cqc-alec](https://github.com/cqc-alec) I have taken a look on this, it should have the type WasmState, see
> 
> [tket/tket/include/tket/Utils/UnitID.hpp](https://github.com/CQCL/tket/blob/a9add4b7986a255cc3fdf90dc17aaa3d85379c7b/tket/include/tket/Utils/UnitID.hpp#L229)
> 
> Line 229 in [a9add4b](/CQCL/tket/commit/a9add4b7986a255cc3fdf90dc17aaa3d85379c7b)
>  class WasmState : public UnitID { 
> 
> Can you give some more details where you have seen this? I was not able to construct an example where this did not work.

[...]

Issue #1546, opened 2024-08-20
Error when concatenating WASM circuits


yao-cqc:
```python
from pytket import Circuit
from pytket.wasm import WasmFileHandler

wasfile = WasmFileHandler("add1.wasm")
c = Circuit(1)
a = c.add_c_register("a", 8)
c.add_wasm_to_reg("add_one", wasfile, [a], [a])
d = Circuit()
d.append(c)
d
```
shows empty list: []

Adding ops to d can even lead to crash

```python
wasfile = WasmFileHandler("add1.wasm")
c = Circuit(1)
a = c.add_c_register("a", 8)
c.add_wasm_to_reg("add_one", wasfile, [a], [a])

d = Circuit()
for bit in c.bits:
    d.add_bit(bit)
    d.add_c_setbits([False], [bit])
d.append(c)
d
```

Issue #1504, opened 2024-07-30
Add function to get wasm uid from circuit


cqc-melf:
It would be good to have a function to get a list of all the wasm uid used in a circuit.

PR #1691, opened 2024-11-21
fix wasm issues

cqc-melf:
# Description

Fixes:
https://github.com/CQCL/tket/issues/1641
https://github.com/CQCL/tket/issues/1504
https://github.com/CQCL/tket/issues/1546

# Checklist


tket/test/src/test_wasm.cpp
cqc-alec, line 1: All the appending tests here involve appending to an empty circuit. Could we add some tests appending to a circuit that already contains WASM wires?

pytket/tests/classical_test.py
```diff
@@ -481,6 +482,109 @@ def test_add_wasm_to_reg() -> None:
...
+
+    c.add_wasm("add_one", w, [1], [1], [Bit(0), Bit(1)])
+
+    assert c.depth() == 1
+
+    cmd = c.get_commands()[0]
+
+    assert isinstance(cmd.args[0], unit_id.Bit)
+    assert isinstance(cmd.args[1], unit_id.Bit)
+    assert isinstance(cmd.args[2], unit_id.WasmState)
```
cqc-alec, line 502: Can you also remove the workaround in `pytket/pytket/circuit/decompose_classical.py` line 513? Shouldn't be needed now.

tket/src/Circuit/macro_manipulation.cpp
```diff
@@ -139,6 +139,9 @@ void Circuit::append_with_map(const Circuit& c2, const unit_map_t& qm) {
   Circuit copy = c2;
   copy.rename_units(qm);
 
+  copy.add_wasm_register(_number_of_wasm_wires);
+  add_wasm_register(copy._number_of_wasm_wires);
```
cqc-alec, line 143: What happens if the appended-to circuit already contains WASM wires? Will this throw an error?

pytket/binders/circuit/Circuit/main.cpp
```diff
@@ -413,6 +413,9 @@ void def_circuit(py::class_<Circuit, std::shared_ptr<Circuit>> &pyCircuit) {
       .def_property_readonly(
           "n_gates", &Circuit::n_gates,
           ":return: the number of gates in the Circuit")
+      .def_property_readonly(
+          "wasm_uid", &Circuit::get_wasm_file_uid,
+          ":return: the wasm uid of the first wasmop found in the circuit")
```
cqc-alec, line 418: Can a circuit contain more than one WASM UID? If so it would be preferable to return the full set here. If not, we can be explicit about this in the doc.

pytket/tests/classical_test.py
```diff
@@ -481,6 +483,164 @@ def test_add_wasm_to_reg() -> None:
...
+
+    with open("testfile.wasm", "rb") as f:
+        bytecode = f.read()
+    assert w.bytecode() == bytecode
+
+    c = Circuit(0, 6)
+
+    assert c.depth() == 0
+    with pytest.raises(ValueError):
+        c.wasm_uid
```
cqc-alec, line 589: I think it would be a bit nicer if this returned `None` when there is no WASM wire, rather than throw an exception.

cqc-alec (commented):
Just one suggestion, making the wasm ID an optional string (`None` if no WASM ID).

cqc-melf:
@cqc-alec Is it expected that some changes to the stubs do not make the check fail? As you can see in the last commit, there have been some changes, that did not show up on the CI?

pytket/binders/circuit/Circuit/main.cpp
```diff
@@ -414,7 +414,15 @@ void def_circuit(py::class_<Circuit, std::shared_ptr<Circuit>> &pyCircuit) {
           "n_gates", &Circuit::n_gates,
           ":return: the number of gates in the Circuit")
       .def_property_readonly(
-          "wasm_uid", &Circuit::get_wasm_file_uid,
+          "wasm_uid",
+          [](const Circuit &circ) {
+            std::optional<std::string> result;
+            try {
+              result = circ.get_wasm_file_uid();
+            } catch (const std::exception &) {
```
cqc-alec, line 422: Can we propagate the change to the C++, i.e. male `get_wasm_file_uid()` return a `std::optional<std::string>`? I think that's better than a catch-all exception handler, and could actually make the C++ code nicer too.

pytket/tests/zx_diagram_test.py
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from math import isclose, pow
+from math import isclose, pow  # noqa: A004
```
cqc-alec, line 15: I suggest just removing the `pow` import and using the built-in one; as used in this file there is no difference.

cqc-alec:
> @cqc-alec Is it expected that some changes to the stubs do not make the check fail? As you can see in the last commit, there have been some changes, that did not show up on the CI?

Not sure what you mean here. In the last commit both the binder code and the stubs changed, so I would not expect the check fail?

pytket/pytket/_tket/circuit.pyi
```diff
@@ -2572,9 +2572,9 @@ class Circuit:
         A list of all qubit ids in the circuit
         """
     @property
-    def wasm_uid(self) -> str:
+    def wasm_uid(self) -> str | None:
         """
-        :return: the wasm uid of the first wasmop found in the circuit
+        :return: the unique wasm uid of the circuit
```
cqc-melf, line 2577: @cqc-alec this line was already changed in the other commit, but the stubs check has not shown this difference.

cqc-alec: Certainly it should fail if they don't match, but I'm struggling to see where there is/was a mismatch.

cqc-melf: The comment was changed in [14b5146](https://github.com/CQCL/tket/pull/1691/commits/14b51462f6328f618036f2e1f5297a48c3faa69a), but the CI check has not show this difference.

pytket/tests/zx_diagram_test.py
cqc-melf: done in the merge commit

pytket/binders/circuit/Circuit/main.cpp
cqc-melf: done in [00c3b30](https://github.com/CQCL/tket/pull/1691/commits/00c3b307346d3d7b91dc36d0fef2158835227bbb)

pytket/binders/circuit/Circuit/main.cpp
```diff
@@ -415,14 +415,7 @@ void def_circuit(py::class_<Circuit, std::shared_ptr<Circuit>> &pyCircuit) {
...
           "wasm_uid",
-          [](const Circuit &circ) {
-            std::optional<std::string> result;
-            try {
-              result = circ.get_wasm_file_uid();
-            } catch (const std::exception &) {
-            }
-            return result;
-          },
+          [](const Circuit &circ) { return circ.get_wasm_file_uid(); },
```
cqc-alec, line 418:
```suggestion
          &Circuit.get_wasm_file_uid,
```

pytket/binders/circuit/Circuit/main.cpp
```diff
@@ -415,14 +415,7 @@ void def_circuit(py::class_<Circuit, std::shared_ptr<Circuit>> &pyCircuit) {
...
           ":return: the unique wasm uid of the circuit")
```
cqc-alec, line 419:
```suggestion
          ":return: the unique WASM UID of the circuit, or `None` if the circuit has none")
```

tket/src/Circuit/macro_manipulation.cpp
```diff
@@ -143,7 +143,8 @@ void Circuit::append_with_map(const Circuit& c2, const unit_map_t& qm) {
     if (copy.get_wasm_file_uid() != get_wasm_file_uid()) {
       throw Unsupported(
           "Cannot append circuits with different wasm uids: " +
-          get_wasm_file_uid() + " and " + copy.get_wasm_file_uid());
+          get_wasm_file_uid().emplace(" - ") + " and " +
+          copy.get_wasm_file_uid().emplace(" - "));
```
cqc-alec, line 147:
```suggestion
          get_wasm_file_uid().emplace("(none)") + " and " +
          copy.get_wasm_file_uid().emplace("(none)"));
```

pytket/docs/changelog.rst
```diff
@@ -1,13 +1,22 @@
 Changelog
 =========
 
-Unreleased
-----------
+1.36.0 (unreleased)
```
cqc-alec, line 4: Prefer to keep this as "Unreleased" only. We don't know if it will be 1.36.0 or 1.35.1 ...

cqc-melf: done in [37f0f1b](https://github.com/CQCL/tket/pull/1691/commits/37f0f1bef4fd7cda549be65d849e3b5af21282b4)

pytket/binders/circuit/Circuit/main.cpp
cqc-melf: I was not able to get this to work, I already tried a few things before I saw the suggestion, and I think `[](const Circuit &circ) { return circ.get_wasm_file_uid(); },` is the best way to do it.

cqc-alec (approved):

PR #1697, opened 2024-11-27
Only update index map when needed

### LINKED ISSUE #1696: Improve performance of `CliffordReductionPass`
This seems to be a bottleneck in `FullPeepholeOptimise`. According to the attached profiling graph, at least on x86_64 the bottleneck is the call to `Circuit::index_map()` in `CliffordReductionPass::search_back_for_match()`.
Image

cqc-alec:
Closes #1696 .

Wheels built here: https://github.com/CQCL/tket/actions/runs/12053562141 .


johnchildren (approved):
LGTM

PR #1698, opened 2024-11-27
perf: Avoid rehashing in index_map

johnchildren:
# Description
The IndexMap data structure is an unordered_map which means that if the hash table is too small it can result in having to rehash all members of the table as seen in https://github.com/CQCL/tket/issues/1696 .

This commit attempts to reserve an exact size of hash table up front such that rehashing is unlikely as an additional optimization for large circuits.

# Related issues

https://github.com/CQCL/tket/issues/1696 <- original issue
https://github.com/CQCL/tket/pull/1697 <- reduces the number of calls to index_map

# Checklist


johnchildren:
Closing in favor of: https://github.com/CQCL/tket/pull/1699

PR #1707, opened 2024-12-02
Check liveness of bits when computing slice iterator

### LINKED ISSUE #1673: Error returning depth from circuit with only Classical ops
The following code snippet:
```
from pytket.circuit import (
    Circuit,
    MultiBitOp,
    SetBitsOp,
)
from pytket.pauli import Pauli



set_bits = SetBitsOp([True, True])
multi_bit = MultiBitOp(set_bits, 2)
eq_pred_values = [True, False, False, True]
and_values = [bool(i) for i in [0, 0, 0, 1]]

circ = Circuit(4, 4, name="test")
circ.add_gate(multi_bit, [0, 1, 2, 3])

circ.add_c_predicate(eq_pred_values, [0, 1], 2, "EQ")
circ.add_c_modifier(and_values, [1], 2)

circ.measure_all()


print(circ.depth())
```

gives:

```
[tket] [critical] Assertion '!"RAW hazard created in slicing"' (../../src/Circuit/macro_circ_info.cpp : get_next_b_frontier : 400) failed.  Aborting.
```

cqc-alec:
Fixes #1673 .

Most of this PR is just refactoring: moving `SliceIterator` and related functions into its own source file (and making it no longer a member of the `Circuit` class); and combining two versions of `Circuit::next_cut()` with different signatures into one. It was this last refactor that uncovered the source of the bug. In only one of these versions was a certain liveness check being done.

So the whole fix is actually in https://github.com/CQCL/tket/pull/1707/commits/98862a3cba16a370b21fc8d1d7f2d5767e452c68 -- the rest is pure refactoring.


sjdilkes (approved):
LGTM

PR #1716, opened 2024-12-10
Add check for graph validity before commuting conditional gates in squashing passes

### LINKED ISSUE #1324: SIGABRT Error with `SquashRzPhasedX` Pass on Classically Controlled Circuits
### Overview

An issue analogous to the one reported in #1305 and #1306 occurs when applying the `SquashRzPhasedX` pass to circuits containing classically conditioned gates. 

### Example and reproduction steps

Consider the following Python code:

```python
import pytket
from pytket.passes import GlobalisePhasedX

circuit_tket = pytket.Circuit(3, 1)
circuit_tket.CZ(0, 1)
circuit_tket.Rz(1, 1, condition=pytket.Bit(0))
circuit_tket.Measure(2, 0)
circuit_tket.CX(2, 0)
circuit_tket.CZ(0, 1)


print(SquashRzPhasedX().apply(circuit_tket))
print(circuit_tket.get_commands())
```
#### Observed

Executing the above code results in a critical assertion failure and the process terminates unexpectedly with a SIGABRT signal:

```
True
[tket] [critical] Assertion '!"slice is empty"' (../../src/Circuit/macro_circ_info.cpp : operator++ : 935) failed.  Aborting.

Process finished with exit code 134 (interrupted by signal 6:SIGABRT)
```

### LINKED ISSUE #1357: Invalid SliceIterator on Circuit with mid measurement
After investigating #1324, it seems the bug is not in `SquashRzPhasedX`, but in the command iteration utility. In fact, consider the following circuit:

```c++
  Circuit c(2, 1);
  c.add_op<unsigned>(OpType::CZ, {0, 1});
  auto meas = c.add_measure(1, 0);

  auto cl_input = c.c_inputs()[0];
  EdgeVec preds;
  preds.push_back(c.get_nth_out_edge(cl_input, 0));
  preds.push_back(c.get_nth_out_edge(meas, 0));
  auto ptr =
      std::make_shared<Conditional>(get_op_ptr(OpType::Rz, 0.1, 1), 1, 0);
  auto new_v = c.add_vertex(ptr);
  op_signature_t sigs;
  sigs.push_back(EdgeType::Boolean);
  sigs.push_back(EdgeType::Quantum);
  c.rewire(new_v, preds, sigs);
```

This circuit definition is slightly awkward, as I don't think there is any `add_op`-type calls I can use to build the circuit as needed. The "trick" here is that a measurement is performed and a quantum gate is then added on the measured qubit post-measurement. However, the gate is classically controlled on the classical value of the classical bit _before_ the measurement. It is a valid DAG, but there is no linear order of the "measure" and "add_conditional" operations that would yield this graph. As far as I can tell, though, it is valid. It looks as follows:

image

The following assertion then fails:

```c++
  auto slice_it = c.slice_begin();
  assert(slice_it.finished() || (*++slice_it).size() > 0);
```
AFAIK this is an invalid state for the iterator to be in.

Full MWE:
```c++
#include <iostream>
#include <tkassert/Assert.hpp>
#include <tket/Circuit/Circuit.hpp>

#include "tket/Circuit/DAGDefs.hpp"
#include "tket/Gate/OpPtrFunctions.hpp"
#include "tket/OpType/EdgeType.hpp"
#include "tket/Transformations/BasicOptimisation.hpp"
#include "tket/Transformations/Decomposition.hpp"

using namespace tket;

int main() {
  Circuit c(2, 1);
  c.add_op<unsigned>(OpType::CZ, {0, 1});
  auto meas = c.add_measure(1, 0);

  auto cl_input = c.c_inputs()[0];
  EdgeVec preds;
  preds.push_back(c.get_nth_out_edge(cl_input, 0));
  preds.push_back(c.get_nth_out_edge(meas, 0));
  auto ptr =
      std::make_shared<Conditional>(get_op_ptr(OpType::Rz, 0.1, 1), 1, 0);
  auto new_v = c.add_vertex(ptr);
  op_signature_t sigs;
  sigs.push_back(EdgeType::Boolean);
  sigs.push_back(EdgeType::Quantum);
  c.rewire(new_v, preds, sigs);

  auto slice_it = c.slice_begin();
  TKET_ASSERT(slice_it.finished() || (*++slice_it).size() > 0);
  return 0;
}
```

cqc-alec:
Fixes #1324 .
Fixes #1357 .

I was a bit concerned about the performance impact of this but it increases the running time of the tket tests by only about 1% so I think it's OK. (If it does turn out to be a problem, an alternative would be never to commute these gates, but that would give poorer results -- or somehow to find a cheaper way to do the check.)

### DIFF
--- a/pytket/conanfile.py
+++ b/pytket/conanfile.py
@@ -38,7 +38,7 @@ class pytketRecipe(ConanFile):
         self.requires("pybind11_json/0.2.14")
         self.requires("symengine/0.13.0")
         self.requires("tkassert/0.3.4@tket/stable")
-        self.requires("tket/1.3.54@tket/stable")
+        self.requires("tket/1.3.55@tket/stable")
         self.requires("tklog/0.3.3@tket/stable")
         self.requires("tkrng/0.3.3@tket/stable")
         self.requires("tktokenswap/0.3.9@tket/stable")
--- a/pytket/docs/changelog.rst
+++ b/pytket/docs/changelog.rst
@@ -8,6 +8,8 @@ Fixes:
 
 * Fix circuit iteration giving invalid slices in some cases.
 * Use built-in int type for get_counts() instead of numpy int types.
+* Add check for validity of resultant graph before pushing conditional gates
+  through other vertices in squashing passes.
 
 Performance:
 
--- a/tket/conanfile.py
+++ b/tket/conanfile.py
@@ -23,7 +23,7 @@ from conan.errors import ConanInvalidConfiguration
 
 class TketConan(ConanFile):
     name = "tket"
-    version = "1.3.54"
+    version = "1.3.55"
     package_type = "library"
     license = "Apache 2"
     homepage = "https://github.com/CQCL/tket"
--- a/tket/include/tket/Transformations/SingleQubitSquash.hpp
+++ b/tket/include/tket/Transformations/SingleQubitSquash.hpp
@@ -145,6 +145,14 @@ class SingleQubitSquash {
       const Circuit &sub, const VertexVec &single_chain, Edge &e,
       const Condition &condition);
 
+  // whether there is a path in the DAG to v from any vertex in vs (if
+  // reversed_ is true then paths are considered in the reverse direction)
+  bool path_exists(const Vertex &v, const VertexSet &vs) const;
+
+  // whether we are allowed to commute a given conditional single-qubit gate
+  // through the target of e without invalidating the DAG structure
+  bool commute_ok(const Edge &e, const Condition &condition) const;
+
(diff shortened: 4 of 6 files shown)


lmondada (approved):
Thanks @cqc-alec!

For future reference, I'll mention here that an alternative that was discussed was to forbid the commutation of classically controlled gates, when one of the bits it is controlled on is i) overwritten by a future measurement (when squashing forward), or ii) is used as control for another operation in the past (when squashing backward).

This would be a sufficient condition to ensure correctness and would be very quick to check. It would still do the right thing for a (the most common?) subset of conditionally controlled gates, so might be a good half-way solution if the full check implemented here becomes too expensive.

PR #1718, opened 2024-12-11
Rebase to TK2 via TK2 rather than CX

### LINKED ISSUE #1708: `SynthesiseTK` increases 2-qubit gate count for circuits with conditional 2-qubit gates
```python
from pytket.passes import SynthesiseTK

c = Circuit(2, 1).ZZPhase(0.5, 0, 1, condition_bits=[0], condition_value=1)
SynthesiseTK().apply(c)
print(c.n_2qb_gates())
```
outputs 2.

This has knock-on effects on default compilation passes at optimization levels 1 and 2 in backends, which can perform worse than optimization level 0 when the circuit contains many conditional 2-qubit gates.

cqc-alec:
Fixes #1708 .


yao-cqc (approved):
Nice one. Thanks!

PR #1721, opened 2024-12-12
fix: remove unecessary `_tket` perfix in local docs build

### LINKED ISSUE #1719: The docs show `pytket._tket` instead of just `pytket`
See for example: https://docs.quantinuum.com/tket/api-docs/passes.html#pytket._tket.passes.AutoRebase

The `_tket` part shouldn't be visible; it's not needed to import the submodules.

CalMacCQ:
# Description

Note that this only impacts the local build. The website build is already fixed and deployed.

Updating the build script to replace all of the refernce to the internal `_tket` module. Previously this was done by commands in the `Makefile` but this stopped working due to some previous changes.

I've decided to remove the `Makefile` for now as it was unused. Instead I've put the sed replace in the `build-docs.sh` script.

# Related issues
closes #1719


pytket/docs/build-docs.sh
```diff
@@ -3,6 +3,8 @@ cp -R pytket-docs-theming/quantinuum-sphinx .
 cp pytket-docs-theming/conf.py .
 
 sphinx-build -b html . build -D html_title="pytket"
+find build/ -type f -name "*.html" | xargs sed -e 's/pytket._tket/pytket/g' -i ""
```
CalMacCQ, line 6: slightly worried that these sed comands are not architecture independent. They work fine on my mac.

The website uses linux runners in C.I. so will make sure to use the linux syntax there.

cqc-alec: Could also check the platform from within the shell script and perform a switch on that.

cqc-alec: Can confirm that this does not work on Linux.

cqc-alec (commented):
I think we need to make this work at least on Linux and MacOS.

CalMacCQ:
> I think we need to make this work at least on Linux and MacOS.

Yes, happy to update the script to check the platform its running on and adjust the sed command accordingly.

CalMacCQ:
> I think we need to make this work at least on Linux and MacOS.

Now added some handling so that the build script should work on MacOS and Linux. I've tested locally 

Note that the webiste is already fixed. This script is only for local builds

cqc-alec (approved):

Issue #1605, opened 2024-10-04
Pytket QASM Importer Bug: Unsupported ISwap Gate `Gate has an invalid number of parameters`


MattePalte:
## Expected behavior

When using Pytket to import a QASM file containing an ISwap gate generated by Qiskit, we expect the import to succeed without any errors. The Pytket importer should be able to correctly parse the ISwap gate and its parameters.

## Actual behavior

The Pytket importer crashes when trying to import the QASM file, throwing a "Gate has an invalid number of parameters" error. This error occurs because the Pytket importer does not support the ISwap gate with two parameters.

## Additional information

This issue is 100% reproducible. We have verified that the Qiskit-generated QASM file is correct and can be imported by other tools. The issue seems to be specific to the Pytket importer.

## Source code

```python
from qiskit import QuantumCircuit, QuantumRegister, ClassicalRegister
from qiskit.qasm import qasm_data
from pytket.qasm import circuit_from_qasm_str

# Create a Qiskit circuit with an ISwap gate
qr = QuantumRegister(3, 'qr')
cr = ClassicalRegister(3, 'cr')
qc = QuantumCircuit(qr, cr, name='qc')
qc.iswap(qr[1], qr[2])

# Export the Qiskit circuit to a QASM file
qasm_content = qc.qasm()

# Try to import the QASM file using Pytket
try:
    circuit = circuit_from_qasm_str(qasm_content)
    print(f'Circuit (Pytket) imported correctly: ')
except Exception as e:
    print(e)  # Gate has an invalid number of parameters

print(qasm_content)
# Output:
# OPENQASM 2.0;
# include "qelib1.inc";
# gate iswap q0,q1 { s q0; s q1; h q0; cx q0,q1; cx q1,q0; h q1; }
# qreg qr[3];
# creg cr[3];
# iswap qr[1],qr[2];
```

## Tracebacks

```
Gate has an invalid number of parameters
```

## System

```python
import qiskit
import pytket
print(f"Qiskit version: {qiskit.__version__}")
print(f"Pytket version: {pytket.__version__}")
```

Output:

```
Qiskit version: 1.2.0
Pytket version: 1.31.1
```

cqc-alec:
[...]

PR #1739, opened 2025-01-15
Update `CustomGate` handling in `circuit_from_qasm_...`

sjdilkes:
# Description
Reading in QASM supports custom gate definitions. If the custom gate has the same name as an existing supported op, we check whether the circuits are equivalent, and if they are we use are pre-existing definition. If they aren't, we write in the user supported op. However, the equivalency check works by producing the circuit, and if the new definition and our held definition support a different number of parameters, then the code would attempt to produce the circuit while passing the wrong number of parameters, and so error out.

In this code, we now check the number of parameters is correct before producing a circuit. This circuit is still checked foe equivalency.

# Related issues
https://github.com/CQCL/tket/issues/1605

# Checklist


cqc-alec (commented):
LGTM, just a small suggestion.

cqc-alec (approved):

PR #1755, opened 2025-01-24
feat: add `QubitPauliOperator.get_dict`

CalMacCQ:
# Description

Please summarise the changes.

Minor change adding a `get_dict` export for `QubitPauliOperator`. This avoids having to used the undocumented private `_dict` attribute` 

Using `.get_dict` as opposed to `to_dict` here as this new method is not intended for json serialisation.

# Related issues

# Checklist

### DIFF
--- a/pytket/pytket/utils/operators.py
+++ b/pytket/pytket/utils/operators.py
@@ -218,7 +218,8 @@ class QubitPauliOperator:
 
     def get_dict(self) -> dict[QubitPauliString, Expr]:
         """Generate a dict representation of QubitPauliOperator,
-        mapping each :py:class:`QubitPauliString` in the support to its corresponding value.
+        mapping each :py:class:`QubitPauliString` in the support
+        to its corresponding value.
 
         :return: A dict of Pauli strings and their coefficients
           as key-value pairs
--- a/pytket/pytket/utils/operators.py
+++ b/pytket/pytket/utils/operators.py
@@ -218,7 +218,7 @@ class QubitPauliOperator:
 
     def get_dict(self) -> dict[QubitPauliString, Expr]:
         """Generate a dict representation of QubitPauliOperator,
-        not suitable for writing to JSON.
+        mapping each :py:class:`QubitPauliString` in the support to its corresponding value.
 
         :return: A dict of Pauli strings and their coefficients
           as key-value pairs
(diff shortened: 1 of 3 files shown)


pytket/tests/utils_test.py
```diff
@@ -90,6 +90,13 @@ def test_all_paulis() -> None:
     assert len(list(circs)) == 3
 
 
+def test_dict_export() -> None:
+    qps1 = QubitPauliString(Qubit(0), Pauli.Y)
+    qps2 = QubitPauliString(Qubit(0), Pauli.X)
+    op = QubitPauliOperator({qps1: 1j, qps2: 0.5})
+    assert op.get_dict() == op._dict
```
CalMacCQ, line 97: This is maybe a bit cheeky. I could do the explcit comparision if prefferred.

cqc-alec: Yes I think we should do an explicit comparison, to be safe against future surprises.

pytket/pytket/utils/operators.py
```diff
@@ -216,6 +216,15 @@ def subs(self, symbol_dict: dict[Symbol, complex]) -> None:
         for key, value in self._dict.items():
             self._dict[key] = value.subs(symbol_dict)
 
+    def get_dict(self) -> dict[QubitPauliString, Expr]:
+        """Generate a dict representation of QubitPauliOperator,
+        not suitable for writing to JSON.
```
cqc-alec, line 221:
```suggestion
        mapping each :py:class:`QubitPauliString` in the support to its corresponding value.
```

pytket/tests/utils_test.py
```diff
@@ -22,7 +22,7 @@
 from hypothesis import HealthCheck, given, settings, strategies
 from hypothesis.strategies._internal import SearchStrategy
 from simulator import TketSimBackend, TketSimShotBackend  # type: ignore
-from sympy import symbols
+from sympy import I, symbols
```
CalMacCQ, line 25: I noticed that pylint doesn't seem to detect the unused imports here

`Dict, List, Tuple` are not used.

cqc-alec: Ah, this is because we disable https://docs.astral.sh/ruff/rules/unused-import/ (F401) in the `ruff.toml` file. I disabled it because I wanted to allow unused imports in init files, but it would be good to disallow them in other files. Will make an issue to fix this.

cqc-alec (approved):

PR #1758, opened 2025-01-27
Don't ignore unused imports except in init files

### LINKED ISSUE #1756: Disallow unused imports except in init files
Currently `ruff.toml` ignores F401 warnings (unused imports) because I wanted to allow them in init files, but we should disallow them elsewhere.

cqc-alec:
Closes #1756 .


CalMacCQ (approved):
Looks good to me, thanks for the fix

PR #1764, opened 2025-01-31
Minor docs cleanup.

cqc-alec:
Including inserting some narrow non-breaking spaces to keep sphinx happy.

### DIFF
--- a/pytket/docs/changelog.rst
+++ b/pytket/docs/changelog.rst
@@ -116,7 +116,7 @@ Fixes:
 
 * Fix `symbol_substitution` not preserving opgroups.
 * Remove hardware inefficient circuit construction in `_tk1_to_rzsx`
-* Support converting conditional `RangePredicate`s to QASM.
+* Support converting conditional `RangePredicate` s to QASM.
 * Fix `maxwidth` parameter of `circuit_from_qasm_str`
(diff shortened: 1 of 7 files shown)


pytket/docs/changelog.rst
```diff
@@ -116,7 +116,7 @@ Fixes:
 
 * Fix `symbol_substitution` not preserving opgroups.
 * Remove hardware inefficient circuit construction in `_tk1_to_rzsx`
-* Support converting conditional `RangePredicate`s to QASM.
+* Support converting conditional `RangePredicate` s to QASM.
```
cqc-melf, line 119: In the other parts of the changelog we use `` instedt of ` , should we do that consistent?

cqc-melf:
```suggestion
* Support converting conditional ``RangePredicate`` s to QASM.

cqc-melf: As the changelog is now an rst file, can we link to the documentation of the classes / functions in a good way?

cqc-alec:
These are good suggestions but for now I just wanted to eliminate some sphinx warnings, so would rather leave further cleanup for the future.

cqc-melf (approved):

PR #1767, opened 2025-02-05
Add `Circuit.add_circuit_with_map()` to docs.

cqc-alec:
Drive-by: Fix mypy issues following update of mypy to 1.15.0.


CalMacCQ (approved):
Thanks :)

PR #1772, opened 2025-02-10
Add `Circuit.has_implicit_wireswaps` property

### LINKED ISSUE #1665: Add `Circuit.has_implicit_permutation`
This property would give a Boolean value indicating whether a `Circuit` has a non-trivial permutation of the qubits.

This would save having to check in python with something like the following

```python
def _has_implicit_permutation(circ: Circuit) -> bool:
    return any(q0 != q1 for q0, q1 in circ.implicit_qubit_permutation().items())
```

jpacold:
# Description

* Added a readonly property bound to `Circuit::has_implicit_wireswaps()`
* Added a unit test
* Updated some existing tests -- replaced calls to `Circuit.implicit_qubit_permutation()` if the test only used the result to check for non-trivial permutations

# Related issues

Closes #1665.

# Checklist

### DIFF
--- a/pytket/docs/changelog.rst
+++ b/pytket/docs/changelog.rst
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+1.41.0 (Unreleased)
+----------------------
+Features:
+
+* Add :py:attr:`Circuit.has_implicit_wireswaps` property.
+
 1.40.0 (February 2025)
 ----------------------
 
@@ -9,7 +15,6 @@ Features:
 * Add Python 3.13 support to the ZX module.
 * Add :py:meth:`QubitPauliOperator.get_dict` method.
 * Add :py:meth:`Circuit.add_circuit_with_map` method.
-* Add :py:attr:`Circuit.has_implicit_wireswaps` property.
 
 Fixes:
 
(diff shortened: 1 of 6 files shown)


pytket/docs/changelog.rst
```diff
@@ -9,6 +9,7 @@ Features:
 * Add Python 3.13 support to the ZX module.
 * Add :py:meth:`QubitPauliOperator.get_dict` method.
 * Add :py:meth:`Circuit.add_circuit_with_map` method.
+* Add :py:attr:`Circuit.has_implicit_wireswaps` property.
```
cqc-alec, line 12: This should be under a new "Unreleased" heading above please.

pytket/tests/predicates_test.py
```diff
@@ -1140,6 +1140,18 @@ def test_decompose_inside_conditionals() -> None:
...
 
+def test_has_implicit_wireswaps() -> None:
+    c = Circuit(2)
+    c.SWAP(0, 1)
+    assert not c.has_implicit_wireswaps
+    c.replace_SWAPs()
+    assert c.has_implicit_wireswaps
+
+    # Property should be read-only
+    with pytest.raises(AttributeError):
```
cqc-alec, line 1151: It seems that a linter is complaining that this line is unreachable; I don't really see why...

cqc-alec (approved):

PR #1775, opened 2025-02-13
Handle nested conditionals, and subsequent bugs

### LINKED ISSUE #1727: Rebase passes do not work with nested conditionals
Follow-up from https://github.com/CQCL/pytket-qiskit/issues/442 .

The following circuit contains a nested conditional X gate, which is not rebased:

```python
from pytket.circuit import Circuit, Bit, CircBox, Qubit, OpType
from pytket.passes import AutoRebase, DecomposeBoxes

c0 = Circuit(1, 1)
c0.X(0, condition_bits=[0], condition_value=1)
cbox = CircBox(c0)
c = Circuit(1, 2)
c.add_circbox(cbox, [Qubit(0), Bit(1)], condition_bits=[Bit(0)], condition_value=1)
DecomposeBoxes().apply(c)

rpass = AutoRebase({OpType.TK1, OpType.TK2})

rpass.apply(c)
print(c.get_commands())
```

Output:
```
[IF ([c[0]] == 1) THEN IF ([c[1]] == 1) THEN X q[0];]
```

willsimmons1465:
What begun as a small bugfix for #1727 has rabbit-holed into a long sequence of related changes, many of which through stages of "I have made a small change to this bit of code so I need to verify that it preserves the semantics - oh wait, there aren't tests for this behaviour, let me try some - oh wait, there are bugs here, let me fix them - oh wait, this code is used somewhere, let me take a look...".

Below lists the different stages of ~~madness~~ _bugs_ I went through in making these changes. Let me know if it would be helpful for me to split the changes up into separate PRs for easier git history or reviewing.

# Description

- Rebase passes would only unpack up to one layer of conditionals to look for the gates. Anywhere in the codebase that expanded conditionals to find the underlying instruction now iteratively unpacks all conditional layers.
- In particular, `Circuit::substitute_conditional` would take the circuit replacement for the underlying gate and extend it with any conditionals to replace the conditional gate. The `Subcircuit` interface is used to represent the edges incident on the conditional gate being replaced, in which the edges are partitioned by edge type into quantum or classical. The classical list doubles up for both classical (write access) and boolean (read access). When fixing the code for rebasing nested conditionals, I encountered a bug during circuit substitution with nested conditionals where these wires would get mixed up. The solution to this was to rewrite the `Subcircuit` class to more closely match the signature of an `Op`, where there is just one list of in-edges that mixes different kinds but it is easy to identify them as their position in the list matches the port.
  - Whilst fixing `Subcircuit` and substitution in this way, I also extended it to cope with WASM edges.
  - Everywhere in the code that uses `Subcircuit` needed changing to fit the new interface. Many of these were easy to replace with `Circuit::singleton_subcircuit` which helps to centralise the logic for constructing subcircuits.
  - In particular, single qubit squashing now has to track a list of nested conditions rather than just an optional condition.
- Another bug encountered during testing the fix to rebasing with nested conditionals was that I needed to track the global phase within these substitutions, thus giving rise to conditional global phase operations. This lead to some compiler passes failing their predicate checks because they weren't expecting to produce `Phase` ops.
  - At the same time, phase polynomial passes were not tracking the global phase but proptests required them to preserve it, so I fixed that.
- The `Circuit::commuting_basis()` method unpacks conditionals, but seemed needlessly overcomplicated. Every place that calls it asks to look up the basis at a particular port on the vertex, but the method itself interprets that number not as a port but as the index within the subset of quantum ports. Searching for an explanation, I traced this back to the docstring for `Op::commuting_basis` which specified it in this way despite the fact that, in the only implementation (`Gate::commuting_basis`), the notions of qubit index and port coincide. I simplified this by changing the docstring to specify ports, get conditionals to unpack themselves appropriately in their overloaded method, and simplifying `Circuit::commuting_basis`.
  - The same applies to `Circuit::commutes_with_basis`. However, I realised that the implementation of `Gate::commutes_with_basis` is general enough to work for any `Op` (i.e. it would be copied verbatim to an overload for `Conditional`), so I just raised it to `Op`.
  - I moved the definition of `port_t` from DAGDefs.hpp to Ops.hpp to help the documentation of these methods. Given other docstrings in Ops.hpp already refer to ports, this is a fine place for the concept to be defined.
  - Looking at places where `Op::commutes_with_basis` is used, I identified a few bugs in the `DelayMeasures` pass, such as tracing a qubit commutation shifting to a classical wire on a conditional gate, or passing through a conditional SWAP (which it shouldn't do as the qubit wire we would need to subsequently measure is not deterministic). I fixed this up whilst I was here.

# Related issues

Closes #1727 

# Checklist


cqc-alec (commented):
Great work, thanks for giving this area of the code some much-need love.

cqc-alec (approved):

PR #1780, opened 2025-02-24
Remove deprecated APIs

### LINKED ISSUE #1737: Remove deprecated classes and methods
We have a number of formally deprecated classes and methods in pytket. From a quick search these are:

- `ClassicalExpBox` and friends;
- `GlobalisePhasedX`;
- `SynthesiseUMD`;
- `NoiseAwarePlacement.modify_config`;
- `GraphPlacement.modify_config`.

So as not to break semver we should make a pytket 2.0 release with these removed. We could consider other small breaking changes at the same time but I would like to minimize the chances of _actually_ breaking people's code, to encourage updating.

At the same time we should probably also make a pytket-1 branch which we can maintain for a certain period with hotfixes.

I suggest we make the 2.0 release in late February.

cqc-alec:
And bump major version to 2.0.0.

After merging this my plan is to:
- create a pytket1 branch based off the preceding commit;
- make a pytket 1.41.0 release from that branch;
- make a pytket 2.0.0 release from `main`;
- update dependencies that are currently constrained to use pytket < 2 so that they work with pytket 2.

Hopefully no-one will then need to depend on pytket < 2 -- I've done some due diligence here but may have missed some things, so we'll continue to support pytket 1.x until any issues are ironed out.

May be easiest to review one commit at a time.

It's a pity about the drop in coverage but this must be because the APIs I'm removing were better-tested than average -- I don't think I've added any new untested code or removed tests for non-removed code.

Closes #1737 .


cqc-alec:
Needs merge of #1786 to pass CI.

cqc-melf (approved):

PR #1791, opened 2025-02-26
Run normal PR checks on pytket1 branch.

cqc-alec:


.github/workflows/build_and_test.yml
```diff
@@ -3,10 +3,10 @@ name: Build and test
 on:
   pull_request:
     branches:
-      - main
```
cqc-melf, line 6: This changes the current file, not adding a copy? Do we still have the original file somewhere?

cqc-alec: Yes, this is just being merged to the `pytket1` branch; the original file is still on `main`.

I want to make the changes to this branch minimal, so that we're less likely to run into merge conflicts when cherry-picking from `main` in the future.

cqc-melf (approved):

PR #1792, opened 2025-02-27
Zx optimisation swaps

### LINKED ISSUE #1752: Add `allow_swaps` parameter to `ZXGraphlikeOptimisation`
So that implicit swaps can be disallowed if desired.

jpacold:
# Description

* Added an optional argument to `ZXGraphLikeOptimisation` so that the signature is now
```cpp
const PassPtr &ZXGraphlikeOptimisation(bool allow_swaps = true);
```
The functions in PassLibrary.cpp all use a pattern where we create a local `static const PassPtr` and return it by const reference. In this case, the logic in the `Transform` object used to create the `PassPtr` depends on the value of `allow_swaps`. So I think we need to have two `PassPtr`s, and `allow_swaps` just determines which one we return by reference.

* Added a C++ test, the Python binding, and a Python test.

# Related issues

Closes #1752.

### DIFF
--- a/pytket/binders/passes.cpp
+++ b/pytket/binders/passes.cpp
@@ -657,7 +657,7 @@ PYBIND11_MODULE(passes, m) {
       "beforehand and may increase the cost of the circuit."
       "\n:param allow_swaps: Whether to allow implicit wire swaps (default "
       "True).",
-      py::arg("allow_swaps") = false);
+      py::arg("allow_swaps") = true);
 
   /* Pass generators */
 
--- a/pytket/binders/passes.cpp
+++ b/pytket/binders/passes.cpp
@@ -655,7 +655,7 @@ PYBIND11_MODULE(passes, m) {
       "not work if the circuit contains created or discarded qubits. As a "
       "resynthesis pass, this will ignore almost all optimisations achieved "
       "beforehand and may increase the cost of the circuit."
-      "\n:param allow_swaps: Whether to allow implicit wire swaps (default "
+      "\n\n:param allow_swaps: Whether to allow implicit wire swaps (default "
       "True).",
       py::arg("allow_swaps") = true);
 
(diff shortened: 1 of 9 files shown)


cqc-alec:
Thanks for opening this PR. Because you made changes to the C++ code you need to bump the tket version in `tket/conanfile.py` and `pytket/conanfile.py`.

cqc-alec:
One more thing needed to get CI to pass: need to regenerate the type stubs. You can do this by running the script `pytket/stub_generation/regenerate_stubs` (need to `pip install pybind11-stubgen` first); or look at the diff printed in the CI logs.

Also, need to rebase to `main` and again bump the tket library version, sorry about that.

pytket/binders/passes.cpp
```diff
@@ -654,7 +654,10 @@ PYBIND11_MODULE(passes, m) {
       "extracting a circuit back out. Due to limitations in extraction, may "
       "not work if the circuit contains created or discarded qubits. As a "
       "resynthesis pass, this will ignore almost all optimisations achieved "
-      "beforehand and may increase the cost of the circuit.");
+      "beforehand and may increase the cost of the circuit."
+      "\n:param allow_swaps: Whether to allow implicit wire swaps (default "
```
cqc-alec, line 658:
```suggestion
      "\n\n:param allow_swaps: Whether to allow implicit wire swaps (default "
```
Needed for correct formatting I think.

pytket/docs/changelog.rst
```diff
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+2.0.1 (Unreleased)
```
cqc-alec, line 4:
```suggestion
Unreleased
```

cqc-alec (approved):
Thanks!

jpacold:
Thanks for the review and cleanup.

PR #1803, opened 2025-03-07
Fix error in `separate_classical` when quantum ops succeed final measures

### LINKED ISSUE #1794: `Transforms::separate_classical()` fails if circuit has a `Barrier` at the end
```python
from pytket.transform import separate_classical

circuit = Circuit(1, 1).X(0).Measure(0, 0).add_barrier([0])

_, _ = separate_classical(circuit)
```

This fails with the error: `Circuit does not contain unit with id: q[0]`.

cqc-alec:
Fixes #1794 .


yao-cqc (approved):
Thanks!

Issue #1725, opened 2024-12-31
Add CnX decompositions with ancilla qubits


sjdilkes:
See qiskit implementations here -> https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.QuantumCircuit#mcx

https://arxiv.org/pdf/1508.03273

jpacold:
Could I take part of this, say implementing [`MCXVChain`](https://github.com/Qiskit/qiskit/blob/731e4744e9499e2f8063ddc793028aba70e625b0/qiskit/circuit/library/standard_gates/x.py#L1363)?

If so, should it just be a new function in [CircPool.hpp](https://github.com/CQCL/tket/blob/main/tket/include/tket/Circuit/CircPool.hpp)?

yao-cqc:
@jpacold Much appreciated, thanks for taking this on!

Implementing `MCXVChain` in `CircPool` and binding it in [circuit_library.cpp](https://github.com/CQCL/tket/blob/main/pytket/binders/circuit_library.cpp) would be a good start.  

In the bigger picture, tket doesn't have a mechanism to manage ancilla qubits, so integrating this CnX decomposition method into the current rebase pipeline wouldn’t be straightforward. It might be useful to support CnX gates as `Boxes`, allowing users to specify how each should be decomposed. But for now, I think adding `MCXVChain` to `CircPool` is a necessary step forward.

PR #1805, opened 2025-03-11
Decompositions of CnX gates with ancilla qubits

jpacold:
# Description

Implements CnX decompositions from [Maslov (2015)](https://arxiv.org/abs/1508.03273). Related to #1725.

~~Opening this as a draft because there is a problem with the new test that I don't understand, and I wanted to check if it is reproducible in CI. The test passes if I run it with `./test-tket "Scenario: Test CnX_vchain_decomp"`. If I run the whole suite with `./test-tket`, it fails when checking the unitary matrix of the synthesized circuit.~~ I'm still not sure why this is happening; the same check seems to be OK on the Python side.

The paper includes two decompositions (given in Proposition 4 and Proposition 5), both using $\lfloor (n-1)/2 \rfloor$ ancilla qubits to implement a gate with $n$ controls. [Qiskit](https://github.com/Qiskit/qiskit/blob/main/qiskit/synthesis/multi_controlled/mcx_synthesis.py#L137) seems to have a simplified version of the one described in Proposition 4, using roughly the same gate counts but more ancillas ($n-2$).

### DIFF
--- a/pytket/binders/circuit_library.cpp
+++ b/pytket/binders/circuit_library.cpp
@@ -121,6 +121,18 @@ PYBIND11_MODULE(circuit_library, library_m) {
       "C4X_normal_decomp", &CircPool::C4X_normal_decomp,
       "Equivalent to CCCCX, using 36 CX ");
   library_m.def(
+      "CnX_vchain_decomp", &CircPool::CnX_vchain_decomp,
+      "CnX decomposition from https://arxiv.org/abs/1906.01734/1508.03273.\n\n"
+      ":param n: Number of control qubits\n"
+      ":param zeroed_ancillas: If True, the gate will be implemented assuming "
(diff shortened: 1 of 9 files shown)


cqc-alec:
Strange, it seems it only fails on MacOS, and there is a mysterious "bus error" appearing. This reminds me of an issue @sjdilkes encountered recently, which I'm not sure he got to the bottom of...

jpacold:
Still trying to understand the problem. For now I have moved the relevant part of the test to Python.

yao-cqc (approved):
Thanks! Everything looks good to me

cqc-alec:
Thanks @jpacold ! Just need to update from `main` and probably bump the tket version again, and then we can merge this.

cqc-alec (approved):

PR #1809, opened 2025-03-12
docs: Remove old warning from main page, docs fixes

CalMacCQ:
# Description

Removing this banner as I think its outdated and the issue won't impact any current users. Let me know if you disagree.

Screenshot 2025-03-12 at 20 39 22

Driveby: updated the docs theming submodule -> only impacts the local build of the docs.
Driveby II: add some missing Circuit methods into the docs


cqc-alec (approved):
Thanks!

PR #1811, opened 2025-03-13
Restore and fix two tests

cqc-alec:
There were two mysterious test failures in [this commit](https://github.com/CQCL/tket/pull/1805/commits/e323b93508a43269584ddebc4fed2c5541fb569e), which was submitted to [this PR](https://github.com/CQCL/tket/pull/1805). The tests were removed from that PR. This PR [restores the tests](https://github.com/CQCL/tket/pull/1811/commits/7268209eec36924e9bc0ad9758d5cba14994835b) and fixes two issues:
* https://github.com/CQCL/tket/pull/1811/commits/2e9ac14ecab0760c1f78a82adfea9dd2f6d21f7e -- Properly initialize a matrix.
* https://github.com/CQCL/tket/pull/1811/commits/9ba89889f8aac77c130ce1ab41b7a760af792aa5 -- This one is more strange: I don't know why it was giving a segfault, but written this way there is no segfault (and valgrind is happy).


jpacold:
Thanks for fixing this! I should've noticed the uninitialized matrix.

I incorrectly assumed that `Eigen::kroneckerProduct` returns a matrix. It actually returns [an object](https://eigen.tuxfamily.org/dox/unsupported/KroneckerTensorProduct_8h_source.html) that holds copies of the two arguments and evaluates the Kronecker product lazily. For matrices, these are shallow copies, so using a temporary matrix causes a segfault.

cqc-alec:
> Thanks for fixing this! I should've noticed the uninitialized matrix.
> 
> I incorrectly assumed that `Eigen::kroneckerProduct` returns a matrix. It actually returns [an object](https://eigen.tuxfamily.org/dox/unsupported/KroneckerTensorProduct_8h_source.html) that holds copies of the two arguments and evaluates the Kronecker product lazily. For matrices, these are shallow copies, so using a temporary matrix causes a segfault.

Ah that makes sense -- and good to know.

PR #1815, opened 2025-03-17
Check op validity on directed architectures

### LINKED ISSUE #1784: Clarify "directedness" in `Architecture.valid_operation`
```python
from pytket.architecture import Architecture
from pytket.unit_id import Node

arch = Architecture([(Node(0), Node(2))])
print(arch.valid_operation([Node(2), Node(0)])) # prints True, directedness is ignored.
```

This is confusing as this directedness constraint can be expressed in pytket. We should update the docs here or update the method to check for the direction of the edge.

yao-cqc:
# Description

Add `bidirectional` argument to `Architecture.valid_operation()`

# Related issues

resolves #1784 

# Checklist


cqc-alec (approved):
Thanks!

Issue #1580, opened 2024-09-16
Improve error message for mismatching unit ids when adding conditional gates


CalMacCQ:
The following method works fine for adding a condtional-$X$ gate to a `Circuit`.

```python
from pytket import Circuit

circ = Circuit()
q = circ.add_q_register("q", 1) # default "q" register
a = circ.add_c_register("a", 1)

circ.X(q[0], condition_bits=[a[0]], condition_value=1)
```

Notice how I am passing the `Qubit` `q[0] into the `Circuit.add_gate` method.

However if I replace `q[0]` with `0`....

```python
from pytket import Circuit

circ = Circuit()
q = circ.add_q_register("q", 1) # default "q" register
a = circ.add_c_register("a", 1)

circ.X(0, condition_bits=[a[0]], condition_value=1)
``` 

I get the following `RuntimeError`

```python
--
RuntimeError                              Traceback (most recent call last)
Cell In[16], line 1
----> 1 circ.X(0, condition_bits=[a[0]], condition_value=1)

RuntimeError: Unable to cast Python instance of type <class 'list'> to C++ type 'std::__1::vector<unsigned int, std::__1::allocator<unsigned int>>'
```

Its not at all obvoius from this error what has gone wrong here. This error message should make the fix clearer.

In fact, arguably this shouldn't give an error in pytket as in the current interface we support referencing qubits with `Qubit`s and integers. However not sure how practical overload merging would be.

willsimmons1465:
This error is thrown by pybind.
When we expose multiple methods with the same name and arity, pybind creates one python method with that name and performs type inference on the arguments to determine which of the underlying methods to call. It spits out this generic error message whenever this type inference fails.

In order for us to change the error message, we would have to take manual control away from pybind for handling this. I messed around last year with only ever exposing e.g. `_Circuit`, and defining the python class `Circuit` which acts as a thin wrapper around it and I manually do the type inference. This has the added benefits of being able to fully customise the docstring and type annotations as well (which can be a massive improvements to the API docs to cut down on all of the pointless "Overloaded function" blocks with up to 12 subtly different cases for essentially the same method), as well as making everything in pytket searchable through the python source. However, doing this on mass is a very large amount of work and increases the complexity of the codebase by adding yet another distinct binding layer to maintain.

[...]

PR #1832, opened 2025-04-03
Use variants to merge methods in pybind

willsimmons1465:
# Description

Many of the methods on the pytket `Circuit` class are overloaded to account for the ability for the user to specify arguments as either integers (which implicitly refer to the corresponding index in the standard register for qubits or bits) or explicitly using `Qubit`/`Bit`. This change makes use of variant types (a.k.a. unions) to combine methods whenever possible. This allows us to do things like providing custom error messages for certain kinds of type errors, since we are handling some of the type inference ourselves rather than it all being auto-generated by pybind.

If we like this and wish to push this further, I can imagine further changes off the back of this PR where we take more of the control away from pybind in order to have better control of error messages and API docs by performing more of the type inference ourselves.

# Related issues

Issue #1580 gives examples of some of the non-obvious error messages that occur from pybind's defaults. From the examples there, the changes here just allow the first example to work perfectly fine.

# Checklist

### DIFF
--- a/pytket/binders/circuit/Circuit/add_classical_op.cpp
+++ b/pytket/binders/circuit/Circuit/add_classical_op.cpp
@@ -29,6 +29,82 @@ namespace py = pybind11;
 
 namespace tket {
 
+typedef std::variant<
+    py::tket_custom::SequenceVec<unsigned>, py::tket_custom::SequenceVec<Bit>>
+    var_seq_bs_t;
+
+static std::variant<std::vector<unsigned>, std::vector<UnitID>>
+var_sequence_to_var_vecs(const var_seq_bs_t &var_seq) {
+  if (std::holds_alternative<py::tket_custom::SequenceVec<unsigned>>(var_seq)) {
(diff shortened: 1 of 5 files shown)


CalMacCQ (approved):
Looks great :) thanks for working on this.

After resolving some issues with my local build of tket I have verified that you can now use `UnitID`s and `int`s as in the second example without an error.

I've also looked at a local build of the docs and this is a nice improvement to the readibility of the gate methods on `Circuit`.

I think there should be a changelog entry to reflect that users now have more freedom to mix and match argument types?

The code looks good to me (I don't have much experience with binder code). Possibly worth asking someone else for feedback but I'll leave that up to you.

willsimmons1465:
> I think there should be a changelog entry to reflect that users now have more freedom to mix and match argument types. This is a user-facing change.

The code now contains explicit exceptions when users mix and match argument types in many cases to mimic the existing behaviour, so it isn't necessarily that much more flexible, but it is more clear. I have added a changelog entry to reflect the API change here.

cqc-alec (approved):

PR #1842, opened 2025-04-10
`Transforms::greedy_pauli_optimisation` performance

### LINKED ISSUE #1820: Performance improvement for `Transforms::greedy_pauli_optimisation`
`Transforms::greedy_pauli_optimisation` makes extensive use of lookup tables defined in [`GreedyPauliOptimisationLookupTables.hpp`](https://github.com/CQCL/tket/blob/main/tket/include/tket/Transformations/GreedyPauliOptimisationLookupTables.hpp). For at least some of these, the use of `unordered_map` seems to have a significant performance overhead for large circuits.

For example, [this function call in the tests](https://github.com/CQCL/tket/blob/59c45e971b1b0f5c0bc370c10295fd80f27d954a/tket/test/src/test_GreedyPauli.cpp#L836-L838) takes ~100 seconds on my machine (release build compiled with g++13). Profiling shows roughly $8 \times 10^8$ calls to each of [these two functions](https://github.com/CQCL/tket/blob/59c45e971b1b0f5c0bc370c10295fd80f27d954a/tket/src/Transformations/GreedyPauliOps.cpp#L69-L94), which each do one lookup in an `unordered_map`. [Replacing it with an `array`](https://github.com/CQCL/tket/compare/main...jpacold:tket:lookup-tables#diff-9126a091c528ace9a27e0b05a58f9ab88873b9b2451f7850ca236704d1bc5446) reduces the time to ~80 seconds. 

There might be more room to improve performance with other STL usage adjustments. If this seems worth looking into, I can spend some more time profiling and open a PR.

jpacold:
# Description

This PR includes several performance optimisations for `Transforms::greedy_pauli_optimisation`. I used the test `Scenario: Test GreedyPauliSimp with multiple trials and threads` as a benchmark. On my machine, these changes reduce the wall-clock time for [this function call](https://github.com/CQCL/tket/blob/59c45e971b1b0f5c0bc370c10295fd80f27d954a/tket/test/src/test_GreedyPauli.cpp#L836-L838) from 100 seconds to 60 seconds.

The test constructs a large Pauli graph, so most of the improvement came from making small adjustments to functions that are repeatedly called on every node:
* Approximately a 20-second reduction from rewriting the hash function for `TQE_PAULI_MAP` and replacing the `std::unordered_map` with a `std::array`
* Approximately a 13-second reduction due to returning early from `SingleNode::update` when the Pauli string doesn't need to be modified

The following changes each contributed a 1-2 second improvement:
* Adding a move-assignment operator for `Circuit` so that we avoid copying the (large) DAG. This is a significant interface change. Since it has marginal benefit in this case I would be okay with reverting it, though I imagine it could be generally nice to have for large circuits.
* Avoiding some calls to `Circuit::depth`
* Replacing some of the data structures that hold candidate TQE gates and their costs in `pauli_exps_synthesis`
* Reorganising some of the cost calculations

# Related issues

Closes #1820.

# Checklist


cqc-alec:
Thank you! The mypy errors are puzzling. I can't reproduce them locally...

cqc-alec:
Reproduced the mypy errors with numpy 2.2.4. I raised a separate issue: https://github.com/CQCL/tket/issues/1844 .

tket/src/Transformations/GreedyPauliOptimisation.cpp
```diff
@@ -166,33 +168,23 @@ static double default_pauliexp_tqe_cost(
...
-      if (pair.second[cost_index] < mins[cost_index]) {
-        mins[cost_index] = pair.second[cost_index];
-      }
-      if (pair.second[cost_index] > maxs[cost_index]) {
-        maxs[cost_index] = pair.second[cost_index];
-      }
-    }
-  }
-  for (const auto& pair : tqe_candidates_cost) {
-    TKET_ASSERT(pair.second.size() == n_costs);
```
yao-cqc, line 195: 🫠 nice catch

yao-cqc (commented):
All looks good to me! Thanks a lot for working on this. Would you mind adding an entry to the [changelog](https://github.com/CQCL/tket/blob/main/pytket/docs/changelog.rst)?

@cqc-alec do you have any comments on the move assignment operator?

cqc-alec:
> @cqc-alec do you have any comments on the move assignment operator?

Looks good to me!

jpacold:
Thanks for another prompt review! I updated the changelog.

PR #1850, opened 2025-04-11
Add inclusive option to DecomposeBoxes

### LINKED ISSUE #1617: Inclusive option for DecomposeBoxes
`DecomposeBoxes` has the `excluded_types` and `excluded_opgroups` parameters to allow exclusion of certain boxes from decomposition.  It would be very useful to also have inclusive versions of these, where the decomposition will only be performed on the specified types/opgroups.  In particular this would be useful when considering nested boxes.

willsimmons1465:
# Description

The `DecomposeBoxes` pass currently admits some optional arguments to filter which boxes get decomposed. The current filters are all exclusive, prohibiting the decomposition of boxes of given types or with particular opgroup names.

@andrew-tranter suggested (in #1617) adding additional filters that would act inclusively, so `DecomposeBoxes` would only decompose the boxes of the given types and opgroups.

I made the design decision to make all of these filters to interact with each other via intersection: a box is decomposed only if it is in both the inclusive sets (if they are given) AND it is in neither of the exclusive sets. Let me know if a different behaviour would be preferred.

# Related issues

Closes #1617 

# Checklist


pytket/docs/changelog.rst
```diff
@@ -7,6 +7,8 @@ Unreleased
 Features:
 
 * Enabling pickling of `Architecture` objects.
+* Added optional inclusive filters to the :py:meth:`DecomposeBoxes` pass to
```
CalMacCQ, line 10: Minor thing but I think if you do :py:class`~pytket.passes.DecomposeBoxes` the crosslink to the docs should be generated

yao-cqc (commented):
I'm not familiar with the point raised by @CalMacCQ . Everything else looks good to me.

pytket/docs/changelog.rst
willsimmons1465: Thanks for the hint, I have done it with :py:meth: instead to get the method formatting. Let me know if you think it's worth setting up a chore task at some point to go through the docs and make the crosslinks for any class or method reference currently in the docs

CalMacCQ (approved):

PR #1863, opened 2025-04-22
Subclass a couple of `StubGen` methods instead of reproducing the whole class

cqc-alec:


pytket/stub_generation/regenerate_stubs
```diff
@@ -2,14 +2,272 @@
...
 import importlib
 import os
-from pathlib import Path
 import pkgutil
 import re
+import types
+import typing
+from inspect import ismodule, signature
+from pathlib import Path
+from typing import Any, Callable, List, Optional, Sequence, cast
```
CalMacCQ, line 11: The builtin `list` should work here I think

cqc-alec: Yes -- but I wanted to keep the changes from the original nanobind implementation to a minimum, to make maintenance easier. (In fact I'm now thinking I will refactor this into its own file and keep the original formatting as well.)

willsimmons1465 (approved):
Much tidier than the previous file, thanks @cqc-alec !

PR #1865, opened 2025-04-23
Doc fixes from converting pybind11 to nanobind

willsimmons1465:
# Description

#1851 switched our binder code from pybind11 to nanobind. This appeared to cause a number of issues with the docs where things would not be formatted clearly or even included in the docs at all.
- Any short docstring would be created as a single line in the stubs. Any docstring with this trait would not render properly.
- Static methods show as `<nanobind.nb_func object>` with no signature or docstring.
- Many methods rendered without a signature, just the name of the method.
- Many methods at the module level (i.e. not contained within a class) were not included.
- Enum docs listed names with their numeric values but lack any of the actual descriptive docstrings.

All of these can be fixed by replacing `automodule` and `autoclass` directives with manually listing every method and entry we wish to document, as we already do with the page for the `Circuit` class.

There are still a few notable changes but relatively minor all things considered:
- Typehints are now much longer, e.g. `abc.collections.Sequence` instead of just `Sequence` or `list`.
- For any method where the binder does not declare argument names (e.g. with `nb.arg("arg_name")`), the arguments are now position-only and marked with the standard slash indicator.
- The `name` property is removed from enums and so cannot be included in the docs (it is still a property of each specific enum value, but that is not easy to get to print out in the docs).
- Sphinx on my mac outputs a new warning: `docstring of pytket._tket.circuit.OpType:1: WARNING: Explicit markup ends without a blank line; unexpected unindent.` caused by `OpType.from_name`, but I can't figure out what is causing it.

Some additional quick fixes also made, like any read-only property whose docstring is a single line starting with `:return:` having the docstring split into a standard string and an empty return description.

# Related issues

Follows-on from #1851 

# Checklist


pytket/docs/pauli.rst
```diff
@@ -1,5 +1,35 @@
 pytket.pauli
 ==================================
-.. automodule:: pytket._tket.pauli
-    :members:
-    :special-members: __init__
+.. currentmodule:: pytket._tket.pauli
+
+.. autoclass:: pytket.pauli.Pauli
```
cqc-alec, line 5: Should this be an `autoenum`?

willsimmons1465: Yes, thanks for spotting

cqc-alec (approved):
Great, thanks!

CalMacCQ:
Looks great :) The OpType page in particular looks better

Is there a formatting issue with `PhaseGadget` here? I wasn't sure if `OpType.PhaseGadget` should even be exposed.
Screenshot 2025-04-23 at 18 03 17

cqc-alec:
> Looks great :) The OpType page in particular looks better
> 
> Is there a formatting issue with `PhaseGadget` here? I wasn't sure if `OpType.PhaseGadget` should even be exposed to python.

Good spot, there's a missing backslash in the math (should be double-backslash preceding "alpha").

cqc-alec (approved):

PR #1878, opened 2025-04-28
Move special `OpType` handling so that it is always applied

### LINKED ISSUE #1873: OpType.CnZ only accepts >0 control qubits but should accept >=0
Currently fails tests at pytket-qiskit

```
FAILED qiskit_convert_test.py::test_convert_cnz_to_qiskit - qiskit.circuit.exceptions.CircuitError: 'The number of control qubits must be in `[1, num_qubits - base_gate.num_qubits]`.'

cqc-alec:
Fixes #1873 .


yao-cqc (approved):
Thanks!

PR #1888, opened 2025-04-30
Build stable-ABI wheels for Python >= 3.12

### LINKED ISSUE #1722: Investigate adding stable ABI wheels for pytket
See https://docs.python.org/3/c-api/stable.html#stable

They could be uploaded in addition to the existing ones, or replace the less-used ones.

cqc-alec:
Closes #1722 .

We can now build stable ABI wheels for Python >= 3.12. Wheels for Python 3.10 and 3.11 still have to be built separately since nanobind does not support stable ABI wheels below 3.12.

Release workflow tested [here](https://github.com/CQCL/tket/actions/runs/14753995944).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-24.04
     strategy:
       matrix:
-        python3-version: ['10', '11', '12', '13']
+        python3-version: ['10', '11', '12']
     steps:
     - uses: actions/checkout@v4
       with:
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-24.04
     strategy:
       matrix:
-        python3-version: ['10', '11', '12', '13']
```
sjdilkes, line 17: Is dropping '13' here intentional for building wheels >=3.12?

sjdilkes (changes requested):
Just one quick question

.github/workflows/release.yml
cqc-alec: > Is dropping '13' here intentional for building wheels >=3.12?

Yes! We don't need to anymore, since the wheels built with 3.12 also work with 3.13.

sjdilkes: Oops yes ok I thought I'd misunderstood, thanks !

sjdilkes (approved):

Issue #1857, opened 2025-04-16
Fix crosslinks across docs


willsimmons1465:
On the pytket docs site, most cases where a class or method are mentioned inline they either fail to link (e.g. ``:py:class:`Circuit` ``) or are just in italics (e.g. `` `Circuit` ``). In the review of #1850, @CalMacCQ pointed out that the crosslink will only work when we write the full prefix (e.g. ``:py:class:`~pytket.circuit.Circuit` ``). This issue is a chore task to work through the API docs in the python binders and all of the .rst files to reformat any class or method references to actually render and crosslink correctly.

CalMacCQ:
Good idea. This sort of crossreferencing is done a fair bit in the manual but I agree it would be nice to do it across the API docs as well.

I've found MyST markdown easier to work with for this sort of thing so maybe its worth converting the source files to MyST as we did for the manual.

PR #1895, opened 2025-05-07
Convert doc pages from rst to myst

willsimmons1465:
As suggested in #1857, it would be good to unify the docs to just one of rst and myst for consistency and maintenance ease.
Automatically converted the rst files using the [rst-to-myst](https://github.com/executablebooks/rst-to-myst) tool.


pytket/docs/changelog.md
```diff
@@ -0,0 +1,1750 @@
...
+- Enable deserialization of `NormaliseTK2` pass.
+- Don't remove classical wires in `FlattenRelabelRegistersPass`.
+
+## 2.1.0 (March 2025)
+
+Features:
+
+- Add `target_2qb_gate` parameter to `CliffordSimp` pass and `OptimiseCliffords`
+  transform.
+- Add `tk2_to_cx` parameter to `RebaseToCliffordSingles` transform.
```
CalMacCQ, line 80: Did you intend to do the myst crossreffing in this PR or is that to be done in a follow-up?

pytket/docs/getting_started.md
```diff
@@ -0,0 +1,109 @@
...
+[pytket extensions](extensions.html) documentation.
+
+The quantum circuit is an abstraction of computation using quantum resources,
+designed by initializing a system into a fixed state, then mutating it via
+sequences of instructions (gates).
+
+The native circuit interface built into pytket allows us to build circuits and
+use them directly.
+
+```
```
CalMacCQ, line 32: Minor thing but for code snippets like this you can do the following



"```{code-cell} ipython3

from pytket import Circuit

trivial_circ = Circuit()        # no qubits or bits
quantum_circ = Circuit(4)       # 4 qubits and no bits
mixed_circ   = Circuit(4, 2)    # 4 qubits and 2 bits
named_circ   = Circuit(2, 2, "my_circ")


This will match the styling of the manual examples once the msyt fix is in effect.


For pages with cell executions you need to add this metadata to the top of each page

```
---
file_format: mystnb
kernelspec:
  name: python3
---
```

pytket/docs/getting_started.md
```diff
@@ -0,0 +1,109 @@
...
+
+# Compile Circuit to use supported gates of IBMQ Nairobi
+compiled_circ = nairobi_device.get_compiled_circuit(circ)
+result = backend.run_circuit(compiled_circ, n_shots=100)
+```
+
+Here the default compilation pass is applied by {py:meth}`IBMQBackend.get_compiled_circuit`. See [this page](https://docs.quantinuum.com/tket/extensions/pytket-qiskit/#default-compilation) for more details.
+
+As an alternative, We can experiment with constructing our own circuit compilation routines in pytket. Passes from the {py:mod}`pytket.passes` module can be applied individually or composed in sequence.
+See the section of the user manual on [circuit compilation](https://docs.quantinuum.com/tket/user-guide/manual/manual_compiler.html) and the corresponding [notebook example](https://docs.quantinuum.com/tket/user-guide/examples/circuit_compilation/compilation_example.html) for more.
```
CalMacCQ, line 109: We should find a way to remove this full URL links but that's for a follow up PR

pytket/docs/index.md
```diff
@@ -0,0 +1,128 @@
...
+Extensions are separate python packages can be installed using `pip`. The installation command is `pip install pytket-X` where `X` is the name of the extension.
+
+To install the `pytket-quantinuum` package use the following command.
+
+```
+pip install pytket-quantinuum
+```
+
+The extensions supported by tket are described
+[here](extensions.html).
```
CalMacCQ, line 38: .md should work here as well

pytket/docs/index.md
```diff
@@ -0,0 +1,128 @@
...
+
+```
+pip install pytket
+```
+
+If you have issues installing `pytket` please visit the [installation troubleshooting](https://docs.quantinuum.com/tket/api-docs/install.html) page.
+
+To use `pytket`, you can simply import the appropriate modules into your python code or in an interactive Python notebook. We can build circuits directly using the `pytket` interface by creating a blank circuit and adding gates in the order we want to apply them.
+
+See the [Getting Started](getting_started.html) page for a basic tutorial on using
```
CalMacCQ, line 18: `.md` should work here

CalMacCQ (commented):
Looks great :) just a few minor points.

CalMacCQ:
Oh also it could be worth linking to the MyST docs from the README of the docs directory so that any maintainers can get up to speed with the syntax.

pytket/docs/changelog.md
willsimmons1465: No, I figured I would keep this PR as close as possible to just running the converter. Turning on the nitpicky sphinx option showed hundreds of already broken crossrefs that would need to be fixed by hand before we even get to adding new references, both of which would have required this PR to have a full in-depth review rather than just a skim over to check what was done by the automatic tool

CalMacCQ: Makes sense :) Just checking

pytket/docs/getting_started.md
willsimmons1465: I have added these in but made the cells skip execution because they would use a qasm files that doesn't exist or qiskit and pytket-qiskit that aren't installed (not sure how the build system works with environments - would I have to install them with poetry to get them to build?)

CalMacCQ: Yes you would need to install pytket-qiskit with poetry for this to be runable. I think `skip-execution` is the way to go here as I'd prefer to not introduce a pytket-qiskit dependency.

CalMacCQ (approved):

PR #1910, opened 2025-05-14
Resolve nitpicky warnings/broken cross-references

willsimmons1465:
# Description

Turning on `nitpicky` raises warnings for any e.g. ``:py:class:`ClassName` `` that fails to link to the docs for `ClassName`, making sure CI can catch this. This PR fixes all existing failures in the tket docs. Most just need to be fixed using the prefix dot to search the API ref for the corresponding entry and tilde to ignore the path.

There are still a large number of instances of things that could be made into cross-references e.g. currently just `` `ClassName` `` or ``ClassName`` without any ``:py:class:`` directive, but those will require an even longer, more tedious manual trawl through the docs to change all of them.

# Related issues

Work towards #1857 

# Checklist


pytket/docs/build-docs.sh
```diff
@@ -8,9 +8,9 @@ PYTKET_VERSION="$(pip show pytket| grep Version | awk '{print $2}')"
 PACKAGE="pytket $PYTKET_VERSION"
 
 # Build the docs setting the html title to show the correct pytket version.
-sphinx-build -W -b html . build -D html_title="$PACKAGE API documentation"
```
CalMacCQ, line 11: Did you intend to remove the `-W` flag here?

willsimmons1465: Ah, forgot to put it back. It is helpful to remove when running locally so that it reports all warnings rather than just exiting after reaching the first one

pytket/docs/build-docs.sh
```diff
@@ -8,9 +8,9 @@ PYTKET_VERSION="$(pip show pytket| grep Version | awk '{print $2}')"
...
+sphinx-build -W -b html . build -D html_title="$PACKAGE API documentation" || exit 1
```
CalMacCQ, line 11: Good spot with the `|| exit 1`

You can also use exit flags `+/-e` like I did in this PR -> https://github.com/CQCL/pytket-qiskit/pull/459/files#diff-24d28dd390892366ca6afd4b10fd1c0bcf44bffdd7735881e341c44d051d4fcdR42

However the effect should be the same I think so happy leaving this as is.

CalMacCQ (approved):
Thanks a lot for fixing this up Will

PR #1920, opened 2025-05-27
Update to boost 1.88.0

cqc-alec:


cqc-alec:
We could apply a patch to fix this, but as there's no urgency to update I think we can just wait for the next release when this will hopefully be fixed.

PR #1956, opened 2025-07-02
Ae/rng

cqc-alec:
# Description

Please summarise the changes.

# Related issues

# Checklist


pytket/binders/circuit/Circuit/add_classical_op.cpp
```diff
@@ -272,6 +272,110 @@ void init_circuit_add_classical_op(nb::class_<Circuit> &c) {
...
+            const nb::kwargs &kwargs) {
+            std::shared_ptr<OpaqueClassicalOp> op = std::make_shared<OpaqueClassicalOp>(OpType::RNGSeed);
+            std::vector<UnitID> args;
+            for (const Bit& arg : creg.to_vector()) {
+              args.push_back(arg);
+            }
+            args.push_back(RngState(rng_wire));
+            return add_gate_method_any(&circ, op, args, kwargs);
+           },
+           "Seed an RNG from the contents of a classical register."
```
cqc-melf, line 286: Why is this limited to 64 bit size registers?

pytket/binders/circuit/main.cpp
```diff
@@ -73,7 +73,8 @@ NB_MODULE(circuit, m) {
       .value("Boolean", EdgeType::Boolean)
       .value("Classical", EdgeType::Classical)
       .value("Quantum", EdgeType::Quantum)
-      .value("WASM", EdgeType::WASM);
+      .value("WASM", EdgeType::WASM)
+      .value("RNG", EdgeType::RNG);
```
cqc-melf, line 77: Is the RNG type only used for the random number function or also for the job ID and job number?

pytket/pytket/circuit/__init__.py
```diff
@@ -160,6 +160,118 @@ def add_wasm_to_reg(  # noqa: PLR0913
 setattr(Circuit, "add_wasm_to_reg", add_wasm_to_reg)  # noqa: B010
 
 
+def set_rng_seed(self: Circuit, creg: BitRegister, **kwargs: Any) -> Circuit:
```
cqc-melf, line 163: Have you thought about adding a second function taking a bit list instead of `BitRegister` ?

cqc-melf (commented):
few comments.

Should we check that other conversion are showing clear errors?

Did you try to import qasm containing this?

pytket/binders/circuit/main.cpp
cqc-alec: Only for the RNG functions. There's no (in-circuit) state associated with the job number functions.

cqc-melf:
> > Should we check that other conversion are showing clear errors?
> 
> Which conversions do you mean here?
> 
> > Did you try to import qasm containing this?
> 
> No, I am not implementing QASM import.

I understand that we don't want to support this at the moment. The question is: are there clear error messages when importing qasm containing this functions? (Would be good do check for QIR as well after this is released)

cqc-alec:
> > > Should we check that other conversion are showing clear errors?
> > 
> > 
> > Which conversions do you mean here?
> > > Did you try to import qasm containing this?
> > 
> > 
> > No, I am not implementing QASM import.
> 
> I understand that we don't want to support this at the moment. The question is: are there clear error messages when importing qasm containing this functions? (Would be good do check for QIR as well after this is released)

I will add a test for that.

cqc-melf (approved):

PR #1973, opened 2025-07-08
Try to make tket thread-safe

### LINKED ISSUE #1953: Allow concurrency between `Transforms`
Running Transforms::two_qubit_squash and Transforms::clifford_resynthesis concurrently (on different circuits), I get a race condition (in the form of a SIGSEGV). Some concurrency on a (very limited) subset of tket transforms would be very useful to have.

I had a quick look at the code but have no idea where any shared memory between passes might hide...

cqc-alec:
Fixes #1953 .

Along with #1971 this seems to solve the concurrency issue described in the bug report. I have not tested other concurrency scenarios so there may be other issues lurking, but this seems like a good step at least.

We skip the concurrency test for debug builds, as it seems there is still an issue there coming from the teuchos library (used by debug builds of symengine). This warrants further investigation but _may_ be outside our control.

### DIFF
--- a/pytket/conanfile.py
+++ b/pytket/conanfile.py
@@ -37,7 +37,7 @@ class pytketRecipe(ConanFile):
         self.requires("nanobind/tci-2.7.0@tket/stable")
         self.requires("symengine/tci-0.14.0.1@tket/stable")
         self.requires("tkassert/0.3.4@tket/stable")
-        self.requires("tket/2.1.32@tket/stable")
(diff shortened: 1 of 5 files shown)


lmondada (approved):
Excellent! So the issue here was the `static` symbols set in `tket/src/Gate/SymTable.cpp`, right? Maybe add a comment next to the lock explaining why it's required, as it's fairly uncommon in the rest of the codebase.

I'll try removing the locks in `tket-c-api` next week and am excited to see if it works!

PR #1986, opened 2025-07-23
Canonicalise docs links to avoid `_tket`

willsimmons1465:
# Description

When building the tket docs, the objects.inv file (setting up the hooks for intersphinx) creates entries for e.g. `pytket._tket.circuit.Circuit` that are assigned to url anchors `pytket.circuit.Circuit` without the `_tket`. Additionally, there are no entries for `pytket.circuit.Circuit`. This forces us to link to `pytket._tket.X` in extension repo docs, which can sometimes cause issues with linkchecks (see https://github.com/CQCL/pytket-braket/actions/runs/16376239935/job/46276780660?pr=194) because a link to `pytket._tket.circuit.Circuit` may naively try to find the anchor of the same name which does not exist.

The solution in this PR is to avoid using `_tket` in `currentmodule` and autodocs for anything other than module entries. The produced objects.inv file still creates some entries under `_tket` for classes (but not their methods), but adds additional entries for their non-`_tket` variants which guarantees we can use them and the links will be found as needed.

I threw in some additional fixes to docstrings of `Backend`: one typo, the others are using fully specified cross-references so that, whenever a subclass in an extension repository inherits the method and tries to build its docstring in a different project, intersphinx will still handle things.

# Related issues

Uses the updated coverage ignores from CQCL/pytket-docs-theming#21
Blocks CQCL/pytket-braket#194

# Checklist


pytket/docs/build-docs.sh
```diff
@@ -2,8 +2,6 @@ cp -R pytket-docs-theming/_static .
 cp -R pytket-docs-theming/quantinuum-sphinx .
 cp pytket-docs-theming/conf.py .
 
-echo "coverage_ignore_functions.extend([\"get_job_shot_num\", \"get_rng_num\", \"set_rng_bound\", \"set_rng_index\", \"set_rng_seed\"])" >> conf.py
-
```
CalMacCQ, line 6: Now that we are removing the `_tket` prefix is it possible to remove these lines from the `build-docs.sh` script?

```shell
if [[ "$OSTYPE" == "darwin"* ]]; then
    find build/ -type f -name "*.html" | xargs sed -e 's/pytket._tket/pytket/g' -i ""
    sed -i '' 's/pytket._tket/pytket/g' build/searchindex.js
else
    find build/ -type f -name "*.html" | xargs sed -i 's/pytket._tket/pytket/g'
    sed -i 's/pytket._tket/pytket/g' build/searchindex.js
fi
```

The purpose of this script is to remove all `_tket` links which show up in the html after the sphinx build runs.

CalMacCQ (commented):
Nice, just one comment about the build script.

I've also approved the docs theming PR so maybe the submodule should be updated once thats merged.

pytket/docs/build-docs.sh
willsimmons1465: I tried removing it locally and everything seemed to build perfectly fine for me. To clarify, any stubs generated through nanobind will still include and reference `_tket` so we can't avoid mentioning it at all on the docs, but they still seem to link fine without this fix

PR #1988, opened 2025-07-24
chore: Single source of truth for tket version

lmondada:
This adds a `TKET_VERSION` file at the root of the repo, to which all conan recipes refer to to set their own versions resp. dependencies.


tket/conanfile.py
```diff
@@ -85,6 +84,28 @@ def generate(self):
             tc.variables["BUILD_TKET_PROPTEST"] = True
         tc.generate()
 
+    def set_version(self):
+        """Set the version of this package from the TKET_VERSION file."""
+
+        # TKET_VERSION is in the parent directory in the repo, but will be in
+        # the current directory once the package has been exported to conan cache
+        for try_dir in [".", ".."]:
```
cqc-alec, line 92:
```suggestion
        for try_dir in [os.curdir, os.pardir]:
```

tket/conanfile.py
```diff
@@ -85,6 +84,28 @@ def generate(self):
...
+                return
+
+        raise FileNotFoundError("TKET_VERSION not found")
+
+    def export(self):
+        # Copy the TKET_VERSION file to the export folder
+        copy(
+            self,
+            "TKET_VERSION",
+            os.path.join(self.recipe_folder, ".."),
```
cqc-alec, line 105:
```suggestion
            os.path.join(self.recipe_folder, os.pardir),
```

.github/workflows/build_and_test.yml
```diff
@@ -54,7 +54,7 @@ jobs:
     - name: parse version from conanfile
```
cqc-alec, line 54:
```suggestion
    - name: Read version from file
```

cqc-alec (approved):
Thanks!

PR #2031, opened 2025-09-15
Don't include workflow file in test for changes.

### LINKED ISSUE #2030: CI: Don't require tket version bump for changes to workflow files only

cqc-alec:
Closes #2030 .


lmondada (approved):
There is logic in leaving this in: changing the workflow file could change _how_ tket libraries are built, and therefore result in different binaries that would have to be recompiled with a new version tag.

This is probably an edge case much rarer than simple amendments/fixes to the non-build related parts of the workflow. We can trigger version bumps manually when we hit this edge case.

I'll approve and leave it to you decide.

cqc-alec:
> There is logic in leaving this in: changing the workflow file could change _how_ tket libraries are built, and therefore result in different binaries that would have to be recompiled with a new version tag.
> 
> This is probably an edge case much rarer than simple amendments/fixes to the non-build related parts of the workflow. We can trigger version bumps manually when we hit this edge case.
> 
> I'll approve and leave it to you decide.

Yes you are quite right, but I do think it is an edge case and the inconvenience and delay of having to update the tket version when modifying the workflow file outweighs the extra safety provided. I think that making modifications to the workflow file that change the resulting binaries, without making any changes to the other files, is going to be pretty rare.

PR #2039, opened 2025-09-23
vscode circuit display issue

DNA386:
# Description

Amend the circuit display code to fix display issue that occurs in VS code.

This occurred due to how external links embedded in an iframe are handled: we must hence load these directly so that they are cached and available when the Iframe loads.

# Related issues

N/A

# Checklist

- [x] (N/A) I have made corresponding changes to the public API documentation.
- [x] (N/A) I have added tests that prove my fix is effective or that my feature works.
- [x] (N/A) I have updated the changelog with any user-facing changes.

### DIFF
--- a/pytket/pytket/circuit/display/static/circuit.html
+++ b/pytket/pytket/circuit/display/static/circuit.html
@@ -15,7 +15,7 @@
     top:0;bottom:0;left:0;right:0;
     margin:auto;resize:both;display:block;overflow:hidden;
     width:min({{ min_width }}, 100%);
-    height:min({{ min_height }},100%);
+    height:min({{ min_height }}, 100%);
     max-width:100%;
     max-height:100%;">
 {% endif %}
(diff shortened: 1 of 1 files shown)


CalMacCQ (approved):
I've tested this locally with the circuit that Maria was having trouble viewing and it works, thanks.

I noticed that some circuits seem to still display fine in VSCode notebooks for instance

```python
circ = Circuit(3).H(0).CX(0, 1).CX(0, 2)
render_circuit_jupyter(circ)
```

However Maria's example seems not to work for me (prior to your change).

```python
circ = Circuit(2).CX(0, 1).Rz(0.2, 1).Rx(0.1, 1).CX(0, 1)
render_circuit_jupyter(circ)
```

Any idea why one works and not the other?

CalMacCQ:
> hmm. perhaps the issue is to do with the math display (this requires the mathjax external library)? The working circuit you sent doesn't have any parameters, while the ones I was looking at and Maria's all do. I also noticed that this can cause the offline renderer to have some artifacts - will make a separate PR there to add that in properly.

Makes sense, I guess there's something going on with the parametrised gates. Happy for this fix to go in.

PR #2052, opened 2025-10-24
Restore MacOS x86 support

### LINKED ISSUE #2051: pytket version 2.10.1 not compatible with MacOS 14
I was trying to upgrade to the latest PyTKET version, as it's a dependency of [ucc-bench](https://github.com/unitaryfoundation/ucc-bench/pull/181), but as I am running MacOS 14 Sonoma, the v2.10.1 would not install. Full error message:
```bash
error: Distribution `pytket==2.10.1 @ registry+https://pypi.org/simple` can't be installed because it doesn't have a source distribution or wheel for the current platform

hint: You're on macOS (`macosx_14_0_x86_64`), but `pytket` (v2.10.1) only has wheels for the following platforms: `manylinux_2_26_aarch64`, `manylinux_2_27_x86_64`, `manylinux_2_28_aarch64`, `manylinux_2_28_x86_64`, `macosx_13_0_arm64`, `win_amd64`
```
Are you planning to add support for MacOS 14? I noticed a related issue #2048, so perhaps a numba dependency is blocking here?

cqc-alec:
Closes #2051 .

Basically a revert of #2026 , replacing `macos-13` with `macos-15-intel`.

Release build tested [here](https://github.com/CQCL/tket/actions/runs/18776102528).

### DIFF
--- a/pytket/pytket/circuit/decompose_classical.py
+++ b/pytket/pytket/circuit/decompose_classical.py
@@ -55,11 +55,11 @@ class VarHeap(Generic[T]):
 
     def pop(self) -> T:
         """Pop from top of heap."""
-        return heappop(self._heap)
+        return heappop(self._heap)  # type: ignore
 
     def push(self, var: T) -> None:
         """Push var to heap."""
-        heappush(self._heap, var)
+        heappush(self._heap, var)  # type: ignore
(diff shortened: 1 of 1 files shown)


cqc-melf (approved):
Feel free to do a release directly

PR #2102, opened 2026-01-19
Upload manylinux builds of tket and tket-c-api to artifactory

### LINKED ISSUE #2100: Upload manylinux builds of tket-c-api to artidactory
The tket2 builds should be able to fetch these. We currently upload tket-c-api builds for `ubuntu-24.04`, `macos-15-intel`, `macos-26` and `windows-2025`, but should add x86 and arm64 builds for manylinux 2_28.

cqc-alec:
Closes #2100 .

I had to add the linker flag to allow threading when building executables linking against tket, as the gcc on the arm64 manylinux container complained otherwise.


cqc-alec:
> Just a question; `pytket` also builds wheels for `musllinux`. Should we do that here too?

This is news to me, where do you see these?

PR #2125, opened 2026-03-06
chore: Set up explicit re-exports for circuits

johnchildren:
# Description

The recommended pytket convention is to generally import pytket circuits either from the root `pytket` module or the `pytket.circuit` module where they are re-exported from `pytket._tket.circuit`. However some tooling such as `pyright` requires re-exports to be explicit to avoid linting errors which causes code to often end up importing from the "private" module instead. This commit adds explicit re-exports to both `pytket` and `pytket.circuit` based on what the tests and examples in this import from those modules.

This change will allow users to import pytket circuits the recommended way without getting a warning from their editor in future.

# Related issues

# Checklist


cqc-alec (approved):
Thanks John! I'll do an `rc` release after merging this to make sure the extensions are happy and then do a proper release afterwards if so.

Issue #2095, opened 2026-01-12
Migrate docs build to use uv rather than poetry.


CalMacCQ:
All of the extensions now use uv for managing the dependencies required by the docs build. I think we should migrate to uv here for consistency. 

Should be a quick job.

CalMacCQ:
closed by #2130

Issue #2044, opened 2025-10-15
Docs build fails because of a sphinx warning


cqc-alec:
```
Warning, treated as error:
<unknown>:1:py:data reference target not found: typing.Union
error: Recipe `build` failed on line 15 with exit code 2
error: Recipe `build-strict` failed on line 25 with exit code 2
```

CalMacCQ:
closed by #2130 

The strict sphinx build is now enenabled. Sphinx warnings are now treated as errors. We also do link and coverage checking.

PR #2130, opened 2026-03-12
docs: replace poetry with uv in docs build, use strict sphinx build

CalMacCQ:
# Description

This PR refactors the docs build to use `uv` rather than `poetry` closing #2095. This means we no longer need to pin to a specific poetry version #2033 

I've also reenabled the strict sphinx build which treats warnings as errors resolving #2044

Note this PR underscores some internal functions in `combinecondpass` -> [b4ecfd3](https://github.com/Quantinuum/tket/pull/2130/commits/b4ecfd3617666c28c9072704e573fafc8a28f777)
 
# Related issues
closes #2033, #2095, #2044

### DIFF
--- a/pytket/tests/classical_test.py
+++ b/pytket/tests/classical_test.py
@@ -1100,7 +1100,7 @@ def composite_bit_logic_exps(
     for _ in range(terms):
         chosen_operator = draw(operators)
         if chosen_operator is operator.not_:
-            exp = chosen_operator(exp)
+            exp = chosen_operator(exp)  # type: ignore
         else:
             exp = chosen_operator(exp, draw(strategies.one_of(bits, constants)))
     return exp


CalMacCQ:
> > This docs workflow seems to be failing due to a flaky scipy link https://github.com/Quantinuum/tket/actions/runs/23751421534/job/69194304242
> 
> Try rerunning?

I've tried rerunning and altering sphinx config to ignore external broken links. Still seems to be broken :(

pytket/tests/classical_test.py
```diff
@@ -1100,7 +1100,7 @@ def composite_bit_logic_exps(
     for _ in range(terms):
         chosen_operator = draw(operators)
         if chosen_operator is operator.not_:
-            exp = chosen_operator(exp)
+            exp = chosen_operator(exp)  # type: ignore
```
CalMacCQ, line 1103: mypy ....  https://github.com/Quantinuum/tket/actions/runs/23859225024/job/69561030558

pytket/docs/passes.md
```diff
@@ -313,6 +313,12 @@ For more on pytket passes see the [compilation](https://docs.quantinuum.com/tket
 .. automethod:: pytket.passes.ZZPhaseToRz
 ```
 
+```{eval-rst}
+.. automodule:: pytket.passes.combinecondpass
```
cqc-alec, line 317: Do we need to make this public, as well as `CombineCondPass`?

CalMacCQ: This seemed to be the best way to get the coverage checks to pass. Its a module with a single public member. I guess you can also import from pytket.passes.

CalMacCQ: We could also underscore the `combinecondpass` module and reexport from .passes but I'd rather not as that would break anyone using `CombineCondPass`

cqc-alec: Yeah, I do think that would make more sense. I suspect this pass has at most 1 current user so I think we can tolerate the break. But probably it should be in another PR.

PR #2186, opened 2026-06-09
fix: replace _tket modules in docs html pages

CalMacCQ:
# Description

This PR fixes an apparent regression where `_tket` was showing up in the Sphinx API docs. Note that this fix only fixes the local build of the docs. I'll need to make a separate PR to the repo that deploys the website.

EDIT: now fixed at the sphinx configuration level thanks to @sburton84  in https://github.com/Quantinuum/pytket-docs-theming/pull/26

 I think its much better to do 

```python
from pytket.circuit import Op
```

than 

```python
from pytket._tket.circuit import Op
``` 

Before this change:

Screenshot 2026-06-09 at 22 47 00

After:

Screenshot 2026-06-09 at 22 47 11

### DIFF
--- a/pytket/docs/justfile
+++ b/pytket/docs/justfile
@@ -10,23 +10,9 @@ install-dev: prepare
     uv sync
     uv pip install ../.[ZX] -v
 
-# replace _tket module in the html pages.
-# Apple MACOSX and Linux have differing sed replace syntax
-replace:
-    #!/usr/bin/env bash
-
-    if [[ "$OSTYPE" == darwin* ]]; then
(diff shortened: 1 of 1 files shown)


cqc-melf (commented):
It looks like this PR is only changing the submodule? Is that intended?

CalMacCQ:
> It looks like this PR is only changing the submodule? Is that intended?

Yes, I was originally making the fix by adding a sed command to the justfile to replace all of the `_tket` instances. However in  https://github.com/Quantinuum/pytket-docs-theming/pull/26 Sean fixed the issue at the sphinx configuration level. This means that the logic for doing the replacement is in one place and we don't have to make separate changes for this repo and the website.

cqc-melf (approved):

PR #2208, opened 2026-07-10
Circuit renderer browser warnings

sburton84:
# Description

When displaying some of our documentation pages that use the TKET circuit renderer, such as at https://docs.quantinuum.com/inquanto/tutorials/InQ_tut_circ_comp.html, there are lots of warnings in the logs about using the development build of Vue and setting the `viewFormat` property to an invalid `"None"` string. This fixes those warnings by setting `viewFormat` only if the corresponding variable is not `None` (otherwise it will be left to the default value, which I believe is `"row"`), and using the production distribution of Vue. I have also included an environment variable that can be set to switch the Vue version back to the development version in case this is useful for debugging, but this could be removed if it's something that doesn't sound like it would be useful?

# Checklist

- [X] I have performed a self-review of my code.
- [X] I have commented hard-to-understand parts of my code.
- [ ] ~~I have made corresponding changes to the public API documentation.~~
- [ ] ~~I have added tests that prove my fix is effective or that my feature works.~~
- [ ] ~~I have updated the changelog with any user-facing changes.~~

### DIFF
--- a/pytket/pytket/circuit/display/__init__.py
+++ b/pytket/pytket/circuit/display/__init__.py
@@ -66,6 +66,22 @@ loader = PrefixLoader(
 
 jinja_env = Environment(loader=loader, extensions=[IncludeRawExtension])
 
+
+def _use_vue_dev_build() -> bool:
+    """Whether to load the development build of Vue.
+
+    The development build emits extra runtime warnings that are useful when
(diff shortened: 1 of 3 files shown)


DNA386 (approved):
seems good to me.
I don't know that the dev vue is really used from pytket since the circuit renderer development is based in another repo, but maybe worth keeping in just in case

PR #2221, opened 2026-08-18
Update to nanobind 2.15.0

cqc-alec:


pytket/stub_generation/stubgen.py
cqc-alec, line 1: Changes in this file transcribed from corresponding changes in nanobind's `stubgen.py`.

cqc-melf (approved):

PR #2230, opened 2026-09-14
Use correct port when rewiring in `DelayMeasures` pass

### LINKED ISSUE #2229: DelayMeasures crashes with "Vertex has multiple inputs on the same port" when a measured qubit is reused
### Environment
- pytket version: 2.18.1
- Python version: 3.10.12
- Operating system: Ubuntu 22.04.5 LTS (WSL2)

### What is happening?

`DelayMeasures().apply(circuit)` raises a native `RuntimeError: Vertex has multiple inputs on the same port` (no Python frame inside the pass itself) on circuits that measure a qubit mid-circuit and then reuse that qubit in a later gate. The crash specifically needs a *diagonal* single-qubit gate (`t`/`s`/`z`/`tdg`/`sdg`/`rz`) between the measurement and a subsequent two-qubit gate on the same qubit; a non-diagonal gate (`x`, `h`) there does not trigger it, and neither does a bare `measure` immediately followed by the two-qubit gate.

This is not the same case fixed in the 1.33.0 changelog ("`DelayMeasures()` for circuits where bits are reused as measurement targets", i.e. a classical bit reused). Here it's a qubit reused after measurement, and it still reproduces on the current, latest 2.18.1.

### How can we reproduce the issue?

```python
from pytket.passes import DelayMeasures

qasm = """OPENQASM 2.0;
include "qelib1.inc";
qreg q[2];
creg c[1];
measure q[0] -> c[0];
t q[0];
CX q[1],q[0];
"""

circuit = circuit_from_qasm_str(qasm)
DelayMeasures().apply(circuit)
```

Output:
```
RuntimeError: Vertex has multiple inputs on the same port
```

### What should happen?

Either `DelayMeasures` should handle this (fully valid OpenQASM 2) circuit without crashing, or it should reject it with a clear, documented error instead of an internal graph-invariant `RuntimeError`.

### Trigger conditions

Replacing the `t q[0];` line above:

| gate | result |
|---|---|
| `t`, `s`, `z`, `tdg`, `sdg`, `rz(θ)` (Z-diagonal) | crash |
| `x`, `h` (not diagonal) | no crash |
| removed entirely (bare `measure` then `CX`) | no crash |
| same circuit, measurement moved to the end instead | no crash |

So the crash needs all three: (1) a mid-circuit measurement, (2) a Z-diagonal gate on that qubit afterward, (3) that qubit appearing again in a two-qubit gate.

cqc-alec:
Fixes #2229 .


sjdilkes (approved):
LGTM

PR #2233, opened 2026-09-15
Insert `RemoveRedundancies` at the beginning of `FullPeepholeOptimise`

### LINKED ISSUE #2232: A self-cancelling gate pair exposes a potential missed optimization in `FullPeepholeOptimise`
### Environment
- pytket version: 2.18.1
- Python version: 3.10.12
- Operating system: Ubuntu 22.04.5 LTS (WSL2)

### What is happening?

Two circuits that are unitary-equivalent up to global phase (one is the other with a self-cancelling `CSWAP; CSWAP` pair inserted) get optimized by `FullPeepholeOptimise()` to substantially different sizes: 22 gates versus 17 gates. In other words, inserting a redundant gate pair enables the optimizer to find a strictly smaller circuit.

### How can we reproduce the issue?

```python
from pytket.passes import FullPeepholeOptimise

def ops(circuit):
    counts = {}
    for cmd in circuit.get_commands():
        name = cmd.op.type.name
        if name != "Measure":
            counts[name] = counts.get(name, 0) + 1
    return sum(counts.values()), counts

BASE = """OPENQASM 2.0;
include "qelib1.inc";
qreg q[5];
creg c[5];
h q[2];
ch q[2], q[0];
ch q[2], q[0];
cswap q[2], q[1], q[3];
cswap q[2], q[1], q[3];
ccx q[2], q[4], q[0];
cz q[1], q[2];
measure q[0] -> c[0];
measure q[1] -> c[1];
measure q[2] -> c[2];
measure q[3] -> c[3];
measure q[4] -> c[4];
"""

# BASE, with a redundant "cswap q[4],q[2],q[3]; cswap q[4],q[2],q[3];" (== identity)
# inserted immediately after "h q[2];"
WITH_IDENTITY_PAIR = BASE.replace(
    "h q[2];",
    "h q[2];\ncswap q[4], q[2], q[3];\ncswap q[4], q[2], q[3];",
    1,
)

base = circuit_from_qasm_str(BASE)
FullPeepholeOptimise().apply(base)
print("BASE          :", ops(base))

with_pair = circuit_from_qasm_str(WITH_IDENTITY_PAIR)
FullPeepholeOptimise().apply(with_pair)
print("BASE + CSWAP^2:", ops(with_pair))
```

Output:
```
BASE          : (22, {'TK1': 14, 'CX': 8})
BASE + CSWAP^2: (17, {'TK1': 10, 'CX': 7})
```

Both inputs are the same unitary up to global phase (confirmed with `Operator.equiv`). Replaying constituent passes one at a time on both inputs shows they track each other closely right up until `ThreeQubitSquash`, where BASE barely improves and the CSWAP-augmented one drops much further.

### What should happen?

Ideally, removing an immediately self-cancelling gate pair would not be expected to lead to a worse optimization result. In this case, however, the circuit with the redundant CSWAP; CSWAP pair is optimized to a smaller circuit than the otherwise identical circuit without it.

### Possible mitigation

As a workaround, applying `RemoveRedundancies()` immediately before `FullPeepholeOptimise()` closes the gap in this example. We also tested this ordering across random circuits, QASMBench/Feynman/MQT benchmarks, and several redundant insertion types and locations; in our experiments, the added cleanup did not worsen the resulting gate count.

cqc-alec:
Closes #2232 .


cqc-alec:
I still don't know why I was seeing different (in fact, better) results when building locally using cmake directly rather than conan. But I will investigate this separately.

sjdilkes (approved):
LGTM

PR #2239, opened 2026-09-21
Fix endianness of nested `Conditional` values in `GreedyPauliSimp`

### LINKED ISSUE #2237: `GreedyPauliSimp` changes nested conditional semantics
### Environment

pytket 2.18.3, Python 3.10.14, Linux x86-64.

### Problem

`GreedyPauliSimp` silently changes the condition value when flattening nested `Conditional` operations. In this example, `(c[0] == 1) and (c[1] == 0)` becomes `c == 2` instead of `c == 1`, changing the final measurement outcome.

### Reproducer

```python
from pytket import Circuit, OpType
from pytket.circuit import Conditional, Op
from pytket.passes import GreedyPauliSimp
from pytket.qasm import circuit_to_qasm_str

c = Circuit()
q = c.add_q_register("q", 3)
cr = c.add_c_register("c", 2)
out = c.add_c_register("out", 1)
c.X(q[0])
c.Measure(q[0], cr[0]).Measure(q[1], cr[1])
nested = Conditional(Conditional(Op.create(OpType.X), 1, 0), 1, 1)
c.add_gate(nested, [cr[0], cr[1], q[2]])
c.Measure(q[2], out[0])

print("original:", c.get_commands())
GreedyPauliSimp().apply(c)
print("transformed:", c.get_commands())
print(circuit_to_qasm_str(c))
```

The transformed circuit contains `IF ([c[0], c[1]] == 2) THEN X q[2]`, exported as `if(c==2) x q[2];`. With all qubits initially zero, the original circuit deterministically produces `out[0] = 1`, while the transformed circuit produces `out[0] = 0`. The input satisfies the pass's preconditions.

### Expected behavior

Preserve the classical condition and measurement outcomes. The combined condition in this example should be `c == 1`.

### Suggested fix

In `GreedyPauliConverters.cpp`, condition bits are appended outer-first, but line 326 shifts the accumulated outer value to the high bits:

https://github.com/Quantinuum/tket/blob/c9824613522abfc30b564a8a317ac2fb3ecf7f7a/tket/src/Transformations/GreedyPauliConverters.cpp#L319-L328

`Conditional.hpp` specifies little-endian condition values. Synthesis retains the outer-first bit order, so the correct combined value is `v_outer + (v_inner << w_outer)`, not `(v_outer << w_inner) + v_inner`.

https://github.com/Quantinuum/tket/blob/c9824613522abfc30b564a8a317ac2fb3ecf7f7a/tket/include/tket/Circuit/Conditional.hpp#L30-L39

https://github.com/Quantinuum/tket/blob/c9824613522abfc30b564a8a317ac2fb3ecf7f7a/tket/src/Transformations/GreedyPauliOptimisation.cpp#L547-L555

cqc-alec:
Fixes #2237 .


yao-cqc (approved):
Thank you!

